PR #6, opened 2021-05-12
Add sphinx based documentation for the project, and expand API docs

robertmaynard:
The original `rapids-cmake` documentation was added without any verification that it rendered correctly.

This sets up the `docs/`  folder and mimics CMake infrastructure to build Sphinx based documentation for `rapids-cmake`.


docs/conf.py
```diff
@@ -0,0 +1,133 @@
...
+todo_include_todos = False
+
+
+# -- Options for HTML output ----------------------------------------------
+
+# The theme to use for HTML and HTML Help pages.  See the documentation for
+# a list of builtin themes.
+#
+
+html_theme = "sphinx_pydata_theme"
```
robertmaynard, line 87: @kkraus14 Is this the correct new theme for RAPIDS docs?

kkraus14: I would defer to @raydouglass as the ops team was planning on using this theme as a base and then giving it a bit of a RAPIDS tweak.

PR #7, opened 2021-05-14
Support searching existing cpm build dirs

### LINKED ISSUE #1: [FEA] set `${pkg}_DIR` for project dependencies to assist with building dependencies in a common build dir
In [node-rapids](https://github.com/rapidsai/node-rapids) we're taking advantage of a centralized source and build cache for all our C++ dependencies. We have multiple `CMakeLists.txt` project files that depend on certain expensive-to-build RAPIDS libraries like `libcudf`, and we don't want to build those libraries multiple times.

CPM has a utility for telling it to centralize where it downloads dependency sources via `CPM_SOURCE_CACHE`, but it has no corresponding setting for centralizing dependency build dirs i.e. `CPM_BINARY_CACHE`.

We can trick CPM (everything else that uses FetchContent) into centralizing build dirs via the `FETCHCONTENT_BASE_DIR` variable.

However, we later run into issues when a built project's export targets go to look up its own dependencies via `FindPackage()`. The dependencies aren't installed, and they're not in a location `FindPackage()` knows to search.

And since `FetchContent` doesn't create directory names `FindPackage` recognizes as locations a package could be (i.e. `thrust` vs. `thrust-build`), we can't simply do `list(APPEND CMAKE_PREFIX_PATH ${FETCHCONTENT_BASE_DIR})`.

Our workaround for this is to [set `${pkg}_DIR` for a project's dependencies](https://github.com/rapidsai/node-rapids/blob/dfdc0a7349646ad1dfe3b543a2d7d694d779eb8e/modules/core/cmake/Modules/ConfigureCUDF.cmake#L23-L29) before calling `CPMFindPackage()`. We have this [little CMake function](https://github.com/rapidsai/node-rapids/blob/dfdc0a7349646ad1dfe3b543a2d7d694d779eb8e/modules/core/cmake/Modules/get_cpm.cmake#L79-L88) to help us:

```cmake
function(_set_package_dir_if_exists pkg dir)
  if (EXISTS "${CPM_BINARY_CACHE}/${dir}-build")
      set(${pkg}_DIR "${CPM_BINARY_CACHE}/${dir}-build" PARENT_SCOPE)
  endif()
endfunction()
```

robertmaynard:
Search for an existing CPM build directory for each package we add via `rapids_cpm_add`. If we find a suitable build directory with a config module we will specify `<PackageName>_DIR`.

Fixes #1

### DIFF
--- a/rapids-cmake/cpm/find.cmake
+++ b/rapids-cmake/cpm/find.cmake
@@ -101,7 +101,9 @@ function(rapids_cpm_find name version )
   if(DEFINED FETCHCONTENT_BASE_DIR)
     if (EXISTS "${FETCHCONTENT_BASE_DIR}/${name}-build/${lowercase_name}-config.cmake" OR
         EXISTS "${FETCHCONTENT_BASE_DIR}/${name}-build/${name}Config.cmake")
-      set(${name}_DIR "${FETCHCONTENT_BASE_DIR}/${name}-build/")
+      if (NOT DEFINED ${name}_DIR)
+        set(${name}_DIR "${FETCHCONTENT_BASE_DIR}/${name}-build/")
+      endif()
     endif()
   endif()
 
--- a/rapids-cmake/cpm/find.cmake
+++ b/rapids-cmake/cpm/find.cmake
@@ -76,6 +76,13 @@ consistency. List all targets used by your project in `GLOBAL_TARGET`.
   Required placeholder to be provied before any extra arguments that need to
   be passed down to :cmake:command:`CPMFindPackage`.
 
+Search Behavior
+^^^^^^^^^^^^^^^
+
+  In addition to the default `CPM` search behavior, :cmake:command:`rapids_cpm_find`
+  looks for existing build directories of the given project and will prefer
+  those when they exist and have a `<PackageName>-config.cmake` or a
+  `<PackageName>Config.cmake` file.
 
 #]=======================================================================]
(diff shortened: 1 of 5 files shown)


rapids-cmake/cpm/find.cmake
```diff
@@ -81,6 +96,15 @@ function(rapids_cpm_find name version )
     message(FATAL_ERROR "rapids_cpm_find requires you to specify CPM_ARGS before any CPM arguments")
   endif()
 
+  # Prefer an existing build of the package if it exists
+  string(TOLOWER ${name} lowercase_name)
+  if(DEFINED FETCHCONTENT_BASE_DIR)
+    if (EXISTS "${FETCHCONTENT_BASE_DIR}/${name}-build/${lowercase_name}-config.cmake" OR
+        EXISTS "${FETCHCONTENT_BASE_DIR}/${name}-build/${name}Config.cmake")
+      set(${name}_DIR "${FETCHCONTENT_BASE_DIR}/${name}-build/")
```
trxcllnt, line 104:
```suggestion
      if (NOT DEFINED ${name}_DIR)
        set(${name}_DIR "${FETCHCONTENT_BASE_DIR}/${name}-build/")
      endif()
```

robertmaynard:
I am going to close this for now as @trxcllnt and I review the requirements as this could break projects that depend on different git SHA1's of a project ( with the same version number ).

PR #9, opened 2021-05-14
Rapids export properly sets version variables

robertmaynard:
While `rapids-cmake` generated a `project-config.cmake` that listed it created version variables, it actually did not. 

This now generates the version values as encoded in the `project-config-version` file.

Also corrects an issue discovered while testing where `rapids_export` couldn't handle projects that generated no targets.


robertmaynard:
Depends on https://github.com/rapidsai/rapids-cmake/pull/8

PR #27, opened 2021-06-03
Handle leading zeroes in `rapids_cmake_write_version_file`

robertmaynard:
This removes the chance of generating octal values in C++ when the input looks like `02.09.08`


robertmaynard:
I went with the `math(EXPR RAPIDS_WRITE_MINOR "${PROJECT_VERSION_MINOR} + 0" OUTPUT_FORMAT DECIMAL)` approach as it doesn't have issues with values like `00` which would get converted to an empty string.

testing/cmake/write_version-leading-zeroes/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
+#=============================================================================
+# Copyright (c) 2018-2021, NVIDIA CORPORATION.
```
dantegd, line 2: The most nitpicky nitpick: All these copyright header years should be only 2021 I think

testing/cmake/write_version-relative/CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ include(${rapids-cmake-dir}/cmake/write_version_file.cmake)
 
 cmake_minimum_required(VERSION 3.20)
 
-project(DEMO VERSION 3.2.1 LANGUAGES NONE)
+project(DEMO VERSION 3.2.0 LANGUAGES NONE)
```
dantegd, line 20: I think the only case not being tested is if this is 3.2, that the patch version would be 0

dantegd (approved):
Code looks great and testing it locally it seems to work very well! Just had two super minor questions

testing/cmake/write_version-leading-zeroes/CMakeLists.txt
robertmaynard: I followed the pattern that existed in cudf and raft where a copyright range existed. But at the minimum the copyright range is wrong ( should be 2021-2021? ), or do we change it next year to 2021-2022?

robertmaynard: https://github.com/rapidsai/rapids-cmake/issues/28

PR #33, opened 2021-06-24
rapids-cmake can now be correctly used by multiple adjacent directories

robertmaynard:
rapids-cmake can now be correctly used by multiple adjacent directories    
When in a situation where multiple projects are nested together in the following configuration:
```
- root
 - sub_project_1
 - sub_project_2    
```
If sub_project_1 doesn't import all the rapid-cmake components thatsub_project_2 requires it will cause weird runtime errors


robertmaynard:
@trxcllnt This should fix the issue you are seeing when multiple projects require `rapids-cmake`.

trxcllnt (approved):

PR #37, opened 2021-06-29
rapids_export version support expanded to handle more use-cases

robertmaynard:
The rapids_export command now supports the following features:

Explicit version argument option. This allows projects to specify a
version when calling `rapids_export` instead of having it inferred
from the `project()` call.

Disabling of version exporting by passing `VERSION OFF`. This
removes the generation of any version related information to allow
for projects that don't have API/ABI version matching requirements

Projects that only have a MAJOR version number now don't try
to match on the exact Major.Minor value but instead on Major only


robertmaynard:
rerun tests

ajschmidt8 (approved):

robertmaynard:
rerun tests

robertmaynard:
@gpucibot merge

PR #38, opened 2021-06-30
Bug/rapids handle existing cmake prefix path

robertmaynard:
When in a situation where multiple projects are brought in via CPM and are in sibling
directories, rapids-cmake previously would break.

This occured due to the fact that CPM would establish a local CMAKE_MODULE_PATH variable
that would override the cache variable that `rapids-cmake` would construct when going
into `sub_project_2`.

  - root
    - sub_project_1 [cpm]
      - uses rapids_cmake
    - sub_project_2 [cpm]
      - uses rapids_cmake

Now when `sub_project_1` exits it will properly propagate up the CMAKE_MODULE_PATH
information in a manner that sub_project_2 can see it.


robertmaynard:
@trxcllnt This time I think I fixed your bug

trxcllnt (approved):

PR #45, opened 2021-07-20
Introduce RAPIDS.cmake a better way to fetch rapids-cmake

robertmaynard:
The original approach of using FetchContent naively to acquire
rapids-cmake has a subtle bug, which can easily be resolved
by using the new and improved RAPIDS.cmake approach.

Now to acquire rapids-cmake projects can do the following:
```cmake
file(DOWNLOAD
    https://github.com/rapidsai/rapids-cmake/blob/branch-<Major>.<Minor>/RAPIDS.cmake
    ${CMAKE_BINARY_DIR})
include(${CMAKE_BINARY_DIR}/RAPIDS.cmake)
```

This will automatically fetch rapids-cmake if that hasn't happened
before. Otherwise unlike the old approach it will ensure that
rapids-cmake is in CMAKE_MODULE_PATH, solving a subtle bug.
When multiple adjacent projects included rapids-cmake via CPM they
would hit a bug where the second project couldn't find any rapids-cmake
modules.


ajschmidt8 (approved):

robertmaynard:
rerun tests

robertmaynard:
@gpucibot merge

PR #46, opened 2021-07-21
rapids-cmake tests properly state what C++ std levels they require

robertmaynard:


robertmaynard:
rerun tests

robertmaynard:
@gpucibot merge

Issue #2, opened 2021-05-07
[FEA] detect and clean out build dirs if building in a central location


trxcllnt:
Related to #1, sometimes project dependencies fail to build for various reasons. Doing a clean rebuild of the dependent project won't detect that one of its dependencies didn't build and clean-build it too, because clean builds of the dependent project don't (shouldn't?) clean out the dependency build cache.

It's kind of a hacky solution, but we have [a CMake function](https://github.com/rapidsai/node-rapids/blob/dfdc0a7349646ad1dfe3b543a2d7d694d779eb8e/modules/core/cmake/Modules/get_cpm.cmake#L90-L100) that checks whether the files for a dependency exist, and if not, deletes `${FETCHCONTENT_BASE_DIR}/${pkg}-build` and `${FETCHCONTENT_BASE_DIR}/${pkg}-subbuild` directories [before calling `CPMFindPackage`](https://github.com/rapidsai/node-rapids/blob/dfdc0a7349646ad1dfe3b543a2d7d694d779eb8e/modules/core/cmake/Modules/ConfigureCUDF.cmake#L21).

robertmaynard:
for `rapids-cmake` I expect the signature would look like:

```
rapids_cpm_clean_build_directory( 
   <ProjectName>
   SENTINEL_FILE [relative_path]
   )
```

PR #48, opened 2021-07-29
Tests now can be SERIAL and use FetchContent to get rapids-cmake

robertmaynard:
To properly test `cpm` packages in rapids-cmake we need to fix
two major issues.

 1. A way to have tests that depend on the same CPM package
    not execute at the same time. This is required so we don't
    double checkout a project

 2. A way to checkout and build projects such as RMM that themselves
    use rapids-cmake without issue.

Number #1 is solved by the introduction of the `SERIAL` keyword

Number #2 is solved for most tests by having the `project_template`
that `.cmake` tests use get `rapids-cmake` via FETCH_CONENT. This
will allow us to verify local rapids-cmake changes against existing
projects that use rapids-cmake

### DIFF
--- a/testing/utils/cmake_build_test.cmake
+++ b/testing/utils/cmake_build_test.cmake
@@ -18,5 +18,5 @@ include_guard(GLOBAL)
 include(utils/cmake_test.cmake)
 
 function(add_cmake_build_test source_or_dir)
-  add_cmake_test(BUILD "${source_or_dir}")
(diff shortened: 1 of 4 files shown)


robertmaynard:
@gpucibot merge

robertmaynard:
@gpucibot merge

robertmaynard:
@gpucibot merge

PR #51, opened 2021-07-30
rapids_cpm_find is more invariant as one would expect

### LINKED ISSUE #49: [BUG] `rapids_cpm_find` isn't as invariant as one would expect
**Describe the bug**

Consider the following reduced example of bringing in `rmm`  via CPM and as a dependency via `find_package`
```
    rapids_cpm_find(A)
        -> find_package(rmm)
    rapids_cpm_find(rmm INSTALL_EXPORT_SET B)
```
If package `A` sets up variables / state that are required to find `rmm`, the subsequent call to `rapids_cpm_find` will try and check out `rmm` and build it by source causing duplicate target errors. If `rapids_cpm_find(A)`  hadn't found a pre-built `rmm` we would have had no error as `CPM` is designed to handle that use case.

The current solution to this is the following:
```
    rapids_cpm_find(A)
        -> find_package(rmm)
   if(NOT TARGET rmm::rmm)
        rapids_cpm_find(rmm INSTALL_EXPORT_SET B)
   endif()
```    

This fixes the biggest issue ( duplicate targets ) but causes rmm not to be entered into the export set of `B`. This is problematic, as `A` isn't in the export set of `B` and therefore we now have dropped the dependency.

The correct logic would be:
```
    rapids_cpm_find(A)
        -> find_package(rmm)
   if(NOT TARGET rmm::rmm)
        rapids_cpm_find(rmm INSTALL_EXPORT_SET B)
   else()
        rapids_export_package(INSTALL rmm B)
   endif()
```    

**Desired behavior**
While we have a work around this requires users of rapids-cmake to understand all the permutations of behavior that is possible, the exact thing that rapids-cmake is trying to alleviate.

What should happen is that `rapids_cpm_find` should handle all the existence checks by looking at the `GLOBAL_TARGETS` argument. If any of those targets exist, it should record `INSTALL_EXPORT_SET` information and terminate. That way users get to write 'easy' code and get the correct behavior no matter what is happening

robertmaynard:
Fixes #49

rapids_cpm_find now looks at the `GLOBAL_TARGETS` entries and will not call CPMFindPackage if any of them already exist.

This makes it safe to call `rapids_cpm_find` no matter what existing targets already exist. therefore users of rapids-cmake don't need to do manual `if(TARGET ...)` checks, resulting in missing packages in the export set.

### DIFF
--- a/rapids-cmake/cpm/find.cmake
+++ b/rapids-cmake/cpm/find.cmake
@@ -44,16 +44,6 @@ consistency. List all targets used by your project in `GLOBAL_TARGET`.
 .. note::
   Requires :cmake:command:`rapids_cpm_init` to be called before usage
 
-.. note::
-  Adding an export set to :cmake:command:`rapids_cpm_find` has different behavior
-  for build and install. Build exports a respective CPM call, since
-  we presume other CPM packages don't generate a correct build directory
-  config module. While install exports a `find_dependency` call as
-  we expect projects to have a valid install setup.
-
-  If you need different behavior you will need to use :cmake:command:`rapids_export_package()`
-  or :cmake:command:`rapids_export_cpm()`.
-
 ``PackageName``
   Name of the package to find.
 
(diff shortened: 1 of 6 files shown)


robertmaynard:
Need to add tests

robertmaynard:
@gpucibot merge

robertmaynard:
@gpucibot merge

PR #52, opened 2021-07-30
Introduce rapids_cpm_<preset>

### LINKED ISSUE #32: [FEA] Add `rapids_cpm_<PackageName>` for common packages

Issue was first brought in discussion: https://github.com/rapidsai/rmm/pull/800#discussion_r651347909

The main ask is for a common shared formula for common dependencies of projects such as:

- Thrust
- GTest
- GBench
- SpdLog

The solution is to provide a limited number of high value packages in the `rapids/cpm` namespace. These packages would use the following template. This template allows for dependency tracking, but enforces a consistent version of dependencies across all consumers.

```
  rapids_cpm_<PackageName>( [BUILD_EXPORT_SET <export-name>]
                            [INSTALL_EXPORT_SET <export-name>]                  
                          )
```

The initial version will limit it self to only support GTest, GBench, SpdLog, and Thrust. Thrust will be added without the patches used by `cudf` initially and we can determine separately if those patches should be applied for all RAPIDS projects.

For packages that aren't as widely used the recommend approach is to continue to use `rapds_cpm_find`.

robertmaynard:
Requires https://github.com/rapidsai/rapids-cmake/pull/48 and https://github.com/rapidsai/rapids-cmake/pull/51

Adds rapids_cpm_<PackageName> for common packages

Fixes #32

The following packages are now are easier to user for RAPIDS projects
as rapids-cmake offers a pre-configured setup for each project.

-   GTest
-   NVBench
-   RMM
-   SpdLog
-   Thrust

On top of providing a consistent version of these packages to all RAPIDS projects, rapids-cmake now is able to deduce when these projects should also be installed.

```cmake

rapids_cpm_gtest(BUILD_EXPORT_SET myproject)
rapdis_cpm_rmm(BUILD_EXPORT_SET myproject
               INSTALL_EXPORT_SET myproject)
```
Given the above snippet when RMM is built as a subcomponent of `myproject` it will be installed as well. This is done since
RMM is part of the INSTALL export set, and therefore must be available from an installed version of `myproject`.


rapids-cmake/cpm/thrust.cmake
```diff
@@ -0,0 +1,78 @@
...
+function(rapids_cpm_thrust NAMESPACE namespaces_name)
+  list(APPEND CMAKE_MESSAGE_CONTEXT "rapids.cpm.thrust")
+
+  set(to_install OFF)
+  if(INSTALL_EXPORT_SET IN_LIST ARGN)
+    set(to_install ON)
+  endif()
+
+  include("${rapids-cmake-dir}/cpm/find.cmake")
+  rapids_cpm_find(Thrust 1.12.0 ${ARGN}
```
dantegd, line 66: It would be tricky to have a central config file to control the versions of non rapids branch naming projects, but it'd be useful to have a readme with a list of the dependencies and versions, particularly if we put more of them here, to be sure they are synced with integration, conda recipes, etc. What do you think?

robertmaynard: I agree

How about something like a json file in the cpm directory that listed all the project url and branch?

rapids-cmake and the docs could both reference it to maintain a single source of truth

dantegd: That sounds like a great solution!

robertmaynard: What I am thinking is something like:

```
{
  "GTest" : {
    "version" : "1.10",
    "git_url" : "https://github.com/google/googletest.git",
    "git_tag" : "release-{version}",
  },
  "nvbench" : {
    "version" : "0.0",
    "git_url" : "https://github.com/NVIDIA/nvbench.git",
    "git_tag" : "main",
  },
  "rmm" : {
    "version" : "rapids-cmake-version",
    "git_url" : "https://github.com/rapidsai/rmm.git",
    "git_tag" : "branch-{version}",
  },
  "spdlog" : {
    "version" : "1.8.5",
    "git_url" : "https://github.com/gabime/spdlog.git",
    "git_tag" : "v{version}",
  },
  "Thrust" : {
    "version" : "1.12.0",
    "git_url" : "https://github.com/NVIDIA/thrust.git",
    "git_tag" : "{version}",
  },
}

```

I am still trying to figure out how best to match `rmm` version to the current rapids calver, so for now I have used a placeholder ( rapids-cmake-version ).

Any thoughts?

robertmaynard: I have updated the PR with an initial implementation that has everything working based on a json file  ( https://github.com/rapidsai/rapids-cmake/pull/52/files#diff-695faffc4b8bc9bc2b2fb2f4c3958c279c2d01744eeead1a30dedd1ead81eb6c )

I still have to do the following:

- Reference the json versions in each packages documentation
- Formally document the versions.json file and what magic placeholders are supported

rapids-cmake/cpm/gtest.cmake
```diff
@@ -0,0 +1,81 @@
...
+  include("${rapids-cmake-dir}/cpm/detail/package_details.cmake")
+  rapids_cpm_package_details(GTest version repository tag)
+
+  include("${rapids-cmake-dir}/cpm/find.cmake")
+  rapids_cpm_find(GTest ${version} ${ARGN}
+                  GLOBAL_TARGETS GTest::gtest GTest::gmock GTest::gtest_main GTest::gmock_main
+                  CPM_ARGS
+                  GIT_REPOSITORY ${repository}
+                  GIT_TAG ${tag}
+                  GIT_SHALLOW TRUE
```
dantegd, line 67: won't `GIT_SHALLOW` shallow being `TRUE` interfere with the ability of setting any commit hash in the json file? Setting it to false would give us more flexibility, no?

dantegd (commented):
The JSON file approach looks really good! Had one question about the fetching

rapids-cmake/cpm/gtest.cmake
robertmaynard: Since it is always combined with a tag it will do a shallow checkout at the tip of that tag/branch.

robertmaynard: Ohh I see what you mean, any arbitrary `hash` value for the GIT_TAG.

Yes in those case we need a shallow equals false. I think the best approach is to extend the json spec to include 
`git_shallow` boolean field with the default being `true`

robertmaynard: I have updated the PR with the optional `git_shallow` field

dantegd (approved):
Looks great with the added field for the json file!

docs/packages/rapids_cpm_versions.rst
```diff
@@ -0,0 +1,63 @@
...
+
+    Supports the following placeholders:
+        - ``${rapids-cmake-version}`` will be evulated to 'major.minor' of the current rapids-cmake cal-ver value.
+        - ``${version}`` will be evulated to the contents of the ``version`` field.
+
+``git_shallow``
+
+    An optional boolean value that represents if we should do a shallow git clone
+    or not. If no such field exists the default is `git_shallow : true`
+
```
kkraus14, line 56: I think we need the ability to patch things as well, i.e. https://github.com/rapidsai/cudf/blob/branch-21.10/cpp/cmake/thirdparty/CUDF_GetThrust.cmake#L33

robertmaynard: I think that is worth doing as a followup Issue/Feature discussion and PR. 

As you are aware since cudf patches thrust it doesn't install it in the default location, which is what this change does. I am hesitant to install a modified Thrust in the 'default' location as that could cause confusion to users ( potential compile time or performance differences ). But once we resolve where patched versions of projects should be installed, this will be required.

robertmaynard:
rerun tests

robertmaynard:
@gpucibot merge

PR #54, opened 2021-08-04
rapids_find_package propagates variables from find_package

### LINKED ISSUE #53: [BUG] rapids_find_package needs to be a macro
**Describe the bug**

Currently `rapids_find_package` is a function and therefore doesn't propagate any variables set by the underlying
`find_package` call. This is problematic when projects want to use variables such as `<PROJECT_FOUND>` or `<PROJECT>_INCLUDE_DIRS`.

**Expected behavior**
Switching `rapids_find_package` over to being a macro will resolve the above issues. 

Since macros don't induce scope we will need to make sure the following occur:
 - All local variables are prefixed with `_rapids`, so we don't clobber any existing local variable
 - All local variables are unset at the end of the macro

** Notes **

Issue found when porting cudf to rapids-cmake

robertmaynard:
Fixes #53

The conversion of `rapids_find_package` to being a macro instead of a function allows it to properly propagate both common `find_package` variables and uncommon ones ( CUDAToolkit_LIBRARY_ROOT ).

### DIFF
--- a/rapids-cmake/find/package.cmake
+++ b/rapids-cmake/find/package.cmake
@@ -35,8 +35,9 @@ tracking of these dependencies for correct export support.
                         all normal find_package options ]
                       )
 
-Generate a :cmake:command:`find_package` call and associate this with the listed
-build and install export set for correct export generation.
+Invokes :cmake:command:`find_package` call and associate this with the listed
+build and install export set for correct export generation. Will propagate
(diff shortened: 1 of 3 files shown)


dantegd (approved):

kkraus14 (approved):

robertmaynard:
@gpucibot merge

PR #55, opened 2021-08-05
rapids_export(BUILD) only creates alias targets to existing targets

robertmaynard:
When generating the build directory `<project>-config.cmake` we only try to create alias targets to existing targets so we don't run into CMake errors.


robertmaynard:
@gpucibot merge

PR #57, opened 2021-08-06
rapids_export will always install cmake files to `lib`

### LINKED ISSUE #56: [BUG] Exported cmake files are in `lib64` instead of `lib`
**Describe the bug**
In at least the librmm and libcudf 21.08 packages the cmake config and target files are in the `lib64/cmake` folder whereas they were previously in the `lib/cmake` folder.  This differs from all other conda packages which also use the `lib/cmake` folder. I also believe it breaks `find_package` searches from working as expected.

See packages here: https://anaconda.org/rapidsai/librmm/files?version=21.08.00

**Expected behavior**
CMake config and target files should be in `lib/cmake`

robertmaynard:
Fixes #56

While using `GNUInstallDirs` is the technically correct way to
determine the correct install location for CMake based projects,
it has no concept of conda enviornments and incorrectly uses
the base distros layout instead.

In the very short term have `rapids-export` always export to `lib` as other
RAPIDS projects hardcode the install location to that

In the short term extend `rapids-cmake` to offer a function to compute
the library location which is conda aware. Update all of RAPIDS to
use that.

Medium term update CMake GNUInstallDirs to be conda aware so we
can just assume that works as expected.


robertmaynard:
@raydouglass @kkraus14

dantegd (approved):

rapids-cmake/export/export.cmake
```diff
@@ -154,7 +155,7 @@ function(rapids_export type project_name)
   string(TOLOWER ${project_name} project_name)
   string(TOUPPER ${project_name} project_name_uppercase)
   if(type STREQUAL "install")
-    set(install_location "${CMAKE_INSTALL_LIBDIR}/cmake/${project_name}")
+    set(install_location "lib/cmake/${project_name}")
```
kkraus14, line 158: Could we maybe do something like?
```
if(${CMAKE_INSTALL_PREFIX} EQUAL $ENV{CONDA_PREFIX})
    set(install_location "lib/cmake/${project_name}")
else()
    set(install_location "${CMAKE_INSTALL_LIBDIR}/cmake/${project_name}")
endif()
```

I hate to make things conda specific when we don't need to.

robertmaynard: I agree, I do want to stick with presuming always a conda environment. 

For `21.08` I want to stick with the minimal change required to fix conda, and I think the current change is the correct approach. While yours is better, we still have the issue that numerous RAPIDS packages hard code the install location of things like `libcudf` to always be `lib` instead of using to GNUInstallDirs. 

For `21.10` we should have a better holistic approach for RAPIDS. This would mean providing a function like `rapids_cmake_install_location` that would compute either `lib` or `lib64` based on CONDA, etc. That way not only is `rapids_export` fixed, but we can also update cudf, cugraph, cuml, ....

robertmaynard: Once we/I figure out a robust approach that we are doing a conda install I will upstream it to CMake

PR #61, opened 2021-08-09
Introduce rapids_cmake_install_lib_dir

### LINKED ISSUE #60: [FEA] introduce `rapids_cmake_install_lib_dir` to compute conda aware installation directory name

As seen by https://github.com/rapidsai/rapids-cmake/issues/56 and https://github.com/rapidsai/rapids-cmake/pull/57 rapids-cmake uses the host distro information to compute the lib directory name ( `lib` vs `lib64` ). This means that conda packages built on CentOS will incorrectly use `lib64`

rapids-cmake should provide the necessary functionality so that consuming projects can correctly compute the correct library location.

As brought up in #57 we need some general API that can be used by both `rapids_export` and users of cmake so that
the following. This function would determine if we are in a conda environment and use `lib` otherwise falling back to `GNUInstallDirs` for all other situations. This will allow RAPIDS projects to work correctly across multiple distros and package managers.

One proposal is to have `rapids_cmake_install_lib_dir(out_variable_name)` which would be used by consumers like this:
```
rapids_cmake_install_lib_dir(install_lib_dir)

install(TARGETS cugraph
        DESTINATION ${install_lib_dir}
        EXPORT cugraph-exports)
        
install(DIRECTORY include/cugraph/
        DESTINATION include/cugraph)

```

We could also provide an option flag to `rapids_cmake_install_lib_dir` that will allow callers to opt-in to modifying the `CMAKE_INSTALL_LIBDIR` value, but I don't want to make that implicit by default as that breaks the design rules of rapids-cmake

robertmaynard:
Fixes #60

rapids_cmake_install_lib_dir computes the correct lib direcotry (lib, lib64, ... ) with respect to how CONDA wants things done.

This is done by looking at `CMAKE_INSTALL_PREFIX` and seeing if it matches env{CONDA_PREFIX} or env{PREFIX} when env{CONDA_BUILD} is set. Under those conditions we will always return `lib` no matter the linux distro, under all other condtions `rapids_cmake_install_lib_dir` defers to `GNUInstallDirs`.

To make `rapids-cmake` consistent we also update `rapids_export` to use `rapids_cmake_install_lib_dir` to correctly compute the location where to install generated `<project>-config.cmake` files`.

### DIFF
--- a/cmake-format-rapids-cmake.json
+++ b/cmake-format-rapids-cmake.json
@@ -7,6 +7,12 @@
           "nargs": 1
         }
       },
+      "rapids_cmake_install_lib_dir": {
+        "pargs": {
+          "nargs": 1,
+          "flags": ["MODIFY_INSTALL_LIBDIR"]
+        }
+      },
       "rapids_cmake_make_global": {
         "pargs": {
           "nargs": 1
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -12,6 +12,7 @@ require.
    :titlesonly:
 
    /command/rapids_cmake_build_type
+   /command/rapids_cmake_install_lib_dir
    /command/rapids_cmake_make_global
    /command/rapids_cmake_parse_version
    /command/rapids_cmake_support_conda_env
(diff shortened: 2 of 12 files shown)


robertmaynard:
@gpucibot merge

PR #64, opened 2021-08-10
Introduce rapids_export_find_package_root command

robertmaynard:
The rapids_export_find_package_root allows projects to set
proper search paths for dependencies.

This is needed when a project either installs a dependency
in a non-standard place, or when a project has extra `rapids_export`
calls to export a dependency for build/install. Both of
these use-cases occur in cudf.


robertmaynard:
@gpucibot merge

PR #65, opened 2021-08-10
rapids_export_cpm(BUILD) captures location of locally found packages

robertmaynard:
When cpm find a local package via `<project>-config.cmake` we cache that location for our users

This fixes issues where a project will find a local install in `CMAKE_INSTALL_PREFIX` but consumers won't as they have a different `CMAKE_INSTALL_PREFIX`.


robertmaynard:
Needs a test

robertmaynard:
@gpucibot merge

PR #67, opened 2021-08-13
Copyright year range now matches when rapids-cmake existed

### LINKED ISSUE #28: [BUG] Correct Copyright version ranges
**Describe the bug**
The current version range is wrong as rapids-cmake didn't exist before 2021. We should update the range to be a single value for now.

robertmaynard:
Fixes #28


ajschmidt8 (approved):

robertmaynard:
@gpucibot merge

PR #69, opened 2021-08-23
[REVIEW] FIX Install sphinxcontrib-moderncmakedomain in docs script

dillon-cullinan:
Installs the pip only package `sphinxcontrib-moderncmakedomain` using the build script. This dependency is missing in our docker images but also has no conda equivalent currently so we aren't adding it to our docs conda meta package.

If this package is ever added to conda in the future this should be removed and added to `rapids-doc-env` meta.yaml here: https://github.com/rapidsai/integration/blob/branch-21.10/conda/recipes/rapids-doc-env/meta.yaml

### DIFF
--- a/ci/docs/build.sh
+++ b/ci/docs/build.sh
@@ -28,6 +28,10 @@ gpuci_logger "Activate conda env"
 . /opt/conda/etc/profile.d/conda.sh
 conda activate rapids
 
+# Using absolute path of python executable to avoid obtaining wrong version of package
+gpuci_logger "Install pip only package"
+/opt/conda/envs/rapids/bin/python -m pip install sphinxcontrib-moderncmakedomain
+
 gpuci_logger "Check versions"
 python --version
 $CC --version


ci/docs/build.sh
```diff
@@ -28,6 +28,10 @@ gpuci_logger "Activate conda env"
 . /opt/conda/etc/profile.d/conda.sh
 conda activate rapids
 
+# Using absolute path of python executable to avoid obtaining wrong version of package
+gpuci_logger "Install pip only package"
+/opt/conda/envs/rapids/bin/python -m pip install sphinxcontrib-moderncmakedomain
```
ajschmidt8, line 33:
```suggestion
pip install sphinxcontrib-moderncmakedomain
```

any reason this can't be just `pip install...`? seems like that should be fine since the conda environment is activated a few lines above.

dillon-cullinan: `which python` in the build script was pointing to a host level python executable and not the python executable inside the `rapids` environment. It caused the install to acquire a `py3.8` version of `sphinxcontrib-moderncmakedomain` inside a `python3.7` container.

dillon-cullinan:
CI failure looks like a problem on the gpuCI side. I don't believe this repo has a `python` build, just a `gpu` build.

ajschmidt8 (approved):

robertmaynard:
@dillon-cullinan Is this ready to be merged?

dillon-cullinan:
rerun tests

dillon-cullinan:
@gpucibot merge

PR #74, opened 2021-08-31
Allow projects to override version.json information

### LINKED ISSUE #73: [FEA] Allow projects to override version.json information

When building / testing a new feature that cross cuts multiple projects, or testing new project dependencies a developer might need to use a different branch / repo for a project offered by rapids_cmake.
 
Currently to have this work you need to do the following:

- Branch rapids-cmake
- modify versions.json
- modify rapids.cmake to point to your repo and branch
- modify the consuming project to fetch from your rapids-cmake fork

This is a really high bar to deploy / verify package changes, and rapids_cmake should offer a better approach.

We need a solution that only requires changes in the consuming project.

To that end I have thought about the following methods:

1. Provide hooks via `rapids_cpm_init` and `rapids_cpm_package_override` where a user defined json file can be specified which contains the project(s) new information. 

2. Provide just `rapids_cpm_package_override( <Package_Name> )` which has explicit `VERSION`, `GIT_URL`, `GIT_TAG`, ... arguments which take the user defined new values. 

The upside of number 1 is that the user facing code ( rapids_cpm_package_override ) doesn't have to change each time the version.json has new entries. The problem being that it is has a higher barrier of entry for consumers.

robertmaynard:
Fixes #73

Introduces `OVERRIDE` argument to rapids_cpm_init() and the rapids_cpm_package_override command.

Both of these allow consuming projects to inject new urls/repository/tag/etc for rapids-cmake cpm pre-configured packages

### DIFF
--- a/cmake-format-rapids-cmake.json
+++ b/cmake-format-rapids-cmake.json
@@ -55,7 +55,7 @@
           "nargs": 0
         },
         "kwargs": {
-          "OVERRIDE": 1,
+          "OVERRIDE": 1
         }
       },
       "rapids_cpm_package_override": {
--- a/cmake-format-rapids-cmake.json
+++ b/cmake-format-rapids-cmake.json
@@ -53,8 +53,17 @@
       "rapids_cpm_init": {
         "pargs": {
           "nargs": 0
+        },
+        "kwargs": {
+          "OVERRIDE": 1,
         }
       },
+      "rapids_cpm_package_override": {
+        "pargs": {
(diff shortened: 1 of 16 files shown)


robertmaynard:
@dantegd Would be okay using this design? Is the requirement of writing a json file to onerous?

robertmaynard:
Had offline signoff on the design.

robertmaynard:
@gpucibot merge

PR #75, opened 2021-09-07
rapids_export verify DOCUMENTATION and FINAL_CODE_BLOCK exist

robertmaynard:
When the user calls `rapids_export` with either of these parameters rapids-cmake will now validate that the variables exist, and error out when they don't


robertmaynard:
rerun tests

trxcllnt (approved):

robertmaynard:
@gpucibot merge

PR #77, opened 2021-09-08
Introduce rapids_cmake_write_git_revision_file

robertmaynard:
The rapids_cmake_write_git_revision_file command allows projects to generate a header file that captures the following relevant git information:

 - The current git branch name
 - If the source tree has any changes not committed
 - The latest git commit SHA1 and how 'far' we are from the latest
   project release

This information will allow project to encode this information into tests so they can see this information when looking at
CI or remote user failures.


robertmaynard:
@gpucibot merge

PR #79, opened 2021-09-14
rapids-cmake SHOULD_FAIL tests verify the CMake Error string

robertmaynard:
We now check the error message output from CMake to make sure we have the expected output on error cases. This makes sure that tests that validate error cases don't fail for other reasons like syntax errors.


robertmaynard:
@gpucibot merge

PR #81, opened 2021-09-16
cpm tests now download less components and can be run in parallel.

robertmaynard:
The cpm tests previously would each fetch at minimum the CPM module from github, and possibly other components. Now we pre-popluate a shared cache with each rapids-cmake package and the cpm module.

The end result of this is two fold. First we reduce the amount of network requests, and testing is quicker as we can run
more tests in parallel.


robertmaynard:
@gpucibot merge

PR #82, opened 2021-09-16
Remove unneeded inclusions of the old setup_cpm_cache.cmake

robertmaynard:
As part of https://github.com/rapidsai/rapids-cmake/pull/81 the `setup_cpm_cache.cmake` file was moved and refactored.

In that PR all callers of setup_cpm_cache() had been updated, but I missed a couple of tests that just included the file. This removes those unneeded inclusions.


robertmaynard:
@gpucibot merge

PR #83, opened 2021-09-16
rapids-cmake defaults to always download overriden packages

robertmaynard:
When a user of rapids-cmake explicitly overrides a pre-configured
package we should download the override version to have
the correct version by default.

For example if a developer specifies a custom rmm branch to test,
we don't want to find the same version of rmm that already exists
on the system as that defeats the override.

To fully support this behavior the rapids-cmake version.json
now supports the search_locally flag which can be used to
control this behavior when users do want to search locally
when using a overriden package.


robertmaynard:
rerun tests

robertmaynard:
@gpucibot merge

PR #98, opened 2021-10-15
rapids_cpm_thrust installs to a location that won't be marked system

robertmaynard:
Projects such as cudf, and rmm require a newer versions of thrust than can
be found in the oldest supported CUDA toolkit.This requires these components to
install/packaged so that consumers use the same version. To make sure that the
custom version of thrust is used over the CUDA toolkit version we need to ensure
we always use an user include and not a system.

By default if we allow thrust to install into `CMAKE_INSTALL_INCLUDEDIR` alongside
rmm (or other pacakges) we will get a install tree that looks like this:

```
  install/include/rmm
  install/include/cub
  install/include/thrust
```

This is a problem for CMake+NVCC due to the rules around import targets, and
user/system includes. In this case both rmm and thrust will specify an
include path of `install/include`, while thrust tries to mark it as an user include,
since rmm uses CMake's default of system include. Compilers when provided the same include
as both user and system always goes with system.

Now while rmm could also mark `install/include` as system this just pushes the issue to
another dependency which isn't built by RAPIDS and comes by and marks `install/include` as
system.

Instead the more reliable option is to make sure that we get thrust to be placed in an
unique include path that to other project will use. In the case of rapids-cmake we install
the headers to `include/rapids/thrust`.


robertmaynard:
> Do you mean 1.14?

Yes I meant 1.14

PR #110, opened 2021-10-26
Build directory config.cmake now sets the correct targets to global

robertmaynard:
The build directory config file will construct the same targets as the install directory ( <namespace>::<target> ) and the previous logic would try to promote <target> instead of <namespace>::<target>

For backwards compat, the build directory will also try to promote `<target>` in-case people relied on that as a side-effect.

Thanks to @trxcllnt for reporting this issue.


trxcllnt (approved):

robertmaynard:
@gpucibot merge

PR #118, opened 2021-11-10
Add `PREFIX` option to write_version_file / write_git_revision_file

### LINKED ISSUE #99: [FEA] Add `PREFIX` option to `write_version_file` / `write_git_revision_file`

NVBench's project name is `NVBench`, so all of the rapids-cmake preprocessor macros currently start with that string. However, we conventionally use all-caps `NVBENCH` to prefix preprocessor macros.

A `PREFIX <string>` option to the `write_version_file` / `write_git_revision_file` function that defaults to the project name, but allows customization.

Renaming my project to NVBENCH is an alternative, but this would be burdensome and disruptive to adoption.

**Additional context**

Documentation should state that the default is the project name.

robertmaynard:
Fixes #99


testing/cmake/write_git_revision-custom-prefix/main.cpp
```diff
@@ -0,0 +1,47 @@
...
+constexpr const char* nvers = DEMO_GIT_VERSION;
+#if defined(NESTED_GIT_IS_DIRTY)
+  constexpr const bool nisdirty = true;
+#else
+  constexpr const bool nisdirty = false;
+#endif
+
+int main()
+{
+  static_assert(dbranch == nbranch);
```
allisonvacanti, line 41: These are comparing c-string pointers, not the string contents.

robertmaynard: Yeap, it is abusing the fact that since these are constexpr the compiler can evaluate if they point to the same static string.

allisonvacanti (approved):

robertmaynard:
rerun tests

robertmaynard:
@gpucibot merge

PR #120, opened 2021-11-10
rapids_cpm_thrust doesn't place temp file in a searched location

robertmaynard:
This makes sure that `find_package(Thrust)` which searches a build directory doesn't accidentally use files meant for installation.


trxcllnt (approved):

robertmaynard:
@gpucibot merge

PR #121, opened 2021-11-11
rapids_cpm_find restores CPM variables when project was already added

### LINKED ISSUE #117: [BUG] `rapids_cpm_find` avoid package_needs_to_be_added check before `CPMFindPackage`
The description of `rapids_cpm_find` makes it seem like it's nearly a drop in replacement for `CPMFindPackage` that adds functionality for tracking global dependencies and setting up correct exports. However, the function subtly changes the behavior of `CPMFindPackage` depending on whether this function has already been called, and overrides CPM functionality. This can lead to bugs that are difficult to diagnose if a user expects every call to `rapids_cpm_find` to function similarly to `CPMFindPackage`. 

The problem lies in this section of the function:
```
  if(package_needs_to_be_added)
    if(CPM_${name}_SOURCE)
      CPMAddPackage(NAME ${name} VERSION ${version} ${RAPIDS_UNPARSED_ARGUMENTS})
    else()
      CPMFindPackage(NAME ${name} VERSION ${version} ${RAPIDS_UNPARSED_ARGUMENTS})
    endif()
  endif()
```
If the user needs some variables set by `CPMFindPackage`, they may or may not be set depending on whether `rapids_cpm_find` has already been called or not. Consider this scenario, partially pulled from cuDF, where two dependencies each call `rapids_cpm_find` with the same arguments:
```
function(find_and_configure_libcudacxx VERSION)
    rapids_cpm_find(libcudacxx ${VERSION} ...)

    set(LIBCUDACXX_INCLUDE_DIR "${libcudacxx_SOURCE_DIR}/include" PARENT_SCOPE)
endfunction()

function(configure_cudf VERSION)
    find_and_configure_libcudacxx(VERSION)

    # This next line may fail!!!!
    file(COPY "${LIBCUDACXX_INCLUDE_DIR}/" DESTINATION "${CMAKE_BINARY_DIR}/include/libcudacxx")
endfunction()

function(configure_matx VERSION)
    rapids_cpm_find(libcudacxx ${VERSION} ...)
endfunction()
```
Everything works smoothly if the configure functions are called in this order:
```
configure_cudf(...)
configure_matx(...)
````
However, they fail if called in this order:
```
configure_matx(...)
configure_cudf(...) # ERROR: Cannot copy /include !!!
```
An error will be generated much later in the code because `libcudacxx_SOURCE_DIR` was never set in the second function call. I recently ran into this issue and it was difficult to debug what was happening because the error message and the cause of the error were very far apart and only showed up depending on the order some third party dependencies were added.

To fix this, I think any function that wraps a CPM function should:
1. Always call the wrapped CPM function
   a. CPM already has mechanisms to handle duplicate calls while still providing the same output
   b. Calling `CPMFindPackage` regardless of the value of `package_needs_to_be_added` ensures the correct variables are set with very little downsides
2. Do not call a different CPM function than the wrapped function
   a. i.e. `rapids_cpm_find` should call `CPMFindPackage` and never `CPMAddPackage` if `CPM_${name}_SOURCE` is set
   4. CPM already has mechanisms to call `CPMFindPackage` from `CPMAddPackage` and `CPMAddPackage` from `CPMFindPackage`. This behavior will only complicates or hides that behavior
3. Provide separate wrappers for each wrapped CPM function
   a. i.e. There should be a `rapids_cpm_find` to wrap `CPMFindPackage` and `rapids_cpm_add` to wrap `CPMAddPackage`
   b. This also helps with dependencies which always need to be added, even if they are found. Its not possible to use `rapids_cpm_find` in this scenario. (GLOG is an example of this. You must build from source if you want a custom log format)

In my opinion, these changes would make the functionality of `rapids_cpm_XXX` functions more clear, lead to less bugs, and make it easier to maintain the library for future updates to CPM.

robertmaynard:
Now when rapdis_cpm_find is used to find an existing project that was introduced via CPM we correctly restore CPM variables.

Fixes #117

### DIFF
--- a/rapids-cmake/cpm/find.cmake
+++ b/rapids-cmake/cpm/find.cmake
@@ -151,6 +151,12 @@ function(rapids_cpm_find name version)
     else()
       CPMFindPackage(NAME ${name} VERSION ${version} ${RAPIDS_UNPARSED_ARGUMENTS})
     endif()
+  else()
+    # Restore any CPM variables that might be cached
+    cpm_check_if_package_already_added(${name} ${version})
+    if(CPM_PACKAGE_ALREADY_ADDED)
+      cpm_export_variables(${name})
+    endif()
   endif()
 
   set(extra_info)
--- a/testing/cpm/CMakeLists.txt
+++ b/testing/cpm/CMakeLists.txt
@@ -21,6 +21,7 @@ add_cmake_config_test( cpm_find-existing-build-dir )
 add_cmake_config_test( cpm_find-existing-target )
 add_cmake_config_test( cpm_find-existing-target-to-export-sets )
 add_cmake_config_test( cpm_find-options-escaped )
+add_cmake_config_test( cpm_find-restore-cpm-vars )
 
 add_cmake_config_test( cpm_init-bad-override-path.cmake SHOULD_FAIL "rapids_cpm_package_override can't load")
 add_cmake_config_test( cpm_init-override-multiple.cmake )
--- a/testing/cpm/cpm_find-restore-cpm-vars/CMakeLists.txt
+++ b/testing/cpm/cpm_find-restore-cpm-vars/CMakeLists.txt
@@ -1,0 +1,72 @@
+#=============================================================================
+# Copyright (c) 2021, NVIDIA CORPORATION.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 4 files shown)


trxcllnt (approved):

robertmaynard:
@gpucibot merge

PR #122, opened 2021-11-11
rapids_cpm_libcudacxx ensures CMAKE_INSTALL_INCLUDEDIR exists

robertmaynard:
Instead of relying on the user having called GNUInstallDirs


robertmaynard:
@gpucibot merge

PR #124, opened 2021-11-11
rapids_cpm_libcudacxx install logic is safe for multiple inclusion

robertmaynard:
Now we only construct the libcudacxx_includes target when it already doesn't exist.

This resolves a bug where `find_package(libcudacxx)` was called multiple times ( once via CPM, once via `find_dependency` ) resulting in the same target trying to be created twice, and failing.


robertmaynard:
rerun tests

robertmaynard:
@gpucibot merge

PR #130, opened 2021-11-12
rapids_cpm_libcudacxx handle CPM already finding libcudacxx before being called

robertmaynard:
Consider the following example CMake example:

```cmake
CPMFindPackage(libcudacxx ..... )
rapids_cpm_libcudacxx()
target_link_libraries(... PRIVATE libcudacxx::libcudacxx)
```

The first line will check out libcudacxx from source. This will result in the second line not checking out libcudacxx, and therefore not creating the `libcudacxx::libcudacxx` target, causing downstream failures.

On first pass this seems to be an impossible state to get into as all RAPIDS projects should be using `rapids_cpm_libcudacxx`. The problem was that before this PR, rapids-cmake would generate the above `CPMFindPackage` call when `rapids_cpm_libcudacxx` was called with a BUILD_EXPORT_SET.

### DIFF
--- a/rapids-cmake/cpm/libcudacxx.cmake
+++ b/rapids-cmake/cpm/libcudacxx.cmake
@@ -56,15 +56,10 @@ Result Variables
 function(rapids_cpm_libcudacxx)
   list(APPEND CMAKE_MESSAGE_CONTEXT "rapids.cpm.libcudacxx")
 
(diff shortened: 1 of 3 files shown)


trxcllnt (approved):

robertmaynard:
@gpucibot merge

PR #135, opened 2021-12-07
Set the `always_download` value in versions.json to the common case

robertmaynard:
It is vastly more common for `always_download` to be set to true for any override project. Since by setting `always_download` to true we turn off searching for a local copy. 

By providing an example that set it to false, we caused issues when people copied the example and ran into bugs that are hard to diagnose.


harrism (approved):

robertmaynard:
@gpucibot merge

PR #136, opened 2021-12-08
Remove rapids_cmake_install_lib_dir unstable side effect checks

robertmaynard:
Remove checks around the `CMAKE_INSTALL_LIBDIR` cache state as that isn't easy to test across CMake versions. 

In addition `rapids_cmake_install_lib_dir` never states anything if this variable would exist or not exist in the cache after call, since it isn't part of the contract of the API we can remove the test.


robertmaynard:
@gpucibot merge

PR #137, opened 2021-12-09
add exclude_from_all flag to version.json

robertmaynard:
Allow overrides of rapids-cmake cpm projects to specify if the project should be excluded from the global `all` CMake build project via the `exclude_from_all` property.


robertmaynard:
@gpucibot merge

robertmaynard:
@gpucibot merge

PR #138, opened 2021-12-14
Update rapids-cmake packages to Thrust 1.15

robertmaynard:
Starting with 22.02, all of RAPIDS is using Thrust 1.15

### DIFF
--- a/rapids-cmake/cpm/versions.json
+++ b/rapids-cmake/cpm/versions.json
@@ -23,7 +23,7 @@
       "git_tag" : "v${version}"
     },
     "Thrust" : {
-      "version" : "1.12.0",
+      "version" : "1.15.0",
       "git_url" : "https://github.com/NVIDIA/thrust.git",
       "git_tag" : "${version}"
     },


robertmaynard:
Need to manually verify the following projects cleanly build with 1.15:

- [ ] rmm
- [ ] raft
- [ ] cudf
- [ ] cuml
- [ ] cugraph

robertmaynard:
rerun tests

robertmaynard:
@gpucibot merge

gnaggnoyil:
AFAIK thrust1.15 will not report an exact version match when executing `find_package`, if the version string passed to `find_package` is not exactly the string "1.15.0.0" instead of a simple "1.15.0". However the current `version.json` specifies thrust version as "1.15.0". Therefore when `CPMFindPackage` tries to find a local thrust package of version 1.15 **exactly** through `find_package`, the find process will fail even if there is a thrust 1.15 package in the find location.

I think the above thrust issue has been fixed since thrust 1.16, according to [this PR](https://github.com/NVIDIA/thrust/pull/1599). By far I think a workaroud that replaces `"version" : "1.15.0",` to `"version" : "1.15.0.0",` would be enough. @robertmaynard do you have any thoughts?

robertmaynard:
> AFAIK thrust1.15 will not report an exact version match when executing `find_package`, if the version string passed to `find_package` is not exactly the string "1.15.0.0" instead of a simple "1.15.0". However the current `version.json` specifies thrust version as "1.15.0". Therefore when `CPMFindPackage` tries to find a local thrust package of version 1.15 **exactly** through `find_package`, the find process will fail even if there is a thrust 1.15 package in the find location.
> 
> I think the above thrust issue has been fixed since thrust 1.16, according to [this PR](https://github.com/NVIDIA/thrust/pull/1599). By far I think a workaroud that replaces `"version" : "1.15.0",` to `"version" : "1.15.0.0",` would be enough. @robertmaynard do you have any thoughts?

Yes that is an issue, see PR https://github.com/rapidsai/rapids-cmake/pull/180 which fixes it

PR #140, opened 2021-12-17
Fix type-o in docs `<PackageName>_BINARY_DIR` instead of `<PackageName>_BINAR_DIR`

dagardner-nv:
I found this type-o while reading the docs at https://docs.rapids.ai/api/rapids-cmake/stable/command/rapids_cpm_find.html#result-variables


GPUtester:
Can one of the admins verify this patch?

ajschmidt8:
ok to test

robertmaynard (approved):

robertmaynard:
@gpucibot merge

PR #144, opened 2022-01-10
Update NVBench for 22.02 to be the latest version

robertmaynard:
Now that NVBench has correct the previous CMake issues, we can move to a newer version.


robertmaynard:
@gpucibot merge

PR #145, opened 2022-01-11
rapids-cmake now supports an empty package entry in the override file

robertmaynard:
This allows projects to have a placeholder file that they add/remove entries as needed instead of having to remove the file when the have no current overrides.


robertmaynard:
@gpucibot merge

PR #154, opened 2022-01-25
Extend rapids_export to support the concept of optional COMPONENTS

robertmaynard:
rapids_export via the COMPONENTS and COMPONENTS_EXPORT_SET commands can now generate find modules that have optional components.


vyasr:
I assume this is slipping to 22.12 right?

robertmaynard:
> I assume this is slipping to 22.12 right?

Yeap

robertmaynard:
https://github.com/rapidsai/raft/pull/1154 is now passing with the latest updates to the `COMPONENT` design. Ready for review

vyasr:
@robertmaynard looks like this completely slipped out of my sight. Should I review this now?

robertmaynard:
> @robertmaynard looks like this completely slipped out of my sight. Should I review this now?

Yes please

rapids-cmake/export/export.cmake
```diff
@@ -187,6 +225,34 @@ function(rapids_export type project_name)
...
+      string(REGEX REPLACE "(${project_name}[-_])|([-_]?${comp}[-_]?)|([-_]?export[s]?)" ""
+                           nice_export_name "${comp_export_set}")
+      if(nice_export_name STREQUAL "")
+        string(PREPEND nice_export_name "${comp}")
+      else()
+        string(PREPEND nice_export_name "${comp}-")
+      endif()
+      rapids_export_component(${type} ${project_name} ${comp} ${comp_export_set}
+                              ${nice_export_name} ${_RAPIDS_PROJECT_NAMESPACE})
+      list(APPEND _RAPIDS_COMPONENTS_EXPORT_SET_NICE_NAMES ${nice_export_name})
```
vyasr, line 252: What is the purpose of tracking the nice names? If it is meant to be user-facing, we should name it without the underscore prefix and add it to the docs. If it is private, it seems like it doesn't actually serve any purpose.

rapids-cmake/export/export.cmake
```diff
@@ -187,6 +225,34 @@ function(rapids_export type project_name)
...
+    )
+  endif()
+
+  if(_RAPIDS_COMPONENTS AND _RAPIDS_COMPONENTS_EXPORT_SET)
+    include("${rapids-cmake-dir}/export/detail/component.cmake")
+    set(_RAPIDS_HAS_COMPONENTS TRUE)
+    set(_RAPIDS_COMPONENTS_EXPORT_SET_NICE_NAMES)
+
+    foreach(comp comp_export_set IN ZIP_LISTS _RAPIDS_COMPONENTS _RAPIDS_COMPONENTS_EXPORT_SET)
+      string(REGEX REPLACE "(${project_name}[-_])|([-_]?${comp}[-_]?)|([-_]?export[s]?)" ""
```
vyasr, line 243: Should we document this transformation (depending on whether the nice names are intended to be public as I asked below)?

rapids-cmake/export/detail/component.cmake
```diff
@@ -0,0 +1,80 @@
...
+
+  rapids_export_component( (BUILD|INSTALL) <project_name> <component_name> <export_set> <unique_name> <namespace>)
+
+The :cmake:command:`rapids_export_component` function generates
+the `<proj>-<unique_name>-targets.cmake` and `<proj>-<unique_name>-dependencies.cmake`
+files necessary for :cmake:command:`rapids_export` to support optional
+components.
+
+.. note::
+  It is an anti-pattern to use this function directly. The vast majority of projects
```
vyasr, line 37: Are there any real use cases of invoking this function directly? Since it's a detail function I default to assuming it should never need to be used. If there are niche use cases maybe it should be moved out of detail and instead documented as advanced a la `rapids_export_package` or `rapids_export_find_package_file`.

rapids-cmake/export/detail/component.cmake
```diff
@@ -0,0 +1,80 @@
...
+  would use :cmake:command:`rapids_export`
+
+
+#]=======================================================================]
+# cmake-lint: disable=W0105,R0913
+function(rapids_export_component type project_name component_name export_set unique_name namespace)
```
vyasr, line 43: For me to learn, when do you prefer `cmake_parse_arguments` to just writing out the function signature? AFAICT you use the simpler approach when you only have a bunch of single value arguments with no multi value or options?

rapids-cmake/export/detail/component.cmake
```diff
@@ -0,0 +1,80 @@
...
+  list(APPEND CMAKE_MESSAGE_CONTEXT "rapids.export.rapids_export_component")
+
+  string(TOLOWER ${type} type)
```
vyasr, line 46: Does rapids-cmake actually support these case-insensitively? I've never tried using anything other than all upper/lowercase and sort of assumed that was required by design to ensure consistent style.

rapids-cmake/export/detail/component.cmake
```diff
@@ -0,0 +1,80 @@
...
+  else()
+    set(install_location "${PROJECT_BINARY_DIR}")
+    set(deps_destination "${install_location}/")
+
+    export(EXPORT ${export_set} NAMESPACE ${namespace}
+           FILE "${install_location}/${project_name}-${unique_name}-targets.cmake")
+
+  endif()
+
+  if(TARGET rapids_export_${type}_${export_set})
```
vyasr, line 74: Minor nit, this could get out of sync if we updated the behavior of rapids_export for any reason. I suspect trying to centralize would be a lot of boilerplate for very little benefit though, so this is probably just something we're better off remembering.

vyasr (changes requested):

rapids-cmake/export/export.cmake
robertmaynard: The nice name is used when generating the files names that the component information will be written too.

The nice names don't need to be known by users, but are generated so that the installed files all have 'reasonable' names

rapids-cmake/export/export.cmake
robertmaynard: I don't think we should since it could change and I don't want people to assume certain transformation

rapids-cmake/export/detail/component.cmake
robertmaynard: There are no real uses cases of invoking this function directly. I am happy to drop that comment

rapids-cmake/export/detail/component.cmake
robertmaynard: Yes rapids-cmake should handle case-insensitively. Everything gets transformed into lowercase internally

rapids-cmake/export/detail/component.cmake
```diff
@@ -0,0 +1,80 @@
...
+            FILE ${project_name}-${unique_name}-targets.cmake
+            NAMESPACE ${namespace}
+            DESTINATION "${install_location}"
+            COMPONENT ${component_name})
+
+  else()
+    set(install_location "${PROJECT_BINARY_DIR}")
+    set(deps_destination "${install_location}/")
+
+    export(EXPORT ${export_set} NAMESPACE ${namespace}
```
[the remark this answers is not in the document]
robertmaynard: Correct. The components are loaded in both cases by the primary `<project>-config.cmake` which assumes the components files are located in the same directory as the `config.cmake`.

It is slightly more complicated as we verify the requested component is valid via an allow list, and do some pattern matching globbing to determine what files to load.

rapids-cmake/export/detail/component.cmake
robertmaynard: Where possible I prefer to avoid `cmake_parse_arguments` for `detail/` methods so that we can keep the complexity of those down.

In general I try to only use `cmake_parse_arguments` when we have optional arguments or multiple arguments that take multi-values. This way we get nicer error messages when calling functions without the required arguments.

rapids-cmake/export/detail/component.cmake
robertmaynard: Upside is that if we change the name of `rapids_export_${type}_${export_set}` multiple tests will fail as they try read the contents of the `${project_name}-${unique_name}-dependencies.cmake` file

rapids-cmake/export/detail/component.cmake
vyasr: Yes please do.

rapids-cmake/export/export.cmake
vyasr: Why do we need this list, then? Can't we simply remove it altogether?

testing/export/export-verify-build-components-dependencies/CMakeLists.txt
```diff
@@ -0,0 +1,78 @@
...
+rapids_export(BUILD FakEProJecT
+  EXPORT_SET fake_set
+  COMPONENTS c1
+  COMPONENTS_EXPORT_SET fake_set_c1_export
+  LANGUAGES CXX
+  NAMESPACE test::
+  )
+
+# Add a custom command that generates a second project
+# that verifies our component targetd are exported correctly
```
vyasr, line 39:
```suggestion
# that verifies our component targets are exported correctly
```

testing/export/export-verify-build-components-dependencies/CMakeLists.txt
```diff
@@ -0,0 +1,78 @@
...
+file(WRITE "${CMAKE_BINARY_DIR}/verify-1/CMakeLists.txt" [=[
+cmake_minimum_required(VERSION 3.20)
+project(verify_build_targets LANGUAGES CXX)
+set(nice_names  c1-fake_set)
```
vyasr, line 43:
```suggestion
set(nice_names c1-fake_set)
```

testing/export/export-verify-build-components-dependencies/CMakeLists.txt
```diff
@@ -0,0 +1,78 @@
...
+foreach(nice_name IN ZIP_LISTS components export_sets)
+  message(STATUS "${CMAKE_CURRENT_LIST_DIR}/../fakeproject-${nice_names}-dependencies.cmake")
```
vyasr, line 45: Minor nit, this message seems unnecessary since all the error cases will fail loudly and provide sufficient information.

testing/export/export-verify-build-components-dependencies/CMakeLists.txt
```diff
@@ -0,0 +1,78 @@
...
+endforeach()
+]=])
+
+add_custom_target(verify_target_dependencies_files ALL
+  COMMAND ${CMAKE_COMMAND} -E rm -rf "${CMAKE_BINARY_DIR}/verify-1/build"
+  COMMAND ${CMAKE_COMMAND} -S="${CMAKE_BINARY_DIR}/verify-1" -B="${CMAKE_BINARY_DIR}/verify-1/build"
+  )
+
+# Add a custom command that generates a second project
+# that verifies our component targetd are exported correctly
```
vyasr, line 63:
```suggestion
# that verifies our component targets are exported correctly
```

vyasr: I think this PR just needs a sed command to replace `targetd` with `targets`.

vyasr (approved):
Some questions and stylistic changes, and a couple of threads from my last review remain open, but everything that's left is very minor.

rapids-cmake/export/export.cmake
robertmaynard: Yes it can be removed. Just realized it was left-over from a previous implementation of the `config.cmake`

PR #156, opened 2022-01-28
Track package versions to the generated `find_dependency` calls

robertmaynard:
When a user uses `rapids_find_package(<pkg> <version>)` or `rapids_cpm_find` we now record the version requested. This allows rapids-cmake to ensure the generated `find_dependency` have the same minimum version required as the project itself.

Note we store the requested version, not the actual version found. This is important since if we used the version found we would break things such as CUDAToolkit backwards compat ( project builds against 11.5, consumers can safely use 11.3 ).


robertmaynard:
@gpucibot merge

PR #159, opened 2022-02-23
Docs use intersphinx correctly to link to CMake command docs

### LINKED ISSUE #141: [DOC] Link Sphinx `ref`s to official CMake Docs
I've been reading through the `rapids-cmake` docs and I noticed that we use Sphinx [refs](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#ref-role) in places like [here](https://github.com/rapidsai/rapids-cmake/blob/e430e4cf08062d312c434497e31bb4ed8dad61d5/rapids-cmake/cmake/support_conda_env.cmake#L39). I think the objective of the `ref` is to link to the official, external CMake docs, but it doesn't seem to be working currently. I think we need to configure our [intersphinx_mapping variable](https://github.com/rapidsai/rapids-cmake/blob/e430e4cf08062d312c434497e31bb4ed8dad61d5/docs/conf.py#L121) to [something like this](https://gitlab.kitware.com/cmake/cmake/-/issues/21610#note_875068) in order for the external links to work correctly.

Here's the official docs on the `intersphinx_mapping` variable for reference: https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html#confval-intersphinx_mapping

I can try to tackle this after the holidays if no one gets to it before then.

robertmaynard:
Fixes #141

### DIFF
--- a/rapids-cmake/cpm/find.cmake
+++ b/rapids-cmake/cpm/find.cmake
@@ -66,7 +66,7 @@ consistency. List all targets used by your project in `GLOBAL_TARGET`.
   our build directory export set.
 
 ``INSTALL_EXPORT_SET``
-  Record a :cmake:command:`find_dependency(<PackageName> ...)` call needs to occur as part of
+  Record a :cmake:command:`find_dependency(<PackageName> ...) <cmake:module:CMakeFindDependencyMacro>` call needs to occur as part of
   our install directory export set.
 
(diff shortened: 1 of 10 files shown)


bdice (approved):
👍

robertmaynard:
@gpucibot merge

PR #160, opened 2022-02-25
Tracked package versions are now not required, but preferred.

robertmaynard:
Instead of generating `find_dependency(<pkg> <version>)` we
now generate:
  find_package(<pkg> <version>)
  find_dependency(<pkg>)

This allows us to search for the versioned type, but if it
doesn't exist fall back to an unversioned one. Since
find_dependency propagates the `REQUIRED` keyword we need
to use `find_package` to properly handle this optional search.

This is needed since `rapids_cpm_find` always records a version
value, but the project might not actually generate the correct
CMake infrastructure to support versioned `find_package` calls

### DIFF
--- a/rapids-cmake/export/package.cmake
+++ b/rapids-cmake/export/package.cmake
@@ -72,10 +72,17 @@ function(rapids_export_package type name export_set)
     endif()
   endif()
 
-  set(version ${RAPIDS_VERSION})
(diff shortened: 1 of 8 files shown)


dantegd (approved):
Tested locally and it seems to work perfectly!

robertmaynard:
@gpucibot merge

PR #161, opened 2022-02-25
CMAKE_CUDA_ARCHITECTURES `ALL` will not insert 62 or 72

robertmaynard:
The logic to determine if we should generate for NVIDIA Drive integrated GPUs was incorrect and could cause bad
code generation for non drive based ARM platforms.

Since RAPIDS doesn't deploy multi-arch projects to DRIVE we can remove this logic.


robertmaynard:
@gpucibot merge

PR #168, opened 2022-03-18
rapids_cpm_gtest will properly find GTest 1.10 packages

robertmaynard:
Previously if CMake < 3.23 was installed the provided FindGTest
would be used which doesn't provide the GMock targets.

In addition in those cases, since we pass EXACT we need to ensure we also exactly match the patch version of GTest.


robertmaynard:
@gpucibot merge

PR #174, opened 2022-03-23
rapids_export now handles explicit version values of 0 correctly

robertmaynard:
The previous logic didn't work as CMake evaluates `0` as FALSE when it comes to `if` statements, and therefore we didn't
build the correct version strings when a value was just `0`.


robertmaynard:
@trxcllnt This should resolve the issue with cuspatial at GTest

trxcllnt (approved):

trxcllnt:
@gpucibot merge

PR #178, opened 2022-03-25
Add more exhaustive set of tests are version values of 0

robertmaynard:
It is common to have a version value of `0` for example in `1.0.1`. But due to how CMake treats `0` as a negative boolean value this can cause logic issues if a conditional checks a variable that holds a version component ( major / minor / patch ).

To ensure that rapids-cmake doesn't introduce regressions around handling `0` as a version component value, this PR expands the test suite to handle more zero value cases.


robertmaynard:
@gpucibot merge

PR #180, opened 2022-04-13
rapids_cpm_thrust: Correctly find version 1.15.0

robertmaynard:
Since Thrust 1.15.0 reports it self as `1.15.0.0` we need to also search with the patch version added.

Starting in Thrust 1.16, the tweak version can be dropped.


gnaggnoyil:
LGTM

robertmaynard:
@gpucibot merge

PR #182, opened 2022-04-14
put $PREFIX before $BUILD_PREFIX in conda build

### LINKED ISSUE #179: [BUG] Conda recipes that have separate `build` and `host` sections don't find packages correctly using `rapids_cmake_support_conda_env`
**Describe the bug**

Conda recipes use the `requirements` section to define build time and runtime dependencies of a project. WIthin the `requirements` sections, there's three subsections `build`, `host`, and `run`. `build` is supposed to contain native tools used for building like `cmake`, `make`, `ninja`, compilers, etc. `host` contains libraries for the target platform of the project (for example to allow cross compiling) typically used for linking. Read more here: https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#requirements-section

It isn't well documented and is likely legacy, but there's also the possibility of running with only two sections a `build` and `run` section, in which the build tools and target platform library dependencies are merged into a single section. Most of the RAPIDS conda recipes are defined in this way, likely due to legacy reasons.

The `build` and `host` sections map to different conda environments during a conda build and have environment variables of `BUILD_PREFIX` and `PREFIX` set accordingly. Where there's only a `build` section in the recipe there can still be a corresponding `PREFIX` environment created alongside the `BUILD_PREFIX` environment depending on the packages included in the `build` sections. Notably using the `{{ compiler(...) }}` jinja templates causes two separate environments to be created.

It looks like if you have a recipe that uses both the `build` and `host` sections as expected, then we end up improperly setting the search paths to find packages in the `build` environment as opposed to the `host` environment: https://github.com/rapidsai/rapids-cmake/blob/branch-22.06/rapids-cmake/cmake/support_conda_env.cmake#L78-L83. This causes issues when building a project that depends on something that CMake or other build tools also depend on, i.e. `libcurl`.

In general, we should probably make searching in the `build` environment opt-in as opposed to default behavior because it can cause issues in cross compiling, but I understand some rapids conda recipes are defined using only `build` and `run` as of now. An easy temporary fix for these recipes would be to just change the `build` section to `host` for now.

**Expected behavior**

When using `rapids_cmake_support_conda_env` I expect the packages defined in the `host` section of my conda recipe to be found before finding packages in the `build` environment.

cc @jakirkham because conda recipes 😄

kkraus14:
Fixes #179

Swaps the ordering to prioritize packages in `$PREFIX` before `$BUILD_PREFIX`. Also updates tests to ensure the same.

### DIFF
--- a/rapids-cmake/cmake/support_conda_env.cmake
+++ b/rapids-cmake/cmake/support_conda_env.cmake
@@ -31,11 +31,11 @@ Creates a global interface target called `target_name` that holds
 the CONDA include and link directories, when executed.
 
 Also offers the ability to modify :cmake:variable:`CMAKE_PREFIX_PATH <cmake:variable:CMAKE_PREFIX_PATH>` to
-include the paths in environment variables `BUILD_PREFIX`, `PREFIX`,
+include the paths in environment variables `PREFIX`, `BUILD_PREFIX`,
 and `CONDA_PREFIX` based on the current CONDA environment.
 
 ``MODIFY_PREFIX_PATH``
-    When in a conda build environment the contents of `$ENV{BUILD_PREFIX}` and `$ENV{PREFIX}`
+    When in a conda build environment the contents of `$ENV{PREFIX}` and `$ENV{BUILD_PREFIX}`
     will be inserted to the front of :cmake:variable:`CMAKE_PREFIX_PATH <cmake:variable:CMAKE_PREFIX_PATH>`.
 
     When in a conda environment the contents of `$ENV{CONDA_PREFIX}` will be inserted to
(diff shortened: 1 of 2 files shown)


robertmaynard:
So is this not an issue?

https://github.com/rapidsai/rapids-cmake/issues/179#issuecomment-1099561563

kkraus14:
@jakirkham was referring to the `PATH` environment variable which makes sense why it would want to find things in the build environment first for using executables.

I would say that maybe we could use `CMAKE_PROGRAM_PATH` so that we can prioritize the build env for `find_program` calls but looks like it still searches in `CMAKE_PREFIX_PATH` first and there isn't any other variables that can be set to influence `find_package` calls.

jakirkham:
In the upstream issue, I tried to articulate why reversing the order might make sense ( https://github.com/conda/conda-build/issues/4432#issuecomment-1099567199 ). Though there are likely other reasons the current ordering is desirable (like cross-compiling).

robertmaynard:
I will need to verify that this doesn't break projects like cudf before we merge

kkraus14:
> I will need to verify that this doesn't break projects like cudf before we merge

Thanks @robertmaynard!

kkraus14:
@robertmaynard gentle ping on this. Any chance we could get this in for 22.10? I'm happy to rebase and retarget.

robertmaynard:
> @robertmaynard gentle ping on this. Any chance we could get this in for 22.10? I'm happy to rebase and retarget.

Yes we can get this in for 22.10, lets rebase and we should be good to merge early next week.

kkraus14:
@robertmaynard rebased and retargeted to 22.10, thanks!

robertmaynard:
@gpucibot merge

PR #184, opened 2022-04-27
Add rapids-cython component for scikit-build based Python package builds

vyasr:
This PR introduces various utilities that will be used across all RAPIDS Python packages that are built using scikit-build. Centralizing these utilities reduces the CMake configurations of our Python libraries to a minimum. For now, all of the content is purely CMake code, so this repo is a good place for them. If we ever add Python utilities (e.g. to simplify setup.py scripts) we may need to move to a new repo rather than polluting rapids-cmake with Python code.


robertmaynard:
Please add `.. versionadded:: v22.06.00` to the documentation in each cython/*.cmake file

rapids-cmake/cython/create_cython_modules.cmake
```diff
@@ -0,0 +1,64 @@
...
+``install_base_directory``
+  The source directory of the project. This directory is used to compute the
+  relative install path, which is necessary to propertly support differently
+  configured installations such as installing in place vs. out of place.
+
+#]=======================================================================]
+function(create_cython_modules cython_modules linked_libraries install_base_directory)
+  foreach(cython_module ${cython_modules})
+    add_cython_target(${cython_module} CXX PY3)
+    add_library(${cython_module} MODULE ${cython_module})
```
robertmaynard, line 49: So am I understanding this correctly, `cython_modules` holds a list of source files and each one is mapped to a new target? 

If that is correct, the style of this should be `rapids_cython_create_modules(MODULES <list of srcs> ... )` instead of having the user specify a variable that holds the sources.

rapids-cmake/cython/create_cython_modules.cmake
```diff
@@ -0,0 +1,64 @@
...
+    python_extension_module(${cython_module})
+
+    # To avoid libraries being prefixed with "lib".
+    set_target_properties(${cython_module} PROPERTIES PREFIX "")
+    foreach(lib ${linked_libraries})
+      target_link_libraries(${cython_module} ${lib})
```
robertmaynard, line 55: target_link_libraries should always explicitly use the `PUBLIC` keyword.

rapids-cmake/cython/skbuild_patches.cmake
```diff
@@ -0,0 +1,53 @@
...
+the visibility of symbols because otherwise any library that exports symbols
+with external linkage will have the visibility of those symbols changed
+undesirably. We can remove this function once this issue is resolved in
+scikit-build.
+
+Issue: https://github.com/scikit-build/scikit-build/issues/668
+PR: https://github.com/scikit-build/scikit-build/pull/703
+
+#]=======================================================================]
+# TODO: Should we guard this based on a scikit-build version? Override this function to avoid
```
robertmaynard, line 30: As we refine this PR, I would like to see this moved to something like cython/detail/... and have a cython/init.cmake function that calls all the needed hooks.

rapids-cmake/rapids-cython.cmake
```diff
@@ -0,0 +1,33 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#=============================================================================
+include_guard(GLOBAL)
+
+find_package(PythonExtensions REQUIRED)
```
robertmaynard, line 18: Move these find_packages to cython/init.cmake and have `init` be a macro.

The `rapids-<component>.cmake` files should all be side-effect free

robertmaynard: Actually move everything before the `includes` to the init.cmake macro.

robertmaynard:
The function names need to change to match the rapids-cmake style so `create_cython_modules` -> `rapids_cython_create_modules`

rapids-cmake/cython/create_cython_modules.cmake
vyasr: I've updated the way that I handle arguments here, let me know if I've done this correctly. I'm still always a little thrown by the way that parameters are handled since anything that isn't named in the function signature just gets passed via argc/argv.

vyasr:
In addition to the inline TODOs, I have one additional question. Currently the `rapids_cython_create_modules` function is set up so that it will be called with modules in the same directory. The implication of that is that every Python subpackage will contain its own CMakeLists.txt. I set it up this way because that's how I've seen CMakeLists.txt files organized in projects in the past. Most RAPIDS projects seem to prefer a single monolithic CMakeLists.txt for all of the C++ code. Should I rework this function to match that pattern instead? That would entail some additional processing of file paths, but would leave us with a single listfile for all of the Python code.

rapids-cmake/cython/init.cmake
```diff
@@ -0,0 +1,86 @@
...
+Result Variables
+^^^^^^^^^^^^^^^^
+  :cmake:variable:`RAPIDS_CYTHON_INITIALIZED` will be set to TRUE.
+  :cmake:variable:`CYTHON_FLAGS` will be set to a standard set of a flags to pass to the command line cython invocation.
+
+#]=======================================================================]
+macro(rapids_cython_init)
+  # Verify that we are using scikit-build.
+  if(NOT DEFINED SKBUILD)
+    message(FATAL_ERROR "rapids-cython must be used with scikit-build")
```
robertmaynard, line 47:
```suggestion
    message(FATAL_ERROR "rapids-cython requires scikit-build to be active before being used")
```

rapids-cmake/cython/init.cmake
```diff
@@ -0,0 +1,86 @@
...
+  endif()
+
+  # Incorporate scikit-build patches.
+  include("${RAPIDS_CYTHON_PATH}/detail/skbuild_patches.cmake")
+
+  # TODO: Do we want to use rapids_find_package here? It's a little odd because they aren't
```
robertmaynard, line 53: I think `find_package` here is fine since the produced binaries don't need either of these packages to be used ( correct? ).

rapids-cmake/cython/init.cmake
```diff
@@ -0,0 +1,86 @@
...
+  # independent packages, they are part of scikit-build, so I don't know if tracking them as
+  # dependencies for an export set really makes sense.
+  find_package(PythonExtensions REQUIRED)
+  find_package(Cython REQUIRED)
+
+  # Set standard Cython directives that all RAPIDS projects should use in compilation.
+  set(CYTHON_FLAGS "--directive binding=True,embedsignature=True,always_allow_keywords=True"
+      CACHE STRING "The directives for Cython compilation.")
+  # TODO: The above flags must be configurable. We may not want/need it to be a cache variable, so
```
robertmaynard, line 62: I would look at `rapids_cmake_install_dir` We can add a `MODIFY_CYTHON_FLAGS` option that will append this flags to the CYTHON_FLAGS variable.

So the end result would look roughly like:
```cmake
if( RAPIDS_CYTHONMODIFY_CYTHON_FLAGS)
  string(APPEND CYTHON_FLAGS "--directive binding=True,embedsignature=True,always_allow_keywords=True"))
endif()
```

rapids-cmake/cython/init.cmake
```diff
@@ -0,0 +1,86 @@
...
+
+TODO: Should the cache variable be documented differently from the normal one?
+
+Result Variables
+^^^^^^^^^^^^^^^^
+  :cmake:variable:`RAPIDS_CYTHON_INITIALIZED` will be set to TRUE.
+  :cmake:variable:`CYTHON_FLAGS` will be set to a standard set of a flags to pass to the command line cython invocation.
+
+#]=======================================================================]
+macro(rapids_cython_init)
```
robertmaynard, line 44:
```suggestion
macro(rapids_cython_init)
  list(APPEND CMAKE_MESSAGE_CONTEXT "rapids.cython.init")
```

rapids-cmake/cython/init.cmake
```diff
@@ -0,0 +1,86 @@
...
+.. versionadded:: v22.06.00
+
+Simple helper function for rapids-cython components to verify that rapids_cython_init has been called before they proceed.
+
+.. code-block:: cmake
+
+  rapids_cython_verify_init()
+
+#]=======================================================================]
+function(rapids_cython_verify_init)
```
robertmaynard, line 82: Move this function into `cython/detail/` namespace. Lets keep it to a single function per file.

rapids-cmake/cython/create_modules.cmake
```diff
@@ -0,0 +1,69 @@
...
+.. versionadded:: v22.06.00
+
+Generate C(++) from Cython and create Python modules.
+
+.. code-block:: cmake
+
+  rapids_cython_create_modules(<ModuleName...>)
+
+Creates a Cython target for a module, then adds a corresponding Python
+extension module. This function must be called after rapids_cython_init.
```
robertmaynard, line 28:
```suggestion
extension module. 

.. note::
  Requires :cmake:command:`rapids_cython_init` to be called before usage
```

rapids-cmake/cython/create_modules.cmake
```diff
@@ -0,0 +1,69 @@
...
+
+``MODULES``
+  The list of modules to build.
```
robertmaynard, line 31: We should clarify what modules means. Maybe `This list of sources to build. Each source will be converted to a unique cython module"

robertmaynard (changes requested):
We need to add tests to validate that everything works as expected. See the `testing/cuda` directory for examples.

I would like to see at least the following tests:
- A check that validates the `rapids-cythoon.cmake` file can be included
- A check that cython_init can be called multiple times and won't add duplicate flags 
- A check that `skbuild_patches` work when `init` is called
  - We can check via `if(COMMAND ` ) that `_set_python_extension_symbol_visibility` exists
- Checks for rapids_cython_create_modules being called with insufficient parameters
  - if LINKED_LIBRARIES isn't provided we work
  - if MODULES are empty we are a no-op
  - if INSTALL_ROOT isn't provided we don't generate install rules.
- A check that rapids_cython_create_modules can handle two duplicate calls by not creating targets that already exist
- We get the correct errors when trying to use `rapids_cython_create_modules` before calling `rapids_cython_init`

rapids-cmake/cython/create_modules.cmake
shwina: Regarding terminology: we might want to consistently use "extension modules"  to refer to the `.so` files resulting from the compilation of  `.pyx` files ("Cython source files").

cmake-format-rapids-cmake.json
```diff
@@ -236,6 +236,16 @@
           "GLOBAL_TARGETS": "+",
           "FIND_ARGS": "+"
         }
+      },
```
robertmaynard, line 239: We need to also add `rapids_cython_init` to the format file.

rapids-cmake/cython/create_modules.cmake
```diff
@@ -0,0 +1,69 @@
...
+  rapids_cython_verify_init()
+
+  set(_rapids_cython_options)
+  set(_rapids_cython_one_value INSTALL_BASE_DIRECTORY)
+  set(_rapids_cython_multi_value MODULES LINKED_LIBRARIES)
+  cmake_parse_arguments(RAPIDS_CYTHON "${_rapids_cython_options}" "${_rapids_cython_one_value}"
+                        "${_rapids_cython_multi_value}" ${ARGN})
+
+  foreach(cython_module ${RAPIDS_CYTHON_MODULES})
+    add_cython_target(${cython_module} CXX PY3)
```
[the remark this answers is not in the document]
vyasr: I made this an option instead. Most Cython-related tooling still assumes C as a default language, so I think it makes more sense to support is as a possibility (and perhaps one day we'll be able to upstream this function to scikit-build).

rapids-cmake/cython/init.cmake
```diff
@@ -0,0 +1,86 @@
...
+
+Result Variables
+^^^^^^^^^^^^^^^^
+  :cmake:variable:`RAPIDS_CYTHON_INITIALIZED` will be set to TRUE.
+  :cmake:variable:`CYTHON_FLAGS` will be set to a standard set of a flags to pass to the command line cython invocation.
+
+#]=======================================================================]
+macro(rapids_cython_init)
+  # Verify that we are using scikit-build.
+  if(NOT DEFINED SKBUILD)
```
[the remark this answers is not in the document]
vyasr: In theory someone could build their own CMake tooling for building Cython modules the same way that scikit-build works. I don't foresee any alternative usage right now, I just added this as a sanity check.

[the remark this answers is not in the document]
vyasr: I can remove it if you think it's superfluous.

rapids-cmake/cython/init.cmake
vyasr: You are correct these are only necessary during the build process. The one caveat is that the `python_extension_module` command I think implicitly links to the Python runtime library since Cython is built around the Python C API. I think we're probably better off leaving that dependency implicit rather than trying to build that into `rapids-cython`'s dependency tracking. Someone would be doing _a lot_ of things wrong if they were trying to build this package without a valid Python runtime environment.

rapids-cmake/cython/init.cmake
vyasr: Rather than appending, I instead chose to override completely. The way that the `cython` command line executable accepts arguments can be confusing (for example the `--directive` parameter itself accepts a comma-separated list, and it's not the only one) so I think overwriting rather than appending will behave more predictably for users.

rapids-cmake/cython/init.cmake
vyasr: I removed this function altogether. Since we want other functions to not assume rapids-cython has been initialized, they will always just call it first, so I don't think there's a need for this checking function.

rapids-cmake/cython/create_modules.cmake
vyasr: Yes, I think this is the right approach. We should consistently use "Python extension module" for Python libraries we are building to differentiate clearly from CMake modules (or pure Python files for that matter).

rapids-cmake/cython/init.cmake
vyasr: Also, since scikit-build's way of specifying the flags is for the calling code to just set the `CYTHON_FLAGS` variable (rather than passing as an argument) I stuck with the same approach where I just check `if(DEFINED CYTHON_FLAGS)`. Let me know if you think we should change that.

rapids-cmake/cython/create_modules.cmake
```diff
@@ -0,0 +1,69 @@
...
+function(rapids_cython_create_modules)
+  rapids_cython_verify_init()
+
+  set(_rapids_cython_options)
+  set(_rapids_cython_one_value INSTALL_BASE_DIRECTORY)
+  set(_rapids_cython_multi_value MODULES LINKED_LIBRARIES)
+  cmake_parse_arguments(RAPIDS_CYTHON "${_rapids_cython_options}" "${_rapids_cython_one_value}"
+                        "${_rapids_cython_multi_value}" ${ARGN})
+
+  foreach(cython_module ${RAPIDS_CYTHON_MODULES})
```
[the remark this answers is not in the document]
vyasr: This worked if I used `IN LISTS` instead of `IN LIST`, which seems consistent with the [`foreach` documentation](https://cmake.org/cmake/help/latest/command/foreach.html), but let me know if I'm missing something.

rapids-cmake/cython/detail/skbuild_patches.cmake
```diff
@@ -0,0 +1,53 @@
...
+
+  if(PYTHON_VERSION_MAJOR VERSION_GREATER 2)
+    set(_modinit_prefix "PyInit_")
+  else()
+    set(_modinit_prefix "init")
+  endif()
+  message("_modinit_prefix:${_modinit_prefix}")
+  if("${CMAKE_C_COMPILER_ID}" STREQUAL "MSVC")
+    set_target_properties(${_target} PROPERTIES LINK_FLAGS "/EXPORT:${_modinit_prefix}${_target}")
+  elseif("${CMAKE_C_COMPILER_ID}" STREQUAL "GNU" AND NOT ${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
```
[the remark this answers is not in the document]
vyasr: Do we want to put much effort into patching this function, or should we be upstreaming these questions to https://github.com/scikit-build/scikit-build/pull/703?

rapids-cmake/cython/init.cmake
vyasr: I'll make this change if we keep the block (the previous thread).

rapids-cmake/cython/create_modules.cmake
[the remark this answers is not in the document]
robertmaynard: Yea just a typo on my side. Should be `IN LISTS`

vyasr:
@robertmaynard ready for another pass. I ran into only one significant hitch with your requests. Since these tests are not being run inside a scikit-build managed build, init will fail. Even if I remove the check for the `SKBUILD` variable test that you commented on, it will still fail afterwards when it tries to use the CMake modules provided by scikit-build. Should we `FetchContent` those scripts from the scikit-build Github repo and then include them? I don't know how else we would be able to test them.

rapids-cmake/cython/create_modules.cmake
```diff
@@ -0,0 +1,101 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software distributed under the License
+# is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+# or implied. See the License for the specific language governing permissions and limitations under
+# the License.
+# =============================================================================
+
+include_guard(GLOBAL)
+
+include(${CMAKE_CURRENT_LIST_DIR}/detail/verify_init.cmake)
```
robertmaynard, line 17: To match rapids-cmake style move the include into the function

rapids-cmake/cython/create_modules.cmake
```diff
@@ -0,0 +1,101 @@
...
+  correspondence with extension modules to build, i.e. for every `<Name>.pyx`
+  in SOURCE_FILES we assume that `<Name>.pyx` is a Cython source file for which
+  an extension module `<Name>` should be built.
+
+``LINKED_LIBRARIES``
+  The list of libraries that need to be linked into all modules. In RAPIDS,
+  this list usually contains (at minimum) the corresponding C++ libraries.
+
+``INSTALL_ROOT``
+  The source directory of the project. This directory is used to compute the
```
robertmaynard, line 52: Am I correct in understanding this needs to be an absolute path? If so we should document that requirement here.

rapids-cmake/cython/create_modules.cmake
```diff
@@ -0,0 +1,101 @@
...
+rapids_cython_create_modules
+---------------------
+
+.. versionadded:: v22.06.00
+
+Generate C(++) from Cython and create Python modules.
+
+.. code-block:: cmake
+
+  rapids_cython_create_modules(<ModuleName...>)
```
robertmaynard, line 29:
```suggestion
  rapids_cython_create_modules(<ModuleName...> [CXX] [SOURCE_FILES <src1> <src2> ...] [LINKED_LIBRARIES <lib1> <lib2> ... ]  [INSTALL_ROOT <install_path> )
```

rapids-cmake/cython/detail/skbuild_patches.cmake
```diff
@@ -0,0 +1,56 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software distributed under the License
+# is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+# or implied. See the License for the specific language governing permissions and limitations under
+# the License.
+# =============================================================================
+
+include_guard(GLOBAL)
+
+include(${CMAKE_CURRENT_LIST_DIR}/verify_init.cmake)
```
robertmaynard, line 17: Same here, move the include into the `_set_python_extension_symbol_visibility` function.

rapids-cmake/cython/detail/skbuild_patches.cmake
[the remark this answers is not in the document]
robertmaynard: We should push these upstream and just match the same logic as upstream here.

rapids-cmake/cython/init.cmake
```diff
@@ -0,0 +1,86 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software distributed under the License
+# is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+# or implied. See the License for the specific language governing permissions and limitations under
+# the License.
+# =============================================================================
+
+# TODO: What is the correct way to get this path inside the macro below? If there isn't one, is
+# there a different solution that would be more appropriate?
+set(RAPIDS_CYTHON_PATH)
```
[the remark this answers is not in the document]
robertmaynard: dang macros. We can just hardcode it to `${rapids-cmake-dir}/cython/` inside the macro

robertmaynard (changes requested):
> I ran into only one significant hitch with your requests. Since these tests are not being run inside a scikit-build managed build, init will fail. Even if I remove the check for the SKBUILD variable test that you commented on, it will still fail afterwards when it tries to use the CMake modules provided by scikit-build. Should we FetchContent those scripts from the scikit-build Github repo and then include them? I don't know how else we would be able to test them.

I expect you can extend `https://github.com/rapidsai/rapids-cmake/blob/branch-22.06/testing/utils/fill_cache/CMakeLists.txt` to provide scikit-build to the tests. No need to add a `rapids_cpm_scikitbuild` you can call `rapids_cpm_find` with the relevant details.

rapids-cmake/cython/init.cmake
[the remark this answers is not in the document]
vyasr: I've changed this to a warning so that we can run tests. I think that's probably the appropriate way to handle this.

rapids-cmake/cython/create_modules.cmake
vyasr: Does it need to be? Does `cmake_path(RELATIVE_PATH <path-var> BASE_DIRECTORY base)` require that `base` be an absolute path, or can it also be relative (to the current directory)?

vyasr:
OK, here's what I think is still outstanding:
- We need to decide how to treat the `CYTHON_FLAGS`. Currently my tests are failing there, in part because I'm probably doing something wrong, but also because of the behavior of cache vs non-cache variables. [scikit-build defines it as a cache variable](https://scikit-build.readthedocs.io/en/latest/cmake-modules/Cython.html), but I'm not clear on whether that means that we want to mimic that or whether that means we need to _avoid_ that to avoid conflicts.
- The logic that I have in place for loading the scikit-build artifacts into the cache works for me locally, but it seems to be failing in CI. My approach involving globs may be fragile, and I'd appreciate pointers.
- I haven't written tests of `rapids_create_cython_modules` yet. My main question here is, should I include actual Cython files in order for this to work? I guess it might work if we just create temporary pyx files since we won't actually compile, we just need the configuration to work?
- I ran into some very strange issues while testing this with raft where the linking worked on one filesystem but not another. It looked like the correct linker flag was set during compilation in both cases. It's possible that what I observed was due to some cached environment variable somewhere helping raft find the library when it wasn't correctly included in the search path, but I'm not convinced that that's the only issue because when I specified `LD_LIBRARY_PATH` Python was able to load the raft library but actually running tests caused it to freeze, which makes me think something is wrong with the build. That could be specific to the raft build, but it could also be due to something here, so it would probably be worth checking on.
- Just FYI, I had to revert [this suggestion](https://github.com/rapidsai/rapids-cmake/pull/184#discussion_r861798039) of yours because I ran into [the issue discussed on this SO thread](https://stackoverflow.com/questions/47737558/uses-of-target-link-libraries-must-be-either-all-keyword-or-all-plain) because of scikit-build's existing usage of `target_link_libraries`.

rapids-cmake/cython/create_modules.cmake
robertmaynard: No that supports relative paths.

I think the root issue is that I am confused on why this variable is needed. It is labelled INSTALL_ROOT, but we are asking for the source of the project. Do we mean the value that is held in [PROJECT_SOURCE_DIR](https://cmake.org/cmake/help/latest/variable/PROJECT_SOURCE_DIR.html)? 

Can you point me to an existing usage as that would be helpful

rapids-cmake/cython/create_modules.cmake
```diff
@@ -0,0 +1,69 @@
...
+  foreach(cython_module ${RAPIDS_CYTHON_MODULES})
+    add_cython_target(${cython_module} CXX PY3)
+    add_library(${cython_module} MODULE ${cython_module})
+    python_extension_module(${cython_module})
+
+    # To avoid libraries being prefixed with "lib".
+    set_target_properties(${cython_module} PROPERTIES PREFIX "")
+    foreach(lib ${RAPIDS_CYTHON_LINKED_LIBRARIES})
+      target_link_libraries(${cython_module} PUBLIC ${lib})
+    endforeach()
```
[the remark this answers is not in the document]
robertmaynard: Since this isn't possible currently due to scikitbuild current setup, we should raise an issue with scikitbuild to switch over to using explicit `PUBLIC`

robertmaynard:
> We need to decide how to treat the CYTHON_FLAGS. Currently my tests are failing there, in part because I'm probably doing something wrong, but also because of the behavior of cache vs non-cache variables. [scikit-build defines it as a cache variable](https://scikit-build.readthedocs.io/en/latest/cmake-modules/Cython.html), but I'm not clear on whether that means that we want to mimic that or whether that means we need to avoid that to avoid conflicts.

So in general I would recommend first going with a non cache variable. When a cache and non-cache variable exist with the same name the non-cache one is used by CMake. So this allows us to easily specify a value.

Looking at this example ( https://github.com/scikit-build/scikit-build/blob/6b112fb69521d724eff27f21c6a2e4aa06743cbb/tests/samples/cython-flags/hello/CMakeLists.txt#L2 ) it seem that using a non-cache variable for CYTHON_FLAGS is valid.

robertmaynard:
We will want the following changes so that we can test `cython/init.cmake`

```
diff --git a/rapids-cmake/cython/init.cmake b/rapids-cmake/cython/init.cmake
index 765830c..ed7fab3 100644
--- a/rapids-cmake/cython/init.cmake
+++ b/rapids-cmake/cython/init.cmake
@@ -46,15 +46,14 @@ macro(rapids_cython_init)
     # Incorporate scikit-build patches.
     include("${rapids-cmake-dir}/cython/detail/skbuild_patches.cmake")
 
-    find_package(PythonExtensions REQUIRED)
-    find_package(Cython REQUIRED)
-
     # Set standard Cython directives that all RAPIDS projects should use in compilation. Users flags
     # will not be overridden.
-    if(NOT CYTHON_FLAGS)
+    if(NOT DEFINED CYTHON_FLAGS)
       set(CYTHON_FLAGS "--directive binding=True,embedsignature=True,always_allow_keywords=True"
           CACHE STRING "The directives for Cython compilation.")
     endif()
+    find_package(PythonExtensions REQUIRED)
+    find_package(Cython REQUIRED)
 
     # Flag
     set(RAPIDS_CYTHON_INITIALIZED TRUE)
diff --git a/testing/CMakeLists.txt b/testing/CMakeLists.txt
index 689a094..6017398 100644
--- a/testing/CMakeLists.txt
+++ b/testing/CMakeLists.txt
@@ -23,7 +23,6 @@ include(utils/cmake_build_test.cmake)
 
 add_subdirectory(cmake)
 add_subdirectory(cuda)
-add_subdirectory(cython)
 add_subdirectory(export)
 add_subdirectory(find)
 
@@ -36,6 +35,7 @@ if(RAPIDS_CMAKE_ENABLE_DOWNLOAD_TESTS)
   setup_cpm_cache()
 
   add_subdirectory(cpm)
+  add_subdirectory(cython)
   add_subdirectory(other)
 endif()
 
diff --git a/testing/cython/init.cmake b/testing/cython/init.cmake
index 75df03b..68edecf 100644
--- a/testing/cython/init.cmake
+++ b/testing/cython/init.cmake
@@ -15,9 +15,12 @@
 include(${rapids-cmake-dir}/cython/init.cmake)
 
+set(SKBUILD ON)
+
 # Ensure that scikit-build's CMake files are discoverable.
-file(GLOB skbuild_resource_dir "${CPM_SOURCE_CACHE}/skbuild/*/skbuild/resources/cmake")
-LIST(APPEND CMAKE_MODULE_PATH "${skbuild_resource_dir}")
+file(GLOB skbuild_resource_dir LIST_DIRECTORIES ON "${CPM_SOURCE_CACHE}/skbuild/*/skbuild/resources/cmake")
+
+list(APPEND CMAKE_MODULE_PATH "${skbuild_resource_dir}")
 
 # Test that rapids_cython_init initializes the expected variables.
 rapids_cython_init()
@@ -25,8 +28,8 @@ if(NOT DEFINED RAPIDS_CYTHON_INITIALIZED)
   message(FATAL_ERROR "rapids_cython_init didn't correctly set RAPIDS_CYTHON_INITIALIZED")
 endif()
 
-STRING(REGEX MATCHALL ".*--directives.*" matches "${CYTHON_FLAGS}")
-LIST(LENGTH matches num_directives)
+string(REGEX MATCHALL ".*--directives.*" matches "${CYTHON_FLAGS}")
+list(LENGTH matches num_directives)
 
 if(NOT CYTHON_FLAGS OR NOT num_directives EQUAL 1)
   message(FATAL_ERROR "rapids_cython_init didn't correctly set CYTHON_FLAGS")
@@ -38,8 +41,8 @@ endif()
 
 # Test that rapids_cython_init is idempotent.
 rapids_cython_init()
-STRING(REGEX MATCHALL ".*--directives.*" matches "${CYTHON_FLAGS}")
-LIST(LENGTH matches num_directives)
+string(REGEX MATCHALL ".*--directives.*" matches "${CYTHON_FLAGS}")
+list(LENGTH matches num_directives)
 
 if(NOT num_directives EQUAL 1)
   message(FATAL_ERROR "rapids_cython_init is not idempotent, num_directives = ${num_directives}")
```

The big changes are setting `CYTHON_FLAGS` before we call the module. This is required when using `CACHE` variables as if a cache variable already exists any call to `set(CACHE)` after the first will be ignored. This is another reason why moving to a normal variable ( with `PARENT_SCOPE` ) should be used if we can.

Second big change is moving the `add_subdirectory(cython)` into the `RAPIDS_CMAKE_ENABLE_DOWNLOAD_TESTS` check so that has access to the CPM source cache

vyasr:
I actually have the change to move `add_subdirectory(cython)` inside the `if(RAPIDS_CMAKE_ENABLE_DOWNLOAD_TESTS)` locally, but I had told git to temporarily ignore changes to that file since I have also commented out builds of all other tests locally, so the changes weren't committed... sorry for the confusion there. Names like `init` are too generic for `ctest -R`

rapids-cmake/cython/create_modules.cmake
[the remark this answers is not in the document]
vyasr: Sounds good. What is the justification? The keyword syntax is the more modern approach? We may need to make a case since this would break prior builds relying on the current behavior.

rapids-cmake/cython/create_modules.cmake
vyasr: I realized that you only get half the picture here. The relevant TODO item is on the corresponding raft PR [here](https://github.com/rapidsai/raft/pull/633/files#diff-e249c4fccd88ad8117d97abbcd719c9c2699ff75679c2e00d1def250132904bcR20). The variable is then used to determine the relative path to the install directory in this PR [here](https://github.com/rapidsai/rapids-cmake/pull/184/files#diff-528f59a9f5467b251f1dbafa7184b1bb1f93776a483b632da0bba7d7790d0b6dR96).

rapids-cmake/cython/detail/skbuild_patches.cmake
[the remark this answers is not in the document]
vyasr: Got it. I'll make a PR upstream and tag you. Leaving this thread open as a reminder that this is a task for me.

cmake-format-rapids-cmake.json
```diff
@@ -236,6 +236,21 @@
...
+        "pargs": {
+          "nargs": "0"
+        }
+      },
+      "rapids_cython_create_modules": {
+        "pargs": {
+          "nargs": "0"
+        },
+        "kwargs": {
+          "MODULES": "+",
```
robertmaynard, line 250: The argument names for `rapids_cython_create_modules` need to be updated

rapids-cmake/cython/create_modules.cmake
```diff
@@ -0,0 +1,96 @@
...
+rapids_cython_create_modules
+----------------------------
+
+.. versionadded:: v22.06.00
+
+Generate C(++) from Cython and create Python modules.
+
+.. code-block:: cmake
+
+  rapids_cython_create_modules([CXX] [SOURCE_FILES <src1> <src2> ...] [LINKED_LIBRARIES <lib1> <lib2> ... ]  [INSTALL_ROOT <install_path> )
```
robertmaynard, line 27:
```suggestion
  rapids_cython_create_modules([CXX] [SOURCE_FILES <src1> <src2> ...] [LINKED_LIBRARIES <lib1> <lib2> ... ]  [INSTALL_DIR <rel_path> )
```

robertmaynard (changes requested):
A couple of doc changes and we are good to go

vyasr:
@robertmaynard I've addressed the two final requests that you had. There are also a couple of TODO items that I had questions about in the tests, `create_modules/CMakeLists.txt` and `create_modules_with_library/CMakeLists.txt`. Given that you're leaving for vacation tomorrow, though, I'm also OK with merging this PR (assuming that you're happy) and addressing these questions in a follow-up. Getting this PR merged would make it easier for me to start experimenting with this on other RAPIDS libraries while you're out.

testing/cython/create_modules/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#=============================================================================
+
+# TODO: For some reason I have to specify the minimum required cmake version
+# before including init.cmake below even though none of them execute code.
```
robertmaynard, line 18: Ahh yes.

Very crudely, CMake at parse time parses the entire macro and stores a representation of it instead of executing it. That state of all CMake policies are captured at that initial parse and used when the macro is called ( crazy right ). So that is why you need to specify a minimum required version before the macro is parsed

testing/cython/create_modules/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+# state saved when the macro in init.cmake is defined because UseCython.cmake
+# is only actually loaded when rapids_cython_init is called.
+
+include(${rapids-cmake-dir}/cython/create_modules.cmake)
+cmake_minimum_required(VERSION 3.20)
+include(${rapids-cmake-dir}/cython/init.cmake)
+
+project(rapids_cython-create_modules LANGUAGES C CXX)
+
+# TODO: Is there any way to handle setting SKBUILD and the CMAKE_MODULE_PATH
```
robertmaynard, line 31: Not currently. Every tests is run as a separate CMake process and only a fixed set of arguments are provided.

For now just set these in each test

testing/cython/create_modules_with_library/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
...
+set(SKBUILD ON)
+
+# Ensure that scikit-build's CMake files are discoverable. The glob is to
+# capture the current git commit hash.
+file(GLOB skbuild_resource_dir LIST_DIRECTORIES ON "${CPM_SOURCE_CACHE}/skbuild/*/skbuild/resources/cmake")
+LIST(APPEND CMAKE_MODULE_PATH "${skbuild_resource_dir}")
+
+rapids_cython_init()
+
+# TODO: This whole file only exists because we cannot create multiple global
```
robertmaynard, line 44: Nope! CMake always uses global scope for targets constructed via `add_library` or `add_executable`. So separate tests are always needed

rapids-cmake/cython/detail/skbuild_patches.cmake
```diff
@@ -0,0 +1,55 @@
...
+with external linkage will have the visibility of those symbols changed
+undesirably. We can remove this function once this issue is resolved in
+scikit-build.
+
+Issue: https://github.com/scikit-build/scikit-build/issues/668
+PR: https://github.com/scikit-build/scikit-build/pull/703
+
+#]=======================================================================]
+function(_set_python_extension_symbol_visibility _target)
+  # TODO: Should we guard this function's existence based on a scikit-build version?
```
robertmaynard, line 35: I think we should create an issue on rapids-cmake that tracks that this needs to be guarded once scikit-build has been fixed. Otherwise for now we need to do this unconditionally since we don't know what version will have the fix.

robertmaynard (approved):

vyasr:
@gpucibot merge

robertmaynard:
@gpucibot merge

PR #187, opened 2022-05-10
Use target name variable.

bdice:
When building libcudf, I see a message `Generate git revision file for {target}` before the build begins. I think this is intended to say `${target}` and insert the variable name.

Also, I notice this message because this stage takes a long time (something like 10 seconds) for most of my builds, even partial builds. What does this command actually do, and why does it take so long? Can it be cached / accelerated / disabled?

### DIFF
--- a/rapids-cmake/cmake/write_git_revision_file.cmake
+++ b/rapids-cmake/cmake/write_git_revision_file.cmake
@@ -98,7 +98,7 @@ function(rapids_cmake_write_git_revision_file target file_path)
 
   add_custom_target(${target}_compute_git_info ALL
                     BYPRODUCTS "${file_path}"
-                    COMMENT "Generate git revision file for {target}"
+                    COMMENT "Generate git revision file for ${target}"
                     COMMAND ${CMAKE_COMMAND} -DWORKING_DIRECTORY=${CMAKE_CURRENT_SOURCE_DIR}
                             -DGIT_EXECUTABLE=${GIT_EXECUTABLE} -DRAPIDS_GIT_PREFIX=${RAPIDS_PREFIX}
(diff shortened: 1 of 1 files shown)


robertmaynard:
>  I think this is intended to say ${target} and insert the variable name.

Correct.

> Also, I notice this message because this stage takes a long time (something like 10 seconds) for most of my builds, even partial builds. What does this command actually do, and why does it take so long? Can it be cached / accelerated / disabled?

It calls into git to compute a set of information ( branch name, SHA1 hash, have local changes, `git describe tag` ) for the source code directory. This evaluation can't be cached since the build-system ( CMake ) has no knowledge of what files have changed. We can't do this computation at CMake configuration time, since it won't re-run whenever source files are changed.

I don't see this function being called in cudf or rmm so it would be great to know what dependency is using it. Maybe it can be removed.

robertmaynard (approved):

robertmaynard:
@bdice Are you good with this being merged?

bdice:
@robertmaynard Thanks for the helpful info! I'll merge it.

> I don't see this function being called in cudf or rmm so it would be great to know what dependency is using it. Maybe it can be removed.

Hmm. I'll see if I can figure out what's calling this next time it comes up.

bdice:
@gpucibot merge

bdice:
@robertmaynard I don't have merge privileges, oops. However, I also noticed that `CODEOWNERS` reports errors: https://github.com/rapidsai/rapids-cmake/blob/branch-22.06/.github/CODEOWNERS

Looks like `@rapidsai/rapids-cmake-codeowners` doesn't exist.

robertmaynard:
@gpucibot merge

PR #189, opened 2022-05-11
Default Cython module RUNPATH to $ORIGIN and return the list of created targets

vyasr:
This PR sets the default RPATH for Cython-generated Python extension modules to $ORIGIN so that they will default to finding libraries within the same directory. If rapids-cmake users want to enforce a different directory layout in their library, they are responsible for overriding the RPATHs themselves. To facilitate this as well as other potential downstream modifications, `rapids_cython_create_modules` now also returns the set of created targets in the `RAPIDS_CYTHON_CREATED_TARGETS` variable.

### DIFF
--- a/RAPIDS.cmake
+++ b/RAPIDS.cmake
@@ -21,8 +21,8 @@ set(rapids-cmake-version 22.06)
 include(FetchContent)
 FetchContent_Declare(
   rapids-cmake
-  GIT_REPOSITORY https://github.com/vyasr/rapids-cmake.git
-  GIT_TAG        fix/cython_rpath_origin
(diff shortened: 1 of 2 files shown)


vyasr:
@robertmaynard I ended up choosing an approach that balances the simplicity of setting a default RPATH with giving users the flexibility to easily override it if they want.

robertmaynard (approved):
Approved presuming we drop the RAPIDS.cmake changes.

vyasr:
@robertmaynard done. I've requested write access to this repo, but until I get it you will also need to update the labels before we can merge.

robertmaynard:
@gpucibot merge

PR #190, opened 2022-05-12
Add rapids_cpm_nvcomp with prebuilt binary support

robertmaynard:
Expands the versions.json support to include the concept of per cpu architecture pre-built proprietary binary files that are used instead of building from source.

Currently the only package that supports this json entry is `nvcomp` which has also been added to `rapids_cpm`.

### DIFF
--- a/rapids-cmake/cpm/detail/get_proprietary_binary.cmake
+++ b/rapids-cmake/cpm/detail/get_proprietary_binary.cmake
@@ -33,7 +33,7 @@ the current CPU target architecture ( x86_64, aarch64, etc )
 function(rapids_cpm_get_proprietary_binary package_name version)
 
   get_property(override_json_data GLOBAL PROPERTY rapids_cpm_${package_name}_override_json)
-  # need to search the `proprietary_binary` dictonary for a key with the same name as
+  # need to search the `proprietary_binary` dictionary for a key with the same name as
   # lower_case(`CMAKE_SYSTEM_PROCESSOR-CMAKE_SYSTEM_NAME`).
   set(key "${CMAKE_SYSTEM_PROCESSOR}-${CMAKE_SYSTEM_NAME}")
   string(TOLOWER ${key} key)
--- a/rapids-cmake/cpm/detail/get_proprietary_binary.cmake
+++ b/rapids-cmake/cpm/detail/get_proprietary_binary.cmake
@@ -1,0 +1,77 @@
+#=============================================================================
+# Copyright (c) 2022, NVIDIA CORPORATION.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#=============================================================================
+include_guard(GLOBAL)
+
+#[=======================================================================[.rst:
+get_proprietary_binary
+-------------------
+
+.. versionadded:: v22.06.00
+
+Download the associated proprietary binary for the given project based on
+the current CPU target architecture ( x86_64, aarch64, etc )
+
+ .. note::
+  if override => the proprietary entry only in the override will be evaluated
+  if no override => the proprietary entry only in the default will be evaluated
+
+
+#]=======================================================================]
+function(rapids_cpm_get_proprietary_binary package_name version)
+
+  get_property(override_json_data GLOBAL PROPERTY rapids_cpm_${package_name}_override_json)
+  # need to search the `proprietary_binary` dictonary for a key with the same name as
+  # lower_case(`CMAKE_SYSTEM_PROCESSOR-CMAKE_SYSTEM_NAME`).
+  set(key "${CMAKE_SYSTEM_PROCESSOR}-${CMAKE_SYSTEM_NAME}")
+  string(TOLOWER ${key} key)
+
+  if(override_json_data)
+    string(JSON value ERROR_VARIABLE have_error GET "${override_json_data}" "proprietary_binary"
+           "${key}")
+  else()
+    get_property(json_data GLOBAL PROPERTY rapids_cpm_${package_name}_json)
+    string(JSON value ERROR_VARIABLE have_error GET "${json_data}" "proprietary_binary" "${key}")
+  endif()
+
+  if(have_error)
+    message(VERBOSE
+            "${package_name} requested usage of a proprietary_binary but none exist for ${CMAKE_SYSTEM_PROCESSOR}"
+    )
+    return()
+  endif()
+
+  # proprietary_binary now is the url that we need to download
+  set(proprietary_binary ${value})
+
+  if(NOT DEFINED rapids-cmake-version)
+    include("${rapids-cmake-dir}/rapids-version.cmake")
+  endif()
+
+  # Evaluate any magic placeholders in the proprietary_binary value including the
+  # `rapids-cmake-version` value
+  cmake_language(EVAL CODE "set(proprietary_binary ${proprietary_binary})")
+
+  if(proprietary_binary)
+    # download and extract the binaries since they don't exist on the machine
+    include(FetchContent)
+    set(pkg_name "${package_name}_proprietary_binary")
+    FetchContent_Declare(${pkg_name} URL ${proprietary_binary})
+    FetchContent_MakeAvailable(${pkg_name})
+
+    # Tell the subsequent rapids_cpm_find where to search so that it uses this binary
+    set(${package_name}_ROOT "${${pkg_name}_SOURCE_DIR}" PARENT_SCOPE)
+  endif()
+endfunction()
--- a/rapids-cmake/cpm/detail/get_proprietary_binary.cmake
+++ b/rapids-cmake/cpm/detail/get_proprietary_binary.cmake
@@ -39,11 +39,12 @@ function(rapids_cpm_get_proprietary_binary package_name version)
   string(TOLOWER ${key} key)
 
   if(override_json_data)
-    string(JSON value ERROR_VARIABLE have_error GET "${override_json_data}" "proprietary_binary"
-           "${key}")
+    string(JSON proprietary_binary ERROR_VARIABLE have_error GET "${override_json_data}"
+           "proprietary_binary" "${key}")
   else()
     get_property(json_data GLOBAL PROPERTY rapids_cpm_${package_name}_json)
-    string(JSON value ERROR_VARIABLE have_error GET "${json_data}" "proprietary_binary" "${key}")
+    string(JSON proprietary_binary ERROR_VARIABLE have_error GET "${json_data}"
+           "proprietary_binary" "${key}")
   endif()
 
   if(have_error)
@@ -52,9 +53,6 @@ function(rapids_cpm_get_proprietary_binary package_name version)
     )
     return()
   endif()
-
-  # proprietary_binary now is the url that we need to download
-  set(proprietary_binary ${value})
 
   if(NOT DEFINED rapids-cmake-version)
     include("${rapids-cmake-dir}/rapids-version.cmake")
--- a/rapids-cmake/cpm/detail/get_proprietary_binary.cmake
+++ b/rapids-cmake/cpm/detail/get_proprietary_binary.cmake
@@ -31,18 +31,21 @@ the current CPU target architecture ( x86_64, aarch64, etc )
 
 #]=======================================================================]
 function(rapids_cpm_get_proprietary_binary package_name version)
+  list(APPEND CMAKE_MESSAGE_CONTEXT "rapids.cpm.rapids_cpm_get_proprietary_binary")
 
-  get_property(override_json_data GLOBAL PROPERTY rapids_cpm_${package_name}_override_json)
+  include("${rapids-cmake-dir}/cpm/detail/get_default_json.cmake")
+  include("${rapids-cmake-dir}/cpm/detail/get_override_json.cmake")
+  get_default_json(${package_name} json_data)
+  get_override_json(${package_name} override_json_data)
+
   # need to search the `proprietary_binary` dictionary for a key with the same name as
   # lower_case(`CMAKE_SYSTEM_PROCESSOR-CMAKE_SYSTEM_NAME`).
   set(key "${CMAKE_SYSTEM_PROCESSOR}-${CMAKE_SYSTEM_NAME}")
   string(TOLOWER ${key} key)
-
   if(override_json_data)
     string(JSON proprietary_binary ERROR_VARIABLE have_error GET "${override_json_data}"
            "proprietary_binary" "${key}")
   else()
-    get_property(json_data GLOBAL PROPERTY rapids_cpm_${package_name}_json)
     string(JSON proprietary_binary ERROR_VARIABLE have_error GET "${json_data}"
            "proprietary_binary" "${key}")
   endif()
--- a/rapids-cmake/cpm/detail/get_proprietary_binary.cmake
+++ b/rapids-cmake/cpm/detail/get_proprietary_binary.cmake
@@ -73,5 +73,6 @@ function(rapids_cpm_get_proprietary_binary package_name version)
 
     # Tell the subsequent rapids_cpm_find where to search so that it uses this binary
     set(${package_name}_ROOT "${${pkg_name}_SOURCE_DIR}" PARENT_SCOPE)
+    set(${package_name}_proprietary_binary ON PARENT_SCOPE)
   endif()
 endfunction()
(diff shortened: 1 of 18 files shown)


robertmaynard:
Thinking about this more, the `proprietary_binary` keys need to encode the operating system as well. So we need something like `<arch>-[linux|win|osx]`

robertmaynard:
Currently trying this out in cudf and realized that I need to add some more install rules.

rapids-cmake/cpm/nvcomp.cmake
```diff
@@ -0,0 +1,141 @@
...
+  # Fix up RAPIDS_UNPARSED_ARGUMENTS to have EXPORT_SETS As this is need for rapids_cpm_find
+  if(RAPIDS_INSTALL_EXPORT_SET)
+    list(APPEND RAPIDS_UNPARSED_ARGUMENTS INSTALL_EXPORT_SET ${RAPIDS_INSTALL_EXPORT_SET})
+  endif()
+  if(RAPIDS_BUILD_EXPORT_SET)
+    list(APPEND RAPIDS_UNPARSED_ARGUMENTS BUILD_EXPORT_SET ${RAPIDS_BUILD_EXPORT_SET})
+  endif()
+
+  set(to_exclude OFF)
+  if(NOT RAPIDS_INSTALL_EXPORT_SET OR exclude)
```
vyasr, line 83: Is `exclude` a variable in the parent scope being inherited here? I don't see it defined anywhere locally.

rapids-cmake/cpm/detail/get_proprietary_binary.cmake
```diff
@@ -0,0 +1,78 @@
...
+
+ .. note::
+  if override => the proprietary entry only in the override will be evaluated
+  if no override => the proprietary entry only in the default will be evaluated
+
+
+#]=======================================================================]
+function(rapids_cpm_get_proprietary_binary package_name version)
+
+  get_property(override_json_data GLOBAL PROPERTY rapids_cpm_${package_name}_override_json)
```
vyasr, line 35: Could we create a function `get_package_override` in `cpm/detail` to encapsulate this functionality? The function would be tiny, just checking if the override json file existed and querying it if so, maybe setting a variable in the parent scope or something to indicate what was found.

That would be cleaner than this function needing to know exactly how `rapids_cpm_package_override` transmits information about overridden packages (by setting a global variable). Even if the global is the only way to communicate that to callers of the function, we could at least hide more of the details.

rapids-cmake/cpm/detail/get_proprietary_binary.cmake
```diff
@@ -0,0 +1,78 @@
...
+  if(override_json_data)
+    string(JSON value ERROR_VARIABLE have_error GET "${override_json_data}" "proprietary_binary"
+           "${key}")
+  else()
+    get_property(json_data GLOBAL PROPERTY rapids_cpm_${package_name}_json)
+    string(JSON value ERROR_VARIABLE have_error GET "${json_data}" "proprietary_binary" "${key}")
+  endif()
+
+  if(have_error)
+    message(VERBOSE
```
vyasr, line 50: Should we fail harder here? Or at least, propagate sufficient information for the parent scope to fail harder? As it stands now, if a user passes `USE_PROPERIETARY_BINARY` to `rapids_cpm_nvcomp` and no binary is found, there will just be a small message. Shouldn't `rapids_cpm_nvcomp` fail in that scenario (even if `rapids_cpm_get_proprietary_binary` shouldn't)?

rapids-cmake/cpm/detail/get_proprietary_binary.cmake
```diff
@@ -0,0 +1,78 @@
...
+            "${package_name} requested usage of a proprietary_binary but none exist for ${CMAKE_SYSTEM_PROCESSOR}"
+    )
+    return()
+  endif()
+
+  # proprietary_binary now is the url that we need to download
+  set(proprietary_binary ${value})
```
vyasr, line 57: Why not set this variable directly in the `string` commands above? The message/return is fairly self-documenting that `proprietary_binary` may not be well-defined if `have_error` is `TRUE`.

rapids-cmake/cpm/detail/get_proprietary_binary.cmake
```diff
@@ -0,0 +1,78 @@
...
+
+  if(NOT DEFINED rapids-cmake-version)
+    include("${rapids-cmake-dir}/rapids-version.cmake")
+  endif()
+
+  # Evaluate any magic placeholders in the proprietary_binary value including the
+  # `rapids-cmake-version` value
+  cmake_language(EVAL CODE "set(proprietary_binary ${proprietary_binary})")
```
vyasr, line 65: Would it be possible to instead evaluate a subset of properties across all of versions.json inside `rapids_cpm_load_preset_versions`? Or will that fail because the variables that aren't defined will get replaced with an empty string, e.g. `${version}` will end up empty because only `rapids-cmake-version` can be defined at the point that `rapids_cpm_load_preset_versions` is called? Trying to find ways to remove confusing non-local boilerplate code.

rapids-cmake/cpm/nvcomp.cmake
```diff
@@ -0,0 +1,141 @@
...
+``INSTALL_EXPORT_SET``
+  Record a :cmake:command:`find_dependency(<PackageName> ...) <cmake:module:CMakeFindDependencyMacro>` call needs to occur as part of
+  our build directory export set.
+
+.. note::
+  Installation of nvcomp will occur if an INSTALL_EXPORT_SET is provided, and nvcomp
+  is added to the project via :cmake:command:`add_subdirectory <cmake:command:add_subdirectory>` by CPM.
+
+``USE_PROPRIETARY_BINARY``
+  By enabling this flag you and using the software, you agree to fully comply with the terms and conditions of
```
vyasr, line 50:
```suggestion
  By enabling this flag and using the software, you agree to fully comply with the terms and conditions of
```

rapids-cmake/cpm/nvcomp.cmake
```diff
@@ -0,0 +1,141 @@
...
+#]=======================================================================]
+function(rapids_cpm_nvcomp)
+  list(APPEND CMAKE_MESSAGE_CONTEXT "rapids.cpm.nvcomp")
+
+  set(options)
+  set(one_value USE_PROPRIETARY_BINARY BUILD_EXPORT_SET INSTALL_EXPORT_SET)
+  set(multi_value)
+  cmake_parse_arguments(RAPIDS "${options}" "${one_value}" "${multi_value}" ${ARGN})
+
+  # Fix up RAPIDS_UNPARSED_ARGUMENTS to have EXPORT_SETS As this is need for rapids_cpm_find
```
vyasr, line 74:
```suggestion
  # Fix up RAPIDS_UNPARSED_ARGUMENTS to have EXPORT_SETS as this is need for rapids_cpm_find
```

rapids-cmake/cpm/nvcomp.cmake
```diff
@@ -0,0 +1,141 @@
...
+  set(to_exclude OFF)
+  if(NOT RAPIDS_INSTALL_EXPORT_SET OR exclude)
+    set(to_exclude ON)
+  endif()
+
+  include("${rapids-cmake-dir}/cpm/detail/package_details.cmake")
+  rapids_cpm_package_details(nvcomp version repository tag shallow exclude)
+
+  # first see if we have a proprietary pre-built binary listed in versions.json and it if requested.
+  #
```
vyasr, line 91:
```suggestion
```

rapids-cmake/cpm/nvcomp.cmake
```diff
@@ -0,0 +1,141 @@
...
+
+Allow projects to find or build `nvComp` via `CPM` with built-in
+tracking of these dependencies for correct export support.
+
+Uses the version of nvComp :ref:`specified in the version file <cpm_versions>` for consistency
+across all RAPIDS projects.
+
+.. code-block:: cmake
+
+  rapids_cpm_nvcomp( [BUILD_EXPORT_SET <export-name>]
```
vyasr, line 32: Do we need to document that all unparsed arguments are forwarded to `rapids_cpm_find`, or is that something that should be obvious?

rapids-cmake/cpm/nvcomp.cmake
```diff
@@ -0,0 +1,141 @@
...
+                  GLOBAL_TARGETS nvcomp::nvcomp
+                  CPM_ARGS
+                  GIT_REPOSITORY ${repository}
+                  GIT_TAG ${tag}
+                  GIT_SHALLOW ${shallow}
+                  EXCLUDE_FROM_ALL ${to_exclude}
+                  OPTIONS "BUILD_STATIC ON" "BUILD_TESTS OFF" "BUILD_BENCHMARKS OFF"
+                          "BUILD_EXAMPLES OFF")
+
+  # nvcomp creates the correct namespace aliases
```
vyasr, line 109: Then why do we need this block? Wouldn't needing this block indicate an improperly configured nvcomp?

rapids-cmake/cpm/nvcomp.cmake
```diff
@@ -0,0 +1,141 @@
...
+  endif()
+
+  # Propagate up variables that CPMFindPackage provide
+  set(nvcomp_SOURCE_DIR "${nvcomp_SOURCE_DIR}" PARENT_SCOPE)
+  set(nvcomp_BINARY_DIR "${nvcomp_BINARY_DIR}" PARENT_SCOPE)
+  set(nvcomp_ADDED "${nvcomp_ADDED}" PARENT_SCOPE)
+  set(nvcomp_VERSION ${version} PARENT_SCOPE)
+  set(nvcomp_proprietary_binary ${nvcomp_proprietary_binary} PARENT_SCOPE)
+
+  # Setup up install rules when using the proprietary_binary When building from source, nvcomp will
```
vyasr, line 121:
```suggestion
  # Set up up install rules when using the proprietary_binary. When building from source, nvcomp will
```

testing/cpm/cpm_nvcomp-export.cmake
```diff
@@ -0,0 +1,31 @@
+#=============================================================================
+# Copyright (c) 2021, NVIDIA CORPORATION.
```
vyasr, line 2:
```suggestion
# Copyright (c) 2022, NVIDIA CORPORATION.
```

rapids-cmake/cpm/nvcomp.cmake
```diff
@@ -0,0 +1,141 @@
...
+    install(FILES "${nvcomp_ROOT}/NOTICE" DESTINATION info/licenses/ RENAME NVCOMP_NOTICE)
+    install(FILES "${nvcomp_ROOT}/LICENSE" DESTINATION info/licenses/ RENAME NVCOMP_LICENSE)
+  endif()
+
+  if(RAPIDS_BUILD_EXPORT_SET AND nvcomp_proprietary_binary)
+    # point our consumers to where they can find the pre-built version
+    rapids_export_find_package_root(BUILD nvcomp "${nvcomp_ROOT}" ${RAPIDS_BUILD_EXPORT_SET})
+  elseif(RAPIDS_BUILD_EXPORT_SET)
+    # point our consumers to where they can find the source version
+    rapids_export_find_package_root(BUILD nvcomp "${nvcomp_BINARY_DIR}" ${RAPIDS_BUILD_EXPORT_SET})
```
vyasr, line 138: Does `rapids_cpm_find` not handle this case?

testing/cpm/cpm_nvcomp-invalid-arch.cmake
```diff
@@ -0,0 +1,30 @@
+#=============================================================================
+# Copyright (c) 2021, NVIDIA CORPORATION.
```
vyasr, line 2:
```suggestion
# Copyright (c) 2022, NVIDIA CORPORATION.
```

testing/cpm/cpm_nvcomp-override-clears-proprietary_binary.cmake
```diff
@@ -0,0 +1,46 @@
+#=============================================================================
+# Copyright (c) 2021, NVIDIA CORPORATION.
```
vyasr, line 2:
```suggestion
# Copyright (c) 2022, NVIDIA CORPORATION.
```

testing/cpm/cpm_nvcomp-override-clears-proprietary_binary.cmake
```diff
@@ -0,0 +1,46 @@
...
+
+
+rapids_cpm_init()
+
+if(TARGET nvcomp::nvcomp)
+  message(FATAL_ERROR "Expected nvcomp::nvcomp expected to not exist")
+endif()
+
+# Need to write out an nvcomp override file
+file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/override.json
```
vyasr, line 28: Should we clean up the file after the test completes?

vyasr (changes requested):
The logic in this PR looks correct. My comments are all either typo fixes or questions/suggestions for improving the clarity of the code. My suggestions aren't entirely relevant to the scope of this PR, so if you did think they were worthwhile we could push that to a follow-up.

Big picture, maybe this is just the best way to work within CMake's design constraints, but is there any way to rely less on the setting of global state? I had to check out this branch and inspect with some grepping to figure out how versions.json and the override data was pulled into a global variable and then made available. Some of my suggestions involve capturing the checks of that global state into functions.

rapids-cmake/cpm/detail/get_proprietary_binary.cmake
robertmaynard: We defer evaluation of placeholders as long as possible. That allows overrides to update only something like the `version` value and have that reflected in evaluations of other fields.

If we went to an eager model we would need to re-evaluate all previous entries excluding the keys that are part of the override set, which seemed harder to implement.

rapids-cmake/cpm/nvcomp.cmake
robertmaynard: That is the existing behavior across all the `rapids_cpm_<pkg>` commands.

I think it is worthwhile to do that as a separate PR.
I have added an issue for this https://github.com/rapidsai/rapids-cmake/issues/191

rapids-cmake/cpm/nvcomp.cmake
robertmaynard: This needs to go after rapids_cpm_package_details. Corrected

rapids-cmake/cpm/nvcomp.cmake
robertmaynard: Droped, you are correct it isn't needed.

rapids-cmake/cpm/detail/get_proprietary_binary.cmake
robertmaynard: After talking with vysar offline  we agreed more documentation is needed in `USE_PROPRIETARY_BINARY` to explain the behavior of the command when an override exists, or if no binary exists for platform ( arm / windows / etc ).

rapids-cmake/cpm/detail/get_proprietary_binary.cmake
robertmaynard: I have updated this PR to move the loading from a global property into a function.

robertmaynard: Does this look better with the refactoring?

vyasr: Yes I like the `get_default_json` and `get_override_json`, that definitely helps!

rapids-cmake/cpm/nvcomp.cmake
vyasr: I'm still a bit confused by this one.

vyasr:
I don't yet have any permissions on this repo so I can't resolve threads, so I just looked through all the threads. All of my concerns look to have been addressed except for https://github.com/rapidsai/rapids-cmake/pull/190#discussion_r872628237 and https://github.com/rapidsai/rapids-cmake/pull/190#discussion_r872633554.

testing/cpm/cpm_nvcomp-override-clears-proprietary_binary.cmake
robertmaynard: This is a pattern I learned from CMake. We keep the test dir in a 'dirty' state so we have an exact record of what was executed. Helps diagnose issues since we can upload an exact state of a failure if needed

rapids-cmake/cpm/nvcomp.cmake
robertmaynard: When nvcomp is built from source it doesn't create a nvcomp::nvcomp alias just an add_library `nvcomp` target. rapids-cmake adds the alias so that all consumers of nvcomp just need a singular code path where they reference `nvcomp::nvcomp`

robertmaynard:
> I don't yet have any permissions on this repo so I can't resolve threads, so I just looked through all the threads. All of my concerns look to have been addressed except for [#190 (comment)](https://github.com/rapidsai/rapids-cmake/pull/190#discussion_r872628237) and [#190 (comment)](https://github.com/rapidsai/rapids-cmake/pull/190#discussion_r872633554).

I believe I addressed these two outstanding comments.

testing/cpm/cpm_nvcomp-override-clears-proprietary_binary.cmake
vyasr: Ah I see.

vyasr (approved):

robertmaynard:
@gpucibot merge

PR #194, opened 2022-05-18
nvcomp install rules need to match the pre-built layout

robertmaynard:
This ensures that the library paths referenced in nvcomp-target.cmake are kept correct.


jbrennan333 (approved):
+1 this looks good to me.

robertmaynard:
@gpucibot merge

PR #198, opened 2022-05-25
Add support for a prefix in Cython module targets

vyasr:
A package composed of subpackages may contain multiple Cython modules with the same filename at different paths i.e. in different parts of the package directory tree. Since we use the filenames to construct the CMake library targets, the resulting library target names will conflict unless there is a way to customize them. This PR adds that functionality. Note that because of the way that scikit-build's Cython is designed, we need to ensure that the generated shared library is named  according to the original filename; the prefix should only be applied to the CMake target to avoid conflicts within CMake.

### DIFF
--- a/testing/cython/create_modules/CMakeLists.txt
+++ b/testing/cython/create_modules/CMakeLists.txt
@@ -40,5 +40,5 @@ rapids_cython_create_modules(
     )
 
 if(NOT TARGET test)
-  message(FATAL_ERROR "rapids_cython_init didn't create the prefixed target test")
+  message(FATAL_ERROR "rapids_cython_init didn't create the target `test`")
 endif()
--- a/testing/cython/create_modules/CMakeLists.txt
+++ b/testing/cython/create_modules/CMakeLists.txt
@@ -38,3 +38,17 @@ rapids_cython_create_modules()
 rapids_cython_create_modules(
     SOURCE_FILES test.pyx
     )
+
+if(NOT TARGET test)
+  message(FATAL_ERROR "rapids_cython_init didn't create the prefixed target test")
+endif()
+
+# Test that we can use a prefix to generate a different library target.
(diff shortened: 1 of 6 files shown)


testing/cython/create_modules/CMakeLists.txt
```diff
@@ -38,3 +38,7 @@ rapids_cython_create_modules()
 rapids_cython_create_modules(
     SOURCE_FILES test.pyx
     )
+
+if(NOT TARGET test)
+  message(FATAL_ERROR "rapids_cython_init didn't create the prefixed target test")
```
robertmaynard, line 43:
```suggestion
  message(FATAL_ERROR "rapids_cython_init didn't create the target `test`")
```

testing/cython/create_modules_with_prefix/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+rapids_cython_create_modules()
+
+# Test that we can use a prefix to generate a different library target.
+rapids_cython_create_modules(
+    SOURCE_FILES test.pyx
+    MODULE_PREFIX "test_"
+    )
+
+if(NOT TARGET test_test)
+  message(FATAL_ERROR "rapids_cython_init didn't create the prefixed library target test::test")
```
robertmaynard, line 44:
```suggestion
  message(FATAL_ERROR "rapids_cython_init didn't create the prefixed library target `test_test`")
```

robertmaynard (commented):
Overall looks great.

robertmaynard (approved):
LGTM, merge whenever you want

vyasr:
Thanks @robertmaynard. I applied your suggestions and undid the changes to RAPIDS.cmake, so we should be good to merge unless you spot anything else on a final review.

robertmaynard:
@gpucibot merge

PR #201, opened 2022-06-16
Adds support for pulling cuCollections using rapids-cmake

vyasr:
This PR allows us to standardize cuco usage across RAPIDS.


robertmaynard (changes requested):
We need to update the docs to expose `rapids_cpm_cuco`

So add a rapids_cpm_cuco.rst to `docs/packages`  and add an entry to `docs/api.rst`

rapids-cmake/cpm/cuco.cmake
```diff
@@ -0,0 +1,88 @@
...
+                   )
+
+``BUILD_EXPORT_SET``
+  Record that a :cmake:command:`CPMFindPackage(cuco)` call needs to occur as part of
+  our build directory export set.
+
+``INSTALL_EXPORT_SET``
+  Record a :cmake:command:`find_dependency(cuco) <cmake:module:CMakeFindDependencyMacro>` call needs to occur as part of
+  our install directory export set.
+
```
robertmaynard, line 43: cuco always produces install rules so we need to have a note here like we do with rmm that installation will always occur.

rapids-cmake/cpm/cuco.cmake
```diff
@@ -0,0 +1,88 @@
...
+                  GIT_REPOSITORY ${repository}
+                  GIT_TAG ${tag}
+                  GIT_SHALLOW ${shallow}
+                  # TODO: I'm not sure if we want to make this configurable. cudf currently sets
+                  # this based on a separate variable, but raft does not, and doing so isn't
+                  # consistent with other rapids-cmake cpm find commands. cudf later manually runs a
+                  # rapids_export_package instead.
+                  EXCLUDE_FROM_ALL ${exclude})
+
+  if(CUCO_BUILD_EXPORT_SET)
```
robertmaynard, line 73: I dont believe the handling of `BUILD_EXPORT_SET` and `INSTALL_EXPORT_SET` are required since cuco like rmm already generates a `cuco-config.cmake` for both the build and install tree.

testing/cpm/CMakeLists.txt
```diff
@@ -59,3 +59,5 @@ add_cmake_config_test( cpm_spdlog-simple.cmake )
 
 add_cmake_config_test( cpm_thrust-export.cmake )
 add_cmake_config_test( cpm_thrust-simple.cmake )
+
+add_cmake_config_test( cpm_cuco.cmake )
```
robertmaynard, line 63: Please rename this to `cpm_cuco-simple.cmake` and add a `cpm_cuco-export` to verify that build and install `EXPORT_SETS` work as expected.

robertmaynard (changes requested):

rapids-cmake/cpm/versions.json
```diff
@@ -39,6 +39,11 @@
       "version" : "1.7.0",
       "git_url" : "https://github.com/NVIDIA/libcudacxx.git",
       "git_tag" : "${version}"
+    },
+    "cuco" : {
+      "version" : "0.0.1",
+      "git_url" : "https://github.com/NVIDIA/cuCollections.git",
+      "git_tag" : "55029034c3f82bca36148c9be29941b37492394d"
```
bdice, line 46: Is there a reason we don't use shallow git clones for packages in `versions.json`?

rapids-cmake/cpm/cuco.cmake
vyasr: Ah yes I will double-check but I think you are right.

rapids-cmake/cpm/versions.json
vyasr: Clones are shallow by default, unless versions.json specifies otherwise.

bdice: Awesome. I looked briefly but didn't see where that was defaulted. I had some issues with CPM timing out earlier this week when my internet was being flaky. I assumed it was due to full clones.

robertmaynard: The default behavior for shallow vs non is here: https://docs.rapids.ai/api/rapids-cmake/nightly/packages/rapids_cpm_versions.html#cpm-versions

rapids-cmake/cpm/versions.json
```diff
@@ -39,6 +39,11 @@
       "version" : "1.7.0",
       "git_url" : "https://github.com/NVIDIA/libcudacxx.git",
       "git_tag" : "${version}"
+    },
+    "cuco" : {
+      "version" : "0.0.1",
```
robertmaynard, line 44: Since we are checking out an abitrary git SHA1 we need to specify `git_shallow` to be `false`. Since we can't presume that SHA1 will always be equal to `HEAD`, a named tag or tip of a branch.

robertmaynard (changes requested):

rapids-cmake/cpm/cuco.cmake
```diff
@@ -0,0 +1,86 @@
...
+  cuco::cuco target will be created
+
+#]=======================================================================]
+function(rapids_cpm_cuco)
+  list(APPEND CMAKE_MESSAGE_CONTEXT "rapids.cpm.cuco")
+
+  set(options)
+  set(one_value INSTALL_EXPORT_SET)
+  set(multi_value)
+  cmake_parse_arguments(CUCO "${options}" "${one_value}" "${multi_value}" ${ARGN})
```
robertmaynard, line 55: Please make the prefix `_RAPIDS` to match the changes in https://github.com/rapidsai/rapids-cmake/pull/203

testing/cpm/cpm_cuco-exports.cmake
```diff
@@ -0,0 +1,26 @@
+#=============================================================================
+# Copyright (c) 2021, NVIDIA CORPORATION.
```
robertmaynard, line 2:
```suggestion
# Copyright (c) 2022, NVIDIA CORPORATION.
```

testing/cpm/cpm_cuco-simple.cmake
```diff
@@ -0,0 +1,31 @@
+#=============================================================================
+# Copyright (c) 2021, NVIDIA CORPORATION.
```
robertmaynard, line 2:
```suggestion
# Copyright (c) 2022, NVIDIA CORPORATION.
```

robertmaynard (changes requested):

rapids-cmake/cpm/cuco.cmake
vyasr: I just saw that PR and commented on that choice there, then I saw this comment here. I will wait on that discussion getting finished up before changing this.

rapids-cmake/cpm/versions.json
bdice: > If no such field exists the default is false.

That conflicts with @vyasr's statement above that clones are shallow (true) by default. 🤔

bdice: The docs are lying, it's true by default. I'll fix this. https://github.com/rapidsai/rapids-cmake/blob/464362f90b122b8a112e1675626fd251676544bb/rapids-cmake/cpm/detail/package_details.cmake#L62-L64

robertmaynard (approved):
Thank you!

testing/cpm/cpm_cuco-simple.cmake
```diff
@@ -0,0 +1,31 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#=============================================================================
+include(${rapids-cmake-dir}/cpm/init.cmake)
+include(${rapids-cmake-dir}/cpm/cuco.cmake)
+
+rapids_cpm_init()
+
+if(TARGET cuco::cuco)
+  message(FATAL_ERROR "Expected cuco::cuco expected to not exist")
```
bdice, line 22: This error message doesn't make sense to me, but I might be reading the sentence wrong.

bdice (commented):
This is a nice change. Comments attached. (I reviewed this to help familiarize myself with more of rapids-cmake.)

testing/cpm/cpm_cuco-simple.cmake
vyasr: The target shouldn't exist until after the first call to `rapids_cpm_cuco` below.

vyasr: Looks like there is a typo with an extra "expected" though. I'll fix.

bdice: Yup. Raising the error makes sense -- the message just had an [unexpected expected](https://giphy.com/clips/ralph-dramatic-turn-hamster-2eeltIELxgHQ9Y5iA3).

rapids-cmake/cpm/versions.json
bdice: Oh, right. I hadn't thought about how this calls through to CPM / CMake itself. 🤦‍♂️ [This comment in CPM.cmake](https://github.com/cpm-cmake/CPM.cmake/blob/a575ab57b7e7449c896ae6fed72808ab91088954/cmake/CPM.cmake#L698-L699) indicates that hashes can't be shallow cloned by CPM. That is echoed in [the CMake docs](https://cmake.org/cmake/help/latest/module/ExternalProject.html?highlight=git_shallow):

> If `GIT_SHALLOW` is enabled then `GIT_TAG` works only with branch names and tags. A commit hash is not allowed.

[Searching the CMake source for `git_clone_options`](https://gitlab.kitware.com/search?search=git_clone_options&nav_source=navbar&project_id=541&group_id=415&search_code=true&repository_ref=master) shows where this code is used.

vyasr:
@gpucibot merge

PR #203, opened 2022-06-16
rapids-cmake won't ever use an existing variable starting with RAPIDS_

### LINKED ISSUE #200: [BUG] `rapids_export` Does Not Work if the Variable `RAPIDS_VERSION` Exists
**Describe the bug**
If you have the cache variable `RAPIDS_VERSION` set before calling `rapids_export()`, the version it chooses can be incorrect if you omit the `VERSION` argument or set `VERSION` with no value. In these situations, the documentation states it will default to use `PROJECT_VERSION`. Instead it simply uses the value in `RAPIDS_VERSION`.

We ran into this in SRF and needed to explicitly set the version to get around the bug (here is the PR with the workaround: https://github.com/nv-morpheus/SRF/pull/30). In SRF, we use the cache variable `RAPIDS_VERSION` to indicate which version of RMM, RAPIDS CMake, etc. we are using to build with. It took a while to discover why the generated `srf-config-version.cmake` was using the incorrect version value.

The following pseudocode should reproduce the bug:

```cmake
project(my_project
  VERSION 22.06.00
  LANGUAGES C CXX
)

# Need to set the RAPIDS_VERSION cache variable
set(RAPIDS_VERSION "21.10" CACHE STRING "RAPIDS Libraries Version")

# This will create a version file with 21.10 as the version, not 22.06
rapids_export(INSTALL ${PROJECT_NAME}
  EXPORT_SET ${PROJECT_NAME}-core-exports
  GLOBAL_TARGETS libmy_project
  VERSION
  NAMESPACE my_project::
)
```
Also ommitting the `VERSION` argument will hit the bug as well.

**Expected behavior**
As indicated by the documentation, when `VERSION` has no value or is ommitted as an argument, `rapids_export` should default to using `PROJECT_VERSION`.

**Environment details (please complete the following information):**
 - CMake version 3.22

**Additional context**
This issue can be quickly fixed by changing the prefix used for `cmake_parse_arguments`. Currently the code uses `cmake_parse_arguments(RAPIDS ...)` which creates too common of variable names like `RAPIDS_VERSION`, `RAPIDS_NAMESPACE`, `RAPIDS_DOCUMENTATION`, etc. Instead the prefix should be something like `_RAPIDS` to prevent collisions.

robertmaynard:
Fixes #200


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -17,6 +17,8 @@ Here are some guidelines to help the review process go smoothly.
    - Each user facing function should be added to the `cmake-format.json` document
     - Run `cmake-genparsers -f json` on the `.cmake` file as a starting point
    - Each function first line should be `list(APPEND CMAKE_MESSAGE_CONTEXT "rapids.<component>.<function>")`
+   - Internal variables for things like `cmake_parse_arguments` should use `_RAPIDS`
```
vyasr, line 20: Is it worth recommending a specific alternative rather than simply proscribing the use of `RAPIDS`? Personally I have been using more specific names in each PR that I make (e.g. `RAPIDS_CYTHON` or `CUCO`).

vyasr (approved):
One question regarding future use, but for now this global find and replace gets the job done.

.github/PULL_REQUEST_TEMPLATE.md
vyasr: Unless there's a reason that we actually _want_ variables from higher scopes to be visible in lower scopes using a common naming scheme?

robertmaynard: When I started with `RAPIDS` I did it as I felt that this prefix would be unused by any consumer of rapids-cmake. Therefore we wouldn't accidently de-reference a variable from the calling project. Issue #200 showed that other projects use variables that start with `RAPIDS` and therefore we moved to `_RAPIDS`.

I have the same concern with project specific prefixes like `CUCO`. The calling project could set variables like `CUCO_VERSION` that could influence rapids-cmake internal behavior.

vyasr: I see, yeah that's true. Maybe we should suggest following the same naming convention as the function names (e.g. `rapids_cpm_find` -> `RAPIDS_CPM_FIND_...`)? It obviously gets very long, though, so there's a significant tradeoff. I'll stick to `_RAPIDS` for now and if we ever run into internal conflicts I guess we can worry about this then. Worst case it's another find-and-replace job.

ajschmidt8 (approved):

robertmaynard:
@gpucibot merge

PR #204, opened 2022-06-16
Fix docs: default behavior is to use a shallow git clone.

bdice:
Fixes the documentation to indicate that git clones are shallow by default.


robertmaynard (approved):

robertmaynard:
@gpucibot merge

PR #209, opened 2022-06-30
Update nvcomp to 2.3.2

### LINKED ISSUE #208: [FEA] Update nvcomp version to 2.3.1
Currently cuDF is pulling in nvcomp release 2.3.   The DevTech team is in the process of releasing 2.3.1, which has a number of stability fixes.  We should update the nvcomp release to 2.3.1 once it is available.

robertmaynard:
Fixes #208

### DIFF
--- a/rapids-cmake/cpm/versions.json
+++ b/rapids-cmake/cpm/versions.json
@@ -13,11 +13,11 @@
       "git_tag" : "12d13bdc5e74801645eba3e46a64081b9b020dca"
     },
     "nvcomp" : {
-      "version" : "2.3.1",
+      "version" : "2.3.2",
       "git_url" : "https://github.com/NVIDIA/nvcomp.git",
       "git_tag" : "v2.2.0",
       "proprietary_binary" : {
-        "x86_64-linux" : "http://developer.download.nvidia.com/compute/nvcomp/${version}/local_installers/nvcomp_${version}_Linux_CUDA_11.x.tgz"
+        "x86_64-linux" : "https://developer.download.nvidia.com/compute/nvcomp/${version}/local_installers/nvcomp_install_CUDA_11.x.tgz"
       }
     },
     "rmm" : {
--- a/rapids-cmake/cpm/versions.json
+++ b/rapids-cmake/cpm/versions.json
@@ -13,7 +13,7 @@
       "git_tag" : "12d13bdc5e74801645eba3e46a64081b9b020dca"
     },
     "nvcomp" : {
-      "version" : "2.3",
+      "version" : "2.3.1",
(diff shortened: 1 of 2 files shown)


jbrennan333 (approved):
+1 this looks good to me.

rapids-cmake/cpm/versions.json
```diff
@@ -13,7 +13,7 @@
       "git_tag" : "12d13bdc5e74801645eba3e46a64081b9b020dca"
     },
     "nvcomp" : {
-      "version" : "2.3",
+      "version" : "2.3.1",
       "git_url" : "https://github.com/NVIDIA/nvcomp.git",
       "git_tag" : "v2.2.0",
```
bdice, line 18: 1. Why does this not match the `version`?
2. It doesn't look like a tag for `v2.3.1` exists yet. https://github.com/NVIDIA/nvcomp/tags

bdice: If this is related to the changes in source availability, we may want to avoid misleading users by posting the git repo URL and tag. If so, can those fields be removed?

robertmaynard: The `version` field is used to find the maximum latest version we support. So the 2.3.X series via binary download. If no binary download is usable ( arm platforms ) we fallback to building from source which means the 2.2 series as it is the last version with public sources.

robertmaynard:
rerun tests

rapids-cmake/cpm/versions.json
bdice: Great, thanks for the explanation. It would be nice to have a comment explaining that, but JSON doesn't support inline comments (at least in the official spec).

jbrennan333:
@robertmaynard, It appears the 2.3.2 has already been released:

https://developer.download.nvidia.com/compute/nvcomp/2.3.2/local_installers/nvcomp_install_CUDA_11.x.tgz

Can we update this PR to pull down 2.3.2 instead?

rapids-cmake/cpm/versions.json
```diff
@@ -13,7 +13,7 @@
       "git_tag" : "12d13bdc5e74801645eba3e46a64081b9b020dca"
     },
     "nvcomp" : {
-      "version" : "2.3",
+      "version" : "2.3.1",
```
jbrennan333, line 16: Looks like 2.3.2 is now released.  Can we please update this to 2.3.2?

jbrennan333 (approved):
+1 looks good. Thanks for updating!

bdice (approved):
Looks good to me!

vyasr:
@gpucibot merge

PR #214, opened 2022-07-13
[FIX] `rapids_find_package()` called with explicit version and REQUIRED should fail

trxcllnt:
Calling `rapids_find_package(<pkg> <version> EXACT REQUIRED)` should cause find_package to fail due to not finding the specified version, but it doesn't because the `REQUIRED` flag [is placed in front](https://github.com/rapidsai/rapids-cmake/blob/branch-22.08/rapids-cmake/find/package.cmake#L89) of the version in the `find_package()` call.


robertmaynard (approved):

robertmaynard:
@gpucibot merge

PR #215, opened 2022-07-14
Timestamps for URL downloads match the download time

robertmaynard:
By enabling CMake policy 135 we ensure that extracted files timestamp match that of the download time, instead of when the
archive is created. This makes sure that if the URL changes to an older version we still rebuild everything as the timestamp
stays newer.

https://cmake.org/cmake/help/latest/policy/CMP0135.html


vyasr (approved):

robertmaynard:
@gpucibot merge

PR #218, opened 2022-07-19
Remove nvcomp dependency on CUDA::cudart_static

robertmaynard:
This dependency isn't needed and stops conda users from building anything that depends on nvcomp


elstehle:
Thanks @robertmaynard for jumping on this. Confirmed with a local build that this fixes the issue 👍

PR #221, opened 2022-07-21
Update nvcomp to 2.3.3

### LINKED ISSUE #219: [FEA] Update nvcomp version to 2.3.3
Currently cuDF is pulling in nvcomp release 2.3.2. The DevTech team is in the process of releasing 2.3.3.

Once it is available, we should consider moving to it for release 22.08.
This version includes fixes for deflate:
```Fixed the issue of missing nvcompBatchedDeflateDecompressGetTempSizeEx.
Fixed the correctness issue of deflate compression.
Update the nvcompBatchedDeflateGetDecompressSizeAsync.
Add API tests for deflate, and change the file structure of deflate (Add template parameters)
```

Note that for cuDF, we worked around the problem with the missing `nvcompBatchedDeflateDecompressGetTempSizeEx` in PR: https://github.com/rapidsai/cudf/pull/11213

vyasr:
This PR updates nvcomp to version 2.3.3. Resolves #219

### DIFF
--- a/RAPIDS.cmake
+++ b/RAPIDS.cmake
@@ -21,8 +21,8 @@ set(rapids-cmake-version 22.08)
 include(FetchContent)
 FetchContent_Declare(
   rapids-cmake
-  GIT_REPOSITORY https://github.com/rapidsai/rapids-cmake.git
-  GIT_TAG        branch-${rapids-cmake-version}
+  GIT_REPOSITORY https://github.com/vyasr/rapids-cmake.git
+  GIT_TAG        feature/nvcomp_2.3.3
 )
 FetchContent_GetProperties(rapids-cmake)
 if(rapids-cmake_POPULATED)
--- a/RAPIDS.cmake
+++ b/RAPIDS.cmake
@@ -21,8 +21,8 @@ set(rapids-cmake-version 22.08)
 include(FetchContent)
 FetchContent_Declare(
   rapids-cmake
-  GIT_REPOSITORY https://github.com/vyasr/rapids-cmake.git
-  GIT_TAG        feature/nvcomp_2.3.3
+  GIT_REPOSITORY https://github.com/rapidsai/rapids-cmake.git
+  GIT_TAG        branch-${rapids-cmake-version}
 )
 FetchContent_GetProperties(rapids-cmake)
 if(rapids-cmake_POPULATED)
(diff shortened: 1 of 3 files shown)


vyasr:
Once I've tested against cudf we can make a final call on whether or not this gets into 22.08

robertmaynard (commented):
We should only activate this conditional for 2.3.2 :
https://github.com/robertmaynard/rapids-cmake/blob/1b29b2e4192f0c6e46416c711de4f7608f6f65e7/rapids-cmake/cpm/nvcomp.cmake#L98

vyasr:
@robertmaynard I think that I've done this correctly, but I'm not entirely familiar with how JSON parses into strings in CMake so let me know if there's an alternative comparison that would make sense here. Also, should we be doing something like `> 2.3.0 AND < 2.3.3` instead?

rapids-cmake/cpm/nvcomp.cmake
```diff
@@ -95,7 +95,7 @@ function(rapids_cpm_nvcomp)
 
     # Remove incorrect public dependency on the static cuda runtime We have to modify the
     # nvcomp-targets.cmake since these entries will cause a failure when rapids_cpm_find is called.
-    if(nvcomp_proprietary_binary)
+    if(nvcomp_proprietary_binary AND ${version} STREQUAL "2.3.2")
```
robertmaynard, line 98: We should use `VERSION_EQUAL` https://cmake.org/cmake/help/latest/command/if.html#version-equal

robertmaynard:
Please re-taret branch-22.10

rapids-cmake/cpm/nvcomp.cmake
vyasr, line 98:
```suggestion
    if(nvcomp_proprietary_binary AND ${version} VERSION_EQUAL "2.3.2")
```

vyasr:
@robertmaynard I think we are good to merge this now. https://github.com/rapidsai/cudf/pull/11327 was the test that this upgrade would be safe.

robertmaynard (approved):

robertmaynard:
Drop the updates to RAPIDS.cmake and you can merge away!

:shipit:

RAPIDS.cmake
```diff
@@ -21,8 +21,8 @@ set(rapids-cmake-version 22.08)
 include(FetchContent)
 FetchContent_Declare(
   rapids-cmake
-  GIT_REPOSITORY https://github.com/rapidsai/rapids-cmake.git
-  GIT_TAG        branch-${rapids-cmake-version}
+  GIT_REPOSITORY https://github.com/vyasr/rapids-cmake.git
+  GIT_TAG        feature/nvcomp_2.3.3
```
vyasr, line 25:
```suggestion
  GIT_REPOSITORY https://github.com/rapidsai/rapids-cmake.git
  GIT_TAG        branch-${rapids-cmake-version}
```

vyasr:
@gpucibot merge

PR #223, opened 2022-07-25
json exclude flag behaves as expected libcudacx//thrust/nvcomp

robertmaynard:
This brings the behavior of the exclude flag inline with the behavior of `EXCLUDE_FROM_ALL` when provided to the `add_subdirectory` call.

### DIFF
--- a/rapids-cmake/cpm/thrust.cmake
+++ b/rapids-cmake/cpm/thrust.cmake
@@ -98,7 +98,7 @@ function(rapids_cpm_thrust NAMESPACE namespaces_name)
   endif()
 
   # only install thrust when we have an in-source version
-  if(Thrust_SOURCE_DIR AND _RAPIDS_INSTALL_EXPORT_SET)
+  if(Thrust_SOURCE_DIR AND _RAPIDS_INSTALL_EXPORT_SET AND NOT exclude)
     #[==[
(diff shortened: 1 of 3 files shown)


robertmaynard:
This will help with packaging such as pip when we don't want any headers installed.

rapids-cmake/cpm/thrust.cmake
```diff
@@ -98,7 +98,7 @@ function(rapids_cpm_thrust NAMESPACE namespaces_name)
   endif()
 
   # only install thrust when we have an in-source version
-  if(Thrust_SOURCE_DIR AND _RAPIDS_INSTALL_EXPORT_SET)
+  if(Thrust_SOURCE_DIR AND _RAPIDS_INSTALL_EXPORT_SET AND NOT exclude)
```
vyasr, line 101: Just making sure, this is equivalent to the `to_exclude` var that's defined in other CMake files now, right? We just don't define that variable here because it's not needed elsewhere?

vyasr (approved):

rapids-cmake/cpm/thrust.cmake
robertmaynard: Correct

vyasr:
@gpucibot merge

PR #224, opened 2022-07-27
More robust solution of CMake policy 135

robertmaynard:
By enabling CMake policy 135 we ensure that extracted files timestamp match that of the download time, instead of when the
archive is created. This makes sure that if the URL changes to an older version we still rebuild everything as the timestamp
stays newer.


robertmaynard:
@gpucibot merge

PR #227, opened 2022-07-29
Update rapids-cmake to require cmake 3.23.1

robertmaynard:
This moves rapids-cmake to require CMake 3.23.1 or newer.

Along with that we are able to do the following improvements:

- Use CMake's FindGTest as it now provides all the targets we require
- Simplify install_lib_dir logic 
- Simplify logic around version values that are all zeroes


robertmaynard:
Due to https://gitlab.kitware.com/cmake/cmake/-/merge_requests/7523 We will need to change the get_gtest logic or maybe require 3.23.4+

robertmaynard:
Dropped the changes to thrust as those are now captured in: https://github.com/rapidsai/rapids-cmake/pull/231

vyasr:
@robertmaynard do you want me to review this now, or do you want to just wait for 3.24? We've been waiting on a CMake upgrade for a while anyway so I'm not opposed, but I thought we were expecting to see 3.24 in time for our 22.10 release anyway.

robertmaynard:
> @robertmaynard do you want me to review this now, or do you want to just wait for 3.24? We've been waiting on a CMake upgrade for a while anyway so I'm not opposed, but I thought we were expecting to see 3.24 in time for our 22.10 release anyway.

My thought was we could roll out these changes now, and verify that all consumers are still building correctly. That will give us time to have 3.24 released and conda packages provided. After that we can make another PR to bump the minimum required and update components with 3.24 improvements.

vyasr:
That's fine with me. But just FYI... https://github.com/conda-forge/cmake-feedstock/pull/165 :smile:

vyasr:
I'll review this afternoon

rapids-cmake/cmake/install_lib_dir.cmake
```diff
@@ -53,54 +53,25 @@ function(rapids_cmake_install_lib_dir out_variable_name)
...
+  # We need to defer to GNUInstallDirs but not allow it to set CMAKE_INSTALL_LIBDIR
+  set(remove_install_dir TRUE)
+  if(DEFINED CMAKE_INSTALL_LIBDIR)
+    set(remove_install_dir FALSE)
+  endif()
+
+  include(GNUInstallDirs)
+  set(computed_path "${CMAKE_INSTALL_LIBDIR}")
+  if(modify_install_libdir)
+    # GNUInstallDirs will have set `CMAKE_INSTALL_LIBDIR` as a cache path So we only need to make
```
bdice, line 68:
```suggestion
    # GNUInstallDirs will have set `CMAKE_INSTALL_LIBDIR` as a cache path so we only need to make
```

bdice (approved):
All seems good to me.

testing/cmake/write_git_revision-custom-prefix/CMakeLists.txt
```diff
@@ -15,7 +15,7 @@
 #=============================================================================
 include(${rapids-cmake-dir}/cmake/write_git_revision_file.cmake)
 
-cmake_minimum_required(VERSION 3.20)
+cmake_minimum_required(VERSION 3.23)
```
vyasr, line 18: Is there a reason that tests are not consistent with the source in requiring the patch version?

testing/export/export-verify-explicit-patch-version-value-of-zero.cmake
```diff
@@ -59,7 +59,7 @@ if(NOT "${TEST_VERSION_MAJOR}.${TEST_VERSION_MINOR}" VERSION_EQUAL 21.09)
   message(FATAL_ERROR "rapids_export failed to export version major/minor information")
 endif()
 
-find_package(test 21.9.0 EXACT REQUIRED)
+find_package(test 21.09.0 EXACT REQUIRED)
```
vyasr, line 62: Was this previously just a typo, or did something actually change?

testing/export/write_language-nested-dirs/CMakeLists.txt
```diff
@@ -15,7 +15,7 @@
 #=============================================================================
 include(${rapids-cmake-dir}/export/write_language.cmake)
 
-cmake_minimum_required(VERSION 3.20 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.23.1 FATAL_ERROR)
```
vyasr, line 18: Amending my previous question about tests... All tests except this one.

vyasr (approved):
Couple of very minor questions, otherwise LGTM.

testing/cmake/write_git_revision-custom-prefix/CMakeLists.txt
robertmaynard: The regex I use searched for `.20` and therefore I missed bumping the tests correctly.
I have updated all the tests to 3.23.1

testing/export/export-verify-explicit-patch-version-value-of-zero.cmake
robertmaynard: This is on purpose. EXACT matches requires string equivalant so the value needs to be `.9`

vyasr: Wait now I"m confused. You changed it from `.9` to `.09`. Which should it be?

robertmaynard: The logic around version generated by `rapids-cmake/export/detail/parse_version.cmake` has changed. We used to always drop leading zeroes from version values as it would break cmake.

Now since that is supported we export the version value as specified by the project ( 21.09.0 ) and as a result the `EXACT` query here needs to be the same.

robertmaynard:
@gpucibot merge

PR #230, opened 2022-08-01
Update to CPM v0.35.3 to support symlink build directories

### LINKED ISSUE #229: [FEA] Update To CPM v0.35.3
CPM v0.35.3 doesn't error out when the build directory is a symlink. This fix will help rapids-compose users.

robertmaynard:
Fixes #229


robertmaynard:
@gpucibot merge

PR #199, opened 2022-05-26
Update thrust tag to 1.17.0.

bdice:
This PR updates Thrust to version 1.17.0 (to be released soon) for RAPIDS 22.08. The tag format has been updated as noted in #181.

See also: https://github.com/rapidsai/cudf/issues/10841

Checklist:
- [x] Wait for Thrust 1.17.0 final release
- [ ] Test packages with updates for header includes:
  - [x] https://github.com/rapidsai/cuml/pull/4675
  - [x] https://github.com/rapidsai/cuspatial/pull/539
  - [x] https://github.com/NVIDIA/cuCollections/pull/161
  - [x] https://github.com/rapidsai/raft/pull/678
  - [x] https://github.com/rapidsai/cugraph/pull/2310
  - [ ] Make updated PR to cuDF with an updated patch for the sort algorithm, benchmark cudf

PR #231, opened 2022-08-01
Update to thrust 1.17

robertmaynard:
This pulls out the Thrust bump from https://github.com/rapidsai/rapids-cmake/pull/227 so that we can roll out changes slowly and have an easier time to track down issues.

Once this is in and RAPIDS is happy, we can go ahead and merge https://github.com/rapidsai/rapids-cmake/pull/199 which has breaking changes.

### DIFF
--- a/rapids-cmake/cpm/versions.json
+++ b/rapids-cmake/cpm/versions.json
@@ -31,9 +31,9 @@
       "git_tag" : "v${version}"
     },
     "Thrust" : {
-      "version" : "1.15.0.0",
+      "version" : "1.17.0",
       "git_url" : "https://github.com/NVIDIA/thrust.git",
-      "git_tag" : "1.15.0"
+      "git_tag" : "${version}"
     },
     "libcudacxx" : {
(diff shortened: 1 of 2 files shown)


rapids-cmake/cpm/versions.json
```diff
@@ -31,9 +31,9 @@
       "git_tag" : "v${version}"
     },
     "Thrust" : {
-      "version" : "1.15.0.0",
+      "version" : "1.16.0",
```
bdice, line 34:
```suggestion
      "version" : "1.17.0",
```

bdice (approved):
LGTM. @robertmaynard:
1. I'll make a libcudf PR as soon as possible so this can be merged at the same time.
2. I think we need to set a [custom Thrust/CUB namespace](https://github.com/rapidsai/cudf/issues/11368) for all rapids-cmake packages. That should be a separate PR, in my opinion, but wanted to bring that to your attention and get your thoughts.

bdice:
@gpucibot merge

PR #236, opened 2022-08-04
Update to CPM v0.35.4 for URL downloads match the download time

robertmaynard:
With https://github.com/cpm-cmake/CPM.cmake/pull/372 merged into CPM we should update our CPM version so that RAPIDS developers don't see CMake policy warnings with CMake 3.24+


vyasr (approved):

robertmaynard:
@gpucibot merge

PR #241, opened 2022-08-08
Docs: RAPIDS.cmake should be placed in current bin dir

### LINKED ISSUE #240: [BUG] Problems with default recipe and dependent projects using RAPIDS.cmake
**Describe the bug**
If a dependent project also uses `RAPIDS.cmake`, the default recipe will download and overwrite the file in the root build directory. This is especially problematic if the dependent project downloads an *older* version of `rapids-cmake`, pulling in an older `versions.json`.

Configure and build https://github.com/NVIDIA/cuCollections; it should work. Then re-run `cmake .` and observe that older versions of `libcudacxx` and `Thrust` are added (based on what https://github.com/NVIDIA/nvbench pulls in). The build will fail afterwards because cuCollections requires the newer versions.

**Expected behavior**
The build should be stable and work after reconfiguration.

robertmaynard:
Instead of placing in the root level binary directory place it in the current binary dir to avoid clobbering an existing RAPIDS.cmake

Fixes #240

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,8 +19,8 @@ Content](https://cmake.org/cmake/help/latest/module/FetchContent.html) into your
 cmake_minimum_required(...)
 
 file(DOWNLOAD https://raw.githubusercontent.com/rapidsai/rapids-cmake/branch-<VERSION_MAJOR>.<VERSION_MINOR>/RAPIDS.cmake
-    ${CMAKE_BINARY_DIR}/RAPIDS.cmake)
-include(${CMAKE_BINARY_DIR}/RAPIDS.cmake)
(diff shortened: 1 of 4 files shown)


bdice (approved):

bdice:
Just noticed the `DO NOT MERGE` label. This PR looks fine to me but I may be unaware of some missing piece of this.

robertmaynard:
> Just noticed the `DO NOT MERGE` label. This PR looks fine to me but I may be unaware of some missing piece of this.

Updated the examples to include guards around downloading the file if it already exists, which we also want to make sure we don't keep downloading the file.

bdice (approved):
LGTM!

robertmaynard:
@gpucibot merge

PR #242, opened 2022-08-08
Revert "Update to CPM v0.35.4 for URL downloads... (#236)"

robertmaynard:
We are seeing CI failures with dependencies of dependencies and I believe that this is causing the issue.

This reverts commit e72fd084e5db9437df3450bdfcdbb4e83597fa4e.


robertmaynard:
@gpucibot merge

PR #243, opened 2022-08-12
Ensure that we install Thrust and Cub correctly.

robertmaynard:
Since thrust ships a `thrust-config.cmake` in the source tree, a re-configure of cmake can disable the built-in thrust install
rules.

In addition thrust 1.17 has a bug in how it install cub so instead rapids-cmake needs to manage installing but packages.

### DIFF
--- a/rapids-cmake/cpm/thrust.cmake
+++ b/rapids-cmake/cpm/thrust.cmake
@@ -55,31 +55,37 @@ Result Variables
 function(rapids_cpm_thrust NAMESPACE namespaces_name)
   list(APPEND CMAKE_MESSAGE_CONTEXT "rapids.cpm.thrust")
 
-  set(options)
-  set(one_value BUILD_EXPORT_SET INSTALL_EXPORT_SET)
-  set(multi_value)
-  cmake_parse_arguments(_RAPIDS "${options}" "${one_value}" "${multi_value}" ${ARGN})
-
-  set(enable_install OFF)
-  if(_RAPIDS_INSTALL_EXPORT_SET)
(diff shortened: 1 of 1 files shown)


thomcom (approved):

robertmaynard:
@gpucibot merge

rapids-cmake/cpm/thrust.cmake
```diff
@@ -55,31 +55,37 @@ Result Variables
...
-    set(enable_install ON)
+  if(Thrust_SOURCE_DIR AND _RAPIDS_INSTALL_EXPORT_SET AND NOT exclude)
     # Make sure we install thrust into the `include/rapids` subdirectory instead of the default
     include(GNUInstallDirs)
     set(CMAKE_INSTALL_INCLUDEDIR "${CMAKE_INSTALL_INCLUDEDIR}/rapids")
-  endif()
 
-  include("${rapids-cmake-dir}/cpm/detail/package_details.cmake")
-  rapids_cpm_package_details(Thrust version repository tag shallow exclude)
+    # Thrust 1.17 has a bug where it doesn't properly install CUB, so we need to manually invoke
```
bdice, line 80: Should this path be conditional for Thrust 1.17 (or some range of affected versions)? Users can specify other Thrust versions than the default of 1.17.

robertmaynard: The existing logic will work with versions of thrust newer than 1.17. 

While we do allow overrides we can't reasonably support all previous versions ( this would break with <= 1.14 )

PR #244, opened 2022-08-15
Patch thrust and cub install rules to have proper header searches

robertmaynard:
Thrust 1.17 has a bug where it doesn't generate proper exclude rules for the `[cub|thrust]-header-search` files, which causes the build tree version to be installed instead of the install version


robertmaynard:
The `PATTERN` rule in CMake adds an implicit `$` to the end. The purpose is so you can exclude / include all `.cpp` or `.h` files. In this case we would fail to match anything so we would have 2 install rules for `[thrust|cub]-header-search.cmake` and the second one would be ignored. The problem being the second rule would generally be the correct one

bdice (approved):

dantegd (approved):

robertmaynard:
@gpucibot merge

PR #254, opened 2022-08-25
rapids_cpm_thrust record build directory location of thrust-config

robertmaynard:
Thrust doesn't place the thrust-config.cmake in the build directory,
instead it has a version in the source tree that works for
build directories.


bdice (approved):

robertmaynard:
@gpucibot merge

PR #256, opened 2022-08-29
Update rapids_find_generate_module to cmake 3.23

robertmaynard:
Missed as part of https://github.com/rapidsai/rapids-cmake/pull/227 was an update to the policy version set in `rapids_find_generate_module`


bdice (approved):

robertmaynard:
@gpucibot merge

PR #258, opened 2022-08-30
rapids_cpm_libcudacxx correct location of libcudacxx-config

robertmaynard:
Originally presumed the layout of libcudacxx matched that of thrust and cub and the file would be in `cmake/`.


bdice (approved):

robertmaynard:
@gpucibot merge

PR #265, opened 2022-09-08
Switch thrust over to use rapids-cmake patches

robertmaynard:
Leverages the infrastructure of https://github.com/rapidsai/rapids-cmake/pull/264 to switch rapids_cpm_thrust over to using the built-in patch support.

### DIFF
--- a/rapids-cmake/cpm/nvcomp.cmake
+++ b/rapids-cmake/cpm/nvcomp.cmake
@@ -93,17 +93,6 @@ function(rapids_cpm_nvcomp)
     include("${rapids-cmake-dir}/cpm/detail/get_proprietary_binary.cmake")
     rapids_cpm_get_proprietary_binary(nvcomp ${version})
 
-    # Remove incorrect public dependency on the static cuda runtime We have to modify the
-    # nvcomp-targets.cmake since these entries will cause a failure when rapids_cpm_find is called.
-    if(nvcomp_proprietary_binary AND ${version} VERSION_EQUAL "2.3.2")
-      set(target_file "${nvcomp_ROOT}/lib/cmake/nvcomp/nvcomp-targets.cmake")
-      if(EXISTS "${target_file}")
-        file(READ "${target_file}" file_contents)
-        string(REPLACE "CUDA::cudart_static" "" file_contents "${file_contents}")
-        file(WRITE "${target_file}" "${file_contents}")
-      endif()
-    endif()
-
     # Record the nvcomp_DIR so that if USE_PROPRIETARY_BINARY is disabled we can safely clear the
     # nvcomp_DIR value
     if(nvcomp_proprietary_binary)
@@ -120,6 +109,14 @@ function(rapids_cpm_nvcomp)
     endif()
   endif()
 
+  include("${rapids-cmake-dir}/cpm/detail/generate_patch_command.cmake")
+  rapids_cpm_generate_patch_command(nvcomp ${version} patch_command)
+
+  # Apply any patch commands to the proprietary binary
+  if(nvcomp_proprietary_binary AND patch_command)
+    execute_process(COMMAND ${patch_command} WORKING_DIRECTORY ${nvcomp_ROOT})
+  endif()
+
   include("${rapids-cmake-dir}/cpm/find.cmake")
   rapids_cpm_find(nvcomp ${version} ${_RAPIDS_UNPARSED_ARGUMENTS}
                   GLOBAL_TARGETS nvcomp::nvcomp
@@ -128,8 +125,12 @@ function(rapids_cpm_nvcomp)
                   GIT_TAG ${tag}
                   GIT_SHALLOW ${shallow}
                   EXCLUDE_FROM_ALL ${to_exclude}
+                  PATCH_COMMAND ${patch_command}
                   OPTIONS "BUILD_STATIC ON" "BUILD_TESTS OFF" "BUILD_BENCHMARKS OFF"
                           "BUILD_EXAMPLES OFF")
+
+  include("${rapids-cmake-dir}/cpm/detail/display_patch_status.cmake")
+  rapids_cpm_display_patch_status(nvcomp)
 
   # provice consistent targets between a found nvcomp and one building from source
   if(NOT TARGET nvcomp::nvcomp AND TARGET nvcomp)
--- a/rapids-cmake/cpm/versions.json
+++ b/rapids-cmake/cpm/versions.json
@@ -33,7 +33,14 @@
     "Thrust" : {
       "version" : "1.17.2",
       "git_url" : "https://github.com/NVIDIA/thrust.git",
-      "git_tag" : "${version}"
+      "git_tag" : "${version}",
+      "patches" : [
+        {
+          "file" : "Thrust/install_rules.diff",
+          "issue" : "Thrust 1.X installs incorrect files [https://github.com/NVIDIA/thrust/issues/1790]",
+          "fixed_in" : "2.0"
+        }
+      ]
     },
     "libcudacxx" : {
       "version" : "1.8.0",
--- a/rapids-cmake/cpm/patches/Thrust/install_rules.diff
+++ b/rapids-cmake/cpm/patches/Thrust/install_rules.diff
@@ -1,0 +1,22 @@
+diff --git a/cmake/ThrustInstallRules.cmake b/cmake/ThrustInstallRules.cmake
+index 93084c1..bf6c195 100644
+--- a/cmake/ThrustInstallRules.cmake
++++ b/cmake/ThrustInstallRules.cmake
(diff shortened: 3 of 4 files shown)


docs/packages/patches.json
```diff
@@ -0,0 +1,10 @@
+
+{
+  "patches" : [
+        {
+          "file" : "Thrust/cub_odr.diff",
+          "issue" : "cub kernel dispatch ODR [https://github.com/NVIDIA/cub/issues/545]",
+          "fixed_in" : ""
+        }
+      ]
+}
```
bdice, line 10: Format this to something like:
```suggestion
{
  "patches" : [
    {
      "file" : "Thrust/cub_odr.diff",
      "issue" : "cub kernel dispatch ODR [https://github.com/NVIDIA/cub/issues/545]",
      "fixed_in" : ""
    }
  ]
}
```

rapids-cmake/cpm/detail/display_patch_status.cmake
```diff
@@ -0,0 +1,40 @@
+#=============================================================================
+# Copyright (c) 2021, NVIDIA CORPORATION.
```
bdice, line 2: Looks like a new file.
```suggestion
# Copyright (c) 2022, NVIDIA CORPORATION.
```

rapids-cmake/cpm/detail/generate_patch_command.cmake
```diff
@@ -0,0 +1,102 @@
+#=============================================================================
+# Copyright (c) 2021, NVIDIA CORPORATION.
```
bdice, line 2:
```suggestion
# Copyright (c) 2022, NVIDIA CORPORATION.
```

rapids-cmake/cpm/detail/generate_patch_command.cmake
```diff
@@ -0,0 +1,102 @@
...
+         patches)
+  if(no_default_patch AND no_override_patch)
+    return() # no patches
+  endif()
+  if(NOT no_override_patch)
+    set(json_data "${override_json_data}")
+  endif()
+
+  # Parse required fields
+  function(rapids_cpm_json_get_value jdata name)
```
bdice, line 54: Does `jdata` stand for `json_data`? Would a more explicit name like `json_data` shadow the `json_data` in the enclosing scope? Not sure how scopes work in CMake. Maybe `json_data_` or something like that.

bdice (commented):
I have only minor comments. Overall this looks good.

rapids-cmake/cpm/detail/generate_patch_command.cmake
robertmaynard: Yeah it would cause shadowing, and cmake-lint will get angry :)

I can do `json_data_`

vyasr (changes requested):
See https://github.com/rapidsai/cudf/pull/11665#pullrequestreview-1104931774: is this urgent, or can we wait for upstream patches for thrust?

I assume that we'll need the nvcomp patch either way.

robertmaynard:
> See [rapidsai/cudf#11665 (review)](https://github.com/rapidsai/cudf/pull/11665#pullrequestreview-1104931774): is this urgent, or can we wait for upstream patches for thrust?
> 
> I assume that we'll need the nvcomp patch either way.

With Thrust 1.17.2 now released I will make the cub patch conditional. The nvcomp and thrust intall rule patches are still needed.

robertmaynard:
We could also merge https://github.com/rapidsai/rapids-cmake/pull/264 first as it is just the infrastructure bits of this PR

rapids-cmake/cpm/versions.json
```diff
@@ -18,7 +18,14 @@
       "git_tag" : "v2.2.0",
       "proprietary_binary" : {
         "x86_64-linux" : "https://developer.download.nvidia.com/compute/nvcomp/${version}/local_installers/nvcomp_install_CUDA_11.x.tgz"
-      }
+      },
+      "patches" : [
+        {
+          "file" : "nvcomp/bad_cudart_dep.diff",
+          "issue" : "leaking cudart static dep to consumers",
+          "fixed_in" : "2.3.3"
```
bdice, line 26: I think the previous patch condition was `==2.3.2` but this will patch `<2.3.3`. Is that desired? Will the patch apply to earlier versions successfully? Is that support important for users (if any) relying on the last open-source version of 2.2?

bdice (approved):
Looks good to me. I have one question.

rapids-cmake/cpm/versions.json
robertmaynard: Good questions. The patch version check is computed based on the `version` entry value. So for nvcomp the `patches` entries won't be applied to the open source fallback since the current `version` value matches the `fixed_in` value.

If we rolled back the `version` value ( say to 2.3.1 ) we will have the patch step execute and fail, but the entire CMake configuration will be considered valid:

```cmake
-- CPM: adding package nvcomp@2.3.1 (v2.2.0)
Release build.
-- Finding CUB
Build nvCOMP version 2.2.0
-- rapids-cmake [nvcomp]: failed to apply diff bad_cudart_dep.diffrapids-cmake [nvcomp]: git diff output: error: lib/cmake/nvcomp/nvcomp-targets.cmake: No such file or directory
-- Configuring done
-- Generating done
-- Build files have been written to: /home/rmaynard/Work/temp/build
```

bdice: If we do not need to support 2.3.1 then this is fine.

robertmaynard: I think we should "live at head" and drop the nvcomp and thrust 1.17 patches. Users that are at head won't need these patches, and people that use an override for older versions can explicit specify the patches themselves.

Thoughts?

bdice: I'm fine with that!

bdice:
> This needs to be merged after #273 so we have a version of CPM that won't warn about dirty source trees after we patch them.

I approved and merged #273, it seemed fine to merge that directly to unblock this PR.

rapids-cmake/cpm/versions.json
robertmaynard: rewrote history and dropped the unneeded patches.

rapids-cmake/cpm/nvcomp.cmake
```diff
@@ -120,6 +109,14 @@ function(rapids_cpm_nvcomp)
     endif()
   endif()
 
+  include("${rapids-cmake-dir}/cpm/detail/generate_patch_command.cmake")
+  rapids_cpm_generate_patch_command(nvcomp ${version} patch_command)
+
+  # Apply any patch commands to the proprietary binary
+  if(nvcomp_proprietary_binary AND patch_command)
+    execute_process(COMMAND ${patch_command} WORKING_DIRECTORY ${nvcomp_ROOT})
```
vyasr, line 117: I think I'm missing something. Where is the nvcomp patch coming from?

vyasr (commented):
Apologies, I let this review slip. I am fine with pushing this if you want, but I can also iterate to get it done in the next couple of days before code freeze if you prefer. It looks to be basically done at this point.

rapids-cmake/cpm/nvcomp.cmake
robertmaynard: Line 113 has the call to `rapids_cpm_generate_patch_command` which will populate the `patch_command` variable

robertmaynard:
> Apologies, I let this review slip. I am fine with pushing this if you want, but I can also iterate to get it done in the next couple of days before code freeze if you prefer. It looks to be basically done at this point.

Since it isn't critical, lets push to 22.12

rapids-cmake/cpm/nvcomp.cmake
vyasr: Right, but I don't actually see any patches for nvcomp in the versions.json file. Are we planning to put that in per-package override.json files?

robertmaynard: Ohhh.

So originally this PR moved the inline logic for patching `2.3.2` to a patch file. While this PR was open we moved 22.10 to nvcomp 2.4 making the patch for 2.3.2 not necessary, and therefore it was dropped from this PR.

So that is why in this PR we see the inline logic for 2.3.2 being removed and no matching patch being added.

vyasr: Got it, so should we remove the nvcomp patching logic for now then, or do you want to keep the framework in place in anticipation of needing future patches?

robertmaynard: I would prefer to keep the framework in place. That way we can use patches in override json files without any modifications to rapids-cmake itself.

vyasr (approved):
LGTM. We can merge early in 22.12 then.

robertmaynard:
@gpucibot merge

PR #271, opened 2022-09-19
Correct broken patch_toolkit API docs, and CMake API cross references

robertmaynard:


rapids-cmake/cmake/build_type.cmake
```diff
@@ -21,13 +21,13 @@ rapids_cmake_build_type
 
 .. versionadded:: v21.06.00
 
-Establish the :cmake:variable:`CMAKE_BUILD_TYPE` default value.
+Establish the :cmake:variable:`CMAKE_BUILD_TYPE <cmake:variable:CMAKE_BUILD_TYPE>` default value.
```
vyasr, line 24: Can you remind me why the double reference is necessary again? I'm wondering if there is some way to update conf.py with some extra role-related information to help Intersphinx link these correctly.

rapids-cmake/cuda/init_runtime.cmake
```diff
@@ -28,8 +28,8 @@ Establish what CUDA runtime library should be propagated
     rapids_cuda_init_runtime( USE_STATIC (TRUE|FALSE) )
 
   Establishes what CUDA runtime will be used, if not already explicitly
-  specified, via the :cmake:variable:`CMAKE_CUDA_RUNTIME_LIBRARY` variable.
-  We also set :cmake:variable:`CUDA_USE_STATIC_CUDA_RUNTIME` to control
+  specified, via the :cmake:variable:`CMAKE_CUDA_RUNTIME_LIBRARY <cmake:variable:CMAKE_CUDA_RUNTIME_LIBRARY>`
+  variable. We also set :cmake:variable:`CUDA_USE_STATIC_CUDA_RUNTIME` to control
```
vyasr, line 32: Do you need to do the same thing for `CUDA_USE_STATIC_CUDA_RUNTIME`?

vyasr (approved):

rapids-cmake/cmake/build_type.cmake
bdice: This may not be a helpful comment, but I looked into this at one point and I _think_ it had to do with a limitation in CMake's rst object inventory or some other thing we cannot control. I concluded it wasn't worth fighting and that we should just accept it, but it's possible that I overlooked some possible solutions.

robertmaynard: What @bdice says is my understanding as well. The CMake rst setup wasn't designed for external usage and we have to jump through some hoops to cross reference it.

vyasr: Got it, that's fine then.

robertmaynard:
@gpucibot merge

PR #275, opened 2022-09-21
Create conda package of patched dependencies

robertmaynard:
Instead of having the rapids version of thrust , cub, and libcudacxx being part of the `librmm` package provide them as a separate package.


conda/recipes/rapids_cuda_patched_dependencies/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+  number: {{ GIT_DESCRIBE_NUMBER }}
+  string: cuda{{ cuda_major }}_py{{ py_version }}_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
+  script_env:
+    - VERSION_SUFFIX
+  ignore_run_exports_from:
+    - {{ compiler('cuda') }}
+
+requirements:
+  build:
+    - cmake>=3.20.1,!=3.23.0
```
bdice, line 25: The minimum CMake version specified above is 3.23.1.
```suggestion
    - cmake>=3.23.1
```

conda/recipes/rapids_cuda_patched_dependencies/build.sh
```diff
@@ -0,0 +1,24 @@
+# Copyright (c) 2022, NVIDIA CORPORATION.
+
+# Script assumes the script is executed from the root of the repo directory
+
+REPODIR=$(cd $(dirname $0); pwd)
```
bdice, line 5: For consistency with other names that use an `_`: 
```suggestion
REPO_DIR=$(cd $(dirname $0); pwd)
```

conda/recipes/rapids_cuda_patched_dependencies/build.sh
```diff
@@ -0,0 +1,24 @@
...
+for bd in ${BUILD_DIRS}; do
+  if [ -d "${bd}" ]; then
+      find "${bd}" -mindepth 1 -delete
+      rmdir "${bd}" || true
+  fi
+done
+
+cmake -S "${SOURCE_DIR}" -B "${BUILD_DIR}" \
+    -DCMAKE_INSTALL_LIBDIR="lib" \
+    -DCMAKE_INSTALL_PREFIX="${INSTALL_PREFIX}" \
```
bdice, line 22: I might be missing something. We need a "special" prefix to prevent these patched dependencies from being found directly in `include/`, right? Where is that defined?

robertmaynard: That logic is defined in the `rapids_cpm_thrust` and `rapids_cpm_libcudacxx` where we modify the `CMAKE_INSTALL_INCLUDEDIR` ( install into a nested `rapids` dir ).

conda/recipes/rapids_cuda_patched_dependencies/meta.yaml
```diff
@@ -0,0 +1,41 @@
+# Copyright (c) 2022, NVIDIA CORPORATION.
+
+{% set version = environ.get('GIT_DESCRIBE_TAG', '0.0.0.dev').lstrip('v') + environ.get('VERSION_SUFFIX', '') %}
+{% set py_version=environ.get('CONDA_PY', 39) %}
+{% set cuda_version = '.'.join(environ.get('CUDA', '11.5').split('.')[:2]) %}
+{% set cuda_major=cuda_version.split('.')[0] %}
+
+package:
+  name: rapids_cuda_patched_dependencies
```
robertmaynard, line 9: What do we think of this name?

@bdice pruposed `rapids_cuda_includes` via slack which I like more

bdice: `rapids` -- It's our package.
`cuda` -- It's CUDA-related libraries but not CUDA itself. We could theoretically have includes that are not CUDA, too. Maybe we drop this word.
`includes` -- We should emphasize that this is a header-only package.

So perhaps `rapids_includes` or `rapids_core_dependencies` or `rapids_common_includes` or a variation on those.

robertmaynard: I like `rapids_core_dependencies`, I will go with that for now

robertmaynard:
@bdice I want to resume working on this for 22.12. Do you like the state of this PR?

My thinking is we would merge this, followed by https://github.com/rapidsai/rapids-cmake/pull/283 and than open a final PR for using the new github actions that will upload the conda package.

ajschmidt8 (approved):
Deferring to you guys for the final review / merge, but looks good from an Ops perspective.

conda/recipes/rapids_core_dependencies/build.sh
```diff
@@ -0,0 +1,24 @@
...
+SOURCE_DIR=${REPO_DIR}/conda/recipes/rapids_core_dependencies
+BUILD_DIR=${BUILD_DIR:=${REPO_DIR}/build}
+
+INSTALL_PREFIX=${INSTALL_PREFIX:=${PREFIX:=${CONDA_PREFIX:=$BUILD_DIR/install}}}
+export PARALLEL_LEVEL=${PARALLEL_LEVEL:-4}
+
+for bd in ${BUILD_DIRS}; do
+  if [ -d "${bd}" ]; then
+      find "${bd}" -mindepth 1 -delete
+      rmdir "${bd}" || true
```
bdice, line 16: How does this compare to `rm -rf`?

bdice (approved):
@robertmaynard This looks great to me. I also support your plan to merge this and then work through GitHub Actions / deployment.

conda/recipes/rapids_core_dependencies/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+package:
+  name: rapids_core_dependencies
+  version: {{ version }}
+
+source:
+  git_url: ../../..
+
+build:
+  number: {{ GIT_DESCRIBE_NUMBER }}
+  string: cuda{{ cuda_major }}_py{{ py_version }}_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
```
bdice, line 17: Actually I'm not sure if this makes sense to include the Python version in the build string. This package doesn't rely on Python at all.
```suggestion
  string: cuda{{ cuda_major }}_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
```

bdice (changes requested):
No Python dependency. See comment.

conda/recipes/rapids_core_dependencies/meta.yaml
bdice: These example packages don't rely on Python, and thus don't include Python versions in their build strings.
- https://anaconda.org/conda-forge/libgcc-ng/files
- https://anaconda.org/rapidsai-nightly/librmm/files

conda/recipes/rapids_core_dependencies/meta.yaml
```diff
@@ -0,0 +1,41 @@
+# Copyright (c) 2022, NVIDIA CORPORATION.
+
+{% set version = environ.get('GIT_DESCRIBE_TAG', '0.0.0.dev').lstrip('v') + environ.get('VERSION_SUFFIX', '') %}
+{% set py_version=environ.get('CONDA_PY', 39) %}
```
bdice, line 4: Not needed, see below.
```suggestion
```

conda/recipes/rapids_core_dependencies/meta.yaml
```diff
@@ -0,0 +1,41 @@
+# Copyright (c) 2022, NVIDIA CORPORATION.
+
+{% set version = environ.get('GIT_DESCRIBE_TAG', '0.0.0.dev').lstrip('v') + environ.get('VERSION_SUFFIX', '') %}
```
bdice, line 3: To verify my assumptions:
- We only release updates to the core dependencies on-cycle with the rest of RAPIDS releases (or hotfix)
- Updates to the dependencies come from `versions.json` and thus are part of the repo's commit hashes (new dependency versions = new commit hash = new package)

bdice: Where does `VERSION_SUFFIX` come from? Our CI? It's not supplied by conda-build afaict.

conda/recipes/rapids_core_dependencies/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+package:
+  name: rapids_core_dependencies
+  version: {{ version }}
+
+source:
+  git_url: ../../..
+
+build:
+  number: {{ GIT_DESCRIBE_NUMBER }}
```
bdice, line 16: Did you copy this from somewhere? Typically build numbers are incremented manually, in my experience. Also `{{ this_syntax }}` inserts a jinja template variable (like those set by `{% set ... %}` above), while [`GIT_DESCRIBE_NUMBER` is an environment variable](https://docs.conda.io/projects/conda-build/en/latest/user-guide/environment-variables.html#git-environment-variables).

conda/recipes/rapids_core_dependencies/build.sh
robertmaynard: I lifted this from rmm / cudf and how the `build.sh --clean` operation works ( https://github.com/rapidsai/rmm/blob/branch-22.12/build.sh#L150 ).

I will update the script to document the why of this over `rm -rf`

conda/recipes/rapids_core_dependencies/meta.yaml
robertmaynard: I lifted this from `librmm` ( https://github.com/rapidsai/rmm/blob/branch-22.12/conda/recipes/librmm/meta.yaml#L44 ) as this is the first conda package I have written.

conda/recipes/rapids_core_dependencies/meta.yaml
robertmaynard: Those assumptions are correct ( excluding nightlies, but the git hash ensures uniqueness ).

The `VERSION_SUFFIX` should come from our CI ( it is listed in the `script_env` section). I based this on the RMM setup for Github Actions.  Looking at librmm nightlies I see them with the expected `VERSION_SUFFIX` of yymmdd`

conda/recipes/rapids_core_dependencies/meta.yaml
bdice: Alright. Let's leave it as-is, because it seems to work (a bit contrary to my expectation).

bdice (approved):
Looks good enough to me. We'll iron out any remaining issues once nightlies are deployed.

robertmaynard:
@gpucibot merge

PR #276, opened 2022-09-27
Restore rapids_cython_create_modules output variable name

robertmaynard:
The changes in #203 are meant to guard rapids-cmake from acting or behaving differently due to an existing variable prefixed with RAPIDS which causes un-intended code paths to be executed.

In the case of `rapids_cython_create_modules` it is perfectly reasonable to have an output variable with the `RAPIDS` prefix. rapids-cmake won't execute differently due this variable being set before function execution.


vyasr (approved):
Thanks!

vyasr:
@gpucibot merge

PR #283, opened 2022-10-04
Add GitHub action workflow to rapids-cmake

robertmaynard:
Enable github actions for rapids-cmake

Requires #275


ci/build_cpp.sh
```diff
@@ -0,0 +1,11 @@
...
+
+set -euo pipefail
+
+source rapids-env-update
+
+rapids-print-env
+
+rapids-logger "Begin cpp build"
+
+rapids-mamba-retry mambabuild conda/recipes/rapids_core_dependencies
```
ajschmidt8, line 11: it looks like the conda packages get built here, but not uploaded to `s3` like `rmm` (see below). Is this intended? Ideally they should be uploaded so that they can be used in the subsequent testing stages without having to re-build anything.


- https://github.com/rapidsai/rmm/blob/dce82cbac75118779c2f09cf2490480d0c3bd4a0/ci/build_cpp.sh#L11-L13

ci/test_cpp.sh
```diff
@@ -0,0 +1,31 @@
...
+
+set -euo pipefail
+
+rapids-logger "Create test conda environment"
+. /opt/conda/etc/profile.d/conda.sh
+
+rapids-dependency-file-generator \
+  --generate conda \
+  --file_key test \
+  --matrix "cuda=${RAPIDS_CUDA_VERSION%.*};arch=$(arch);py=${RAPIDS_PY_VERSION}" > env.yaml
```
ajschmidt8, line 13:
```suggestion
rapids-dependency-file-generator \
  --generate conda \
  --file_key test \
  --matrix "cuda=${RAPIDS_CUDA_VERSION%.*};arch=$(arch)" > env.yaml
```

We updated this in the `rmm` implementation. Since the `cpp` tests don't depend on any particular Python version, we can remove the value here.

ci/test_cpp.sh
```diff
@@ -0,0 +1,31 @@
...
+
+rapids-mamba-retry env create --force -f env.yaml -n test
+set +eu
+conda activate test
+set -u
```
ajschmidt8, line 18: I'm curious about why these `set` values are used. Can you elaborate?

ci/test_cpp.sh
```diff
@@ -0,0 +1,31 @@
...
+
+
+rapids-logger "Check GPU usage"
+nvidia-smi
+
+rapids-logger "Begin cpp tests"
+cmake -S testing -B build
```
ajschmidt8, line 25: is it necessary to build stuff in the `test` stage? Ideally this script should reuse things from the `build` stage

ajschmidt8 (commented):
left a few questions about some things.

We've updated the `rmm` workflows quite a bit since you first started working on this PR. I'm happy to push some updates to this PR to get things in-sync. Just wanted to get clarity on some of the questions here first.

ci/test_cpp.sh
robertmaynard: Correct this is very quick and isn't building anything.

ci/build_cpp.sh
robertmaynard: This is intended. We haven't merged the conda package logic ( https://github.com/rapidsai/rapids-cmake/pull/275 ) yet so I don't want to upload till we have agreed on everything.

ci/test_cpp.sh
robertmaynard: We got unbound errors when running the activate/decativate scripts so I turned off error checking for the conda scripts:

https://github.com/rapidsai/rapids-cmake/actions/runs/3192203924/jobs/5209416886#step:6:392

ci/build_cpp.sh
robertmaynard: If #275 is merged before this PR I will need to update this to do the uploads.

robertmaynard:
@ajschmidt8 Are you okau if I remove the old CI scripts in this PR as well?

ajschmidt8:
> @ajschmidt8 Are you okay if I remove the old CI scripts in this PR as well?

Yes, go for it. I am going to push some changes to this PR. please don't merge until I've completed those changes.

ajschmidt8:
Need to get the small PR below merged and deployed to fix one of these GH Actions errors

- https://github.com/rapidsai/gha-tools/pull/24

ci/test_cpp.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+#!/bin/bash
```
ajschmidt8, line 3:
```suggestion
#!/bin/bash
```

Extra shebang can be removed.

ajschmidt8 (approved):
one suggestions, otherwise LGTM

robertmaynard:
@gpucibot merge

robertmaynard:
rerun tests

PR #286, opened 2022-10-06
Update to nvcomp 2.4.1 to fix zstd decompression

### LINKED ISSUE #282: [BUG] Update to nvcomp-2.4.1
A data corruption issue during zstd decompression on some pascal architectures has been identified for nvcomp-2.4, as described in https://github.com/NVIDIA/spark-rapids/issues/6614.
The DevTech team is working on a fix, which is under test now.  The hope is to release nvcomp-2.4.1 very soon.
If that happens, we will need to update rapids cmake as well, to pull in nvcomp-2.4.1 instead of nvcomp-2.4 for 22.10.

Another related change is: https://github.com/rapidsai/cudf/pull/11856
This change will cause zstd decompression to fail if the architecture is 6.1.  This change will be updated to do so only for nvcomp-2.4 if nvcomp-2.4.1 does indeed fix the problem.

robertmaynard:
Fixes #282

### DIFF
--- a/rapids-cmake/cpm/versions.json
+++ b/rapids-cmake/cpm/versions.json
@@ -13,12 +13,12 @@
       "git_tag" : "12d13bdc5e74801645eba3e46a64081b9b020dca"
     },
     "nvcomp" : {
-      "version" : "2.4",
+      "version" : "2.4.1",
       "git_url" : "https://github.com/NVIDIA/nvcomp.git",
       "git_tag" : "v2.2.0",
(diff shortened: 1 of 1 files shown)


vyasr (approved):
Do the URLs change somewhat arbitrarily from version to version?

robertmaynard:
> Do the URLs change somewhat arbitrarily from version to version?

There has been churn with 2.3 and 2.4. But the names going forward should stay consistent

PR #292, opened 2022-10-14
Patch results are only displayed once per invocation of CMake

robertmaynard:
## Description
Instead of showing the patch status for each call to `rapids_cpm_<pkg>` only display it on the first. This makes it clearer that we aren't applying the patches multiple times.

## Checklist


vyasr (approved):

robertmaynard:
@gpucibot merge

PR #293, opened 2022-10-14
Add support for cloning Google benchmark

### LINKED ISSUE #288: [FEA] rapids-cmake should support gbench
We currently have a number of functions specialized for using CPM to download packages that are used across RAPIDS along with a well-defined dependency management setup. However, Google benchmark is currently managed independently in each repo that uses it.

We should implement `rapids_cpm_gbench` for this purpose.

None

**Additional context**
cudf is moving away from Google benchmark towards nvbench, and we are not interested in using this as an opportunity to upgrade our versions. However, we should pin rapids-cmake to the latest version used across RAPIDS to continue support the needs of different packages, even if that is slightly higher than what cudf currently uses (specifically, cuml is currently on 1.5.3 while cudf is on 1.5.2).

vyasr:
## Description

Many RAPIDS libraries use Google benchmark to manage C++ benchmarks. This PR centralizes the dependency and the version management.

Resolves #288.

## Checklist


vyasr:
I should add that cuML and cuGraph's current versions of this function [specify the lib dir for gbench](https://github.com/rapidsai/cuml/blob/branch-22.12/cpp/cmake/thirdparty/get_gbench.cmake#L33), but I don't think we need or want that. Please correct me if I'm wrong.

rapids-cmake/cpm/gbench.cmake
```diff
@@ -0,0 +1,75 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#=============================================================================
+include_guard(GLOBAL)
+
+#[=======================================================================[.rst:
+rapids_cpm_gbench
+-----------------
+
+.. versionadded:: v21.12.00
```
robertmaynard, line 22:
```suggestion
.. versionadded:: v22.12.00
```

rapids-cmake/cpm/gbench.cmake
```diff
@@ -0,0 +1,75 @@
...
+  rapids_cpm_gbench( [BUILD_EXPORT_SET <export-name>]
+                     [INSTALL_EXPORT_SET <export-name>]
+                     [<CPM_ARGS> ...])
+
+.. |PKG_NAME| replace:: GBench
+.. include:: common_package_args.txt
+
+Result Targets
+^^^^^^^^^^^^^^
+  benchmark::benchmark targets will be created
```
robertmaynard, line 41: Looks like `gbench` also has a benchmark::benchmark_main target so lets expose that as well.

robertmaynard (changes requested):
Looks great. Just a couple of small changes needed

robertmaynard:
> I should add that cuML and cuGraph's current versions of this function [specify the lib dir for gbench](https://github.com/rapidsai/cuml/blob/branch-22.12/cpp/cmake/thirdparty/get_gbench.cmake#L33), but I don't think we need or want that. Please correct me if I'm wrong.

We will want to specify the install directory if `INSTALL_EXPORT_SET` is set. But we will want to use `rapids_cmake_install_lib_dir` to compute the correct location. I expect this was done so it doesn't go into something like `lib64` and therefore breaks conda layout requirements.

bdice (approved):
No comments aside from what @robertmaynard mentioned. Nice, thanks for doing this!

vyasr:
> > I should add that cuML and cuGraph's current versions of this function [specify the lib dir for gbench](https://github.com/rapidsai/cuml/blob/branch-22.12/cpp/cmake/thirdparty/get_gbench.cmake#L33), but I don't think we need or want that. Please correct me if I'm wrong.
> 
> We will want to specify the install directory if `INSTALL_EXPORT_SET` is set. But we will want to use `rapids_cmake_install_lib_dir` to compute the correct location. I expect this was done so it doesn't go into something like `lib64` and therefore breaks conda layout requirements.

I'm now setting the `CMAKE_INSTALL_LIBDIR` for this target unconditionally (using `rapids_cmake_install_lib_dir`). I'm assuming that the `BENCHMARK_ENABLE_INSTALL` option for GBench will take care of whether installation happens at all so it should be safe to always set this.

vyasr:
@robertmaynard are you happy with the current state of this PR?

robertmaynard (approved):
LGTM, thank you

vyasr:
@gpucibot merge

PR #294, opened 2022-10-25
Add missing CPM_ARGS to gbench

vyasr:
## Description

The `rapids_cpm_find` command for Google benchmark is missing a `CPM_ARGS` argument, which is preventing the subsequent arguments from being parsed correctly.

## Checklist


vyasr:
@robertmaynard I am not sure why the tests I wrote did not catch this bug. I am seeing them pass locally either way with the clone successful. It only fails when I actually try to leverage this inside another project. Any ideas why that might be?

robertmaynard (approved):

vyasr:
Ah I think that would be why, I was using an old test directory that seems to have been configured with that off. Reconfiguring a new build directory allows me to repro locally as well.

vyasr:
@gpucibot merge

PR #295, opened 2022-10-25
Add ability to specify library directories for target rpaths

vyasr:
## Description

Python extension modules built using scikit-build and rapids-cmake are typically designed around a model where a C++ library is built and the resulting Cython extension modules must link to it. While it is in principle possible to use both scikit-build and rapids-cmake with only Cython, that is not the case for any of the major RAPIDS packages. When the C++ component is built via scikit-build and installed into the final package, it must also be discoverable by the extension modules. Currently this is accomplished by propagating a variable `RAPIDS_CYTHON_CREATED_MODULES` from `rapids_cython_create_modules` that calling functions can use to get the list of targets whose RPATHs need updating, and setting these RPATHs is the responsibility of the caller. However, this approach is error-prone and frequently results in incorrect RPATH specifications.

This PR automates the process by having the calling code provide a set of directories where compiled libraries may be placed. The RPATHs of all extension modules built via `rapids_cython_create_modules` are then automatically updated to include those library directories.

## Checklist


rapids-cmake/cython/create_modules.cmake
```diff
@@ -117,8 +117,15 @@ function(rapids_cython_create_modules)
     endif()
     install(TARGETS ${cython_module} DESTINATION ${_RAPIDS_CYTHON_INSTALL_DIR})
 
-    # Default the INSTALL_RPATH for all modules to $ORIGIN.
-    set_target_properties(${cython_module} PROPERTIES INSTALL_RPATH "\$ORIGIN")
+    # Default the INSTALL_RPATH for all modules to $ORIGIN. We also append any lib dirs specified in
```
robertmaynard, line 120: We should add a notes to the docs of this fuction that cover what install rpaths will be set.

rapids-cmake/cython/set_lib_dirs.cmake
```diff
@@ -0,0 +1,53 @@
...
+
+include_guard(GLOBAL)
+
+#[=======================================================================[.rst:
+rapids_cython_set_lib_dirs
+--------------------------
+
+.. versionadded:: v22.12.00
+
+Set the directories where any dependent libraries will be placed.
```
robertmaynard, line 23: The short description should outline we will populate the ``RAPIDS_CYTHON_<ProjectName>_LIB_DIRS` variable

rapids-cmake/cython/set_lib_dirs.cmake
```diff
@@ -0,0 +1,53 @@
...
+
+.. code-block:: cmake
+
+  rapids_cython_set_lib_dirs(<dir1> <dir2> ...)
+
+The relative path to the provided directories is appended to the rpath of any
```
robertmaynard, line 29: This comment doesn't seem to match what occurs.

This function looks to convert a set of paths if either abs or relative to all absolute.

robertmaynard (changes requested):
Here are some minor changes.

I think this is a good starting place but we should change the separation of concerns.

I think we should have `create_modules.cmake` just add the `ORIGIN` rpath entry, and have a secondary function that you call providing a target, and a set of directories and it will add those as INSTALL_RPATHS.

So instead of having the logic look like:
```
set_lib_dirs( ..... ) # propagate stuff via side channels to create_modules
create_modules( )
```

We have
```
create_modules()
set_rpath_entries(<cython_target> <dirs>)
```

rapids-cmake/cython/add_rpath_entries.cmake
```diff
@@ -0,0 +1,90 @@
...
+    # Compute the path relative to the current target.
+    set(target_paths)
+    get_target_property(target_source_dir ${target} SOURCE_DIR)
+    foreach(target_path IN LISTS cleaned_paths)
+      cmake_path(RELATIVE_PATH target_path BASE_DIRECTORY "${target_source_dir}")
+      list(APPEND target_paths "$ORIGIN/${target_path}")
+    endforeach()
+    list(JOIN target_paths ";" target_paths)
+
+    get_target_property(current_rpath ${target} INSTALL_RPATH)
```
robertmaynard, line 87: You can simplify this by using `set_property(TARGET` with the APPEND mode. That will remove the need to fetch any existing INSTALL_RPATH entries.

robertmaynard (changes requested):
Overall looks great, have a couple of minor suggestions.

rapids-cmake/cython/add_rpath_entries.cmake
vyasr: In general, is there a reason to prefer `set_target_properties` vs `set_property`? I know to always prefer target-specific operations rather than setting global properties, but with the `set_property(TARGET)` syntax it seems like either would be fine. The former supports multiple properties, while the latter supports some other things like `APPEND`, so I guess it varies on a case by case basis. Am I missing anything there?

rapids-cmake/cython/add_rpath_entries.cmake
```diff
@@ -0,0 +1,90 @@
...
+  endif()
+
+  # Transform all paths to paths relative to the current directory.
+  set(cleaned_paths)
+  foreach(path IN LISTS _RAPIDS_CYTHON_PATHS)
+    if(NOT IS_ABSOLUTE path)
+      # Convert relative paths to paths from the provided root, then make them relative to the
+      # current directory.
+      cmake_path(ABSOLUTE_PATH path BASE_DIRECTORY "${_RAPIDS_CYTHON_ROOT_DIRECTORY}")
+      cmake_path(ABSOLUTE_PATH path)
```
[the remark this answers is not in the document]
vyasr: I suspect that I am doing something wrong with the previous call. Currently this call is necessary to handle the case when a root directory is passed. For some reason I get a relative path back if `BASE_DIRECTORY != PROJECT_SOURCE_DIR`. If I add a `message("The path is ${path}")` before and after each of these `cmake_path` calls, here is what I currently get when [this test](https://github.com/rapidsai/rapids-cmake/pull/295/files#diff-ce676ffc31fdefa61b46bbf0206a8676379dd1a2faad4efd3472bc9be05babd7R48) runs:
```
538: The path is ../libraries
538: The path is ../../libraries
538: The path is /nvme/0/vyasr/rapids/rapids-cmake/testing/cython/add_rpath_entries/../../libraries
```

If I remove the second `cmake(ABSOLUTE_PATH)`, I instead get:
```
538: The path is ../libraries
538: The path is ../../libraries
538: The path is ../../libraries
```
Evidently the first call is in fact _not_ producing an absolute path in this scenario. Note that [the first test (where I don't provide a `BASE_DIRECTORY`)](https://github.com/rapidsai/rapids-cmake/pull/295/files#diff-ce676ffc31fdefa61b46bbf0206a8676379dd1a2faad4efd3472bc9be05babd7R41) does not exhibit this problem. Am I using `cmake_path` incorrectly in some way here?

rapids-cmake/cython/add_rpath_entries.cmake
robertmaynard: `set_target_properties` has a subset of features compared to `set_property`. Personally I only use `set_property` for `APPEND` as other developers are less framilar with the syntax.

rapids-cmake/cython/add_rpath_entries.cmake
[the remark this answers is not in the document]
robertmaynard: The issue is that `_RAPIDS_CYTHON_ROOT_DIRECTORY` isn't an absolute directory path. So when you call `cmake_path` the first time it can't make it a true absolute path as it has no idea on what directory you want to evaluate the relative `BASE_DIRECTORY` in regards too. The second one gets you a proper absolute path as now we evaluate based on `CMAKE_CURRENT_SOURCE_DIR`

robertmaynard (approved):

vyasr:
@gpucibot merge

PR #301, opened 2022-11-09
Add `ninja` as build dependency

ajschmidt8:
## Description

This PR adds `ninja` and `make` as build dependencies of `rapids-cmake`. Additionally, it disables the `sccache` S3 backend for the c++ tests since compile times are negligible.

## Checklist

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -20,6 +20,7 @@ dependencies:
       - output_types: [conda, requirements]
         packages:
           - cmake>=3.23.1
+          - ninja
       - output_types: conda
         packages:
           - c-compiler
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -20,12 +20,12 @@ dependencies:
       - output_types: [conda, requirements]
         packages:
           - cmake>=3.23.1
-          - make
           - ninja
       - output_types: conda
         packages:
           - c-compiler
           - cxx-compiler
+          - make
   cudatoolkit:
     specific:
       - output_types: conda
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -20,6 +20,7 @@ dependencies:
       - output_types: [conda, requirements]
         packages:
           - cmake>=3.23.1
+          - make
           - ninja
       - output_types: conda
         packages:
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -20,7 +20,6 @@ dependencies:
       - output_types: [conda, requirements]
         packages:
           - cmake>=3.23.1
-          - ninja
       - output_types: conda
         packages:
           - c-compiler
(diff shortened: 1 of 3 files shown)


robertmaynard (approved):

jakirkham:
Did we want to add `make` here too (based on offline discussion)?

robertmaynard:
> Did we want to add `make` here too (based on offline discussion)?

Good call, yes we do.

dependencies.yaml
```diff
@@ -20,6 +20,7 @@ dependencies:
       - output_types: [conda, requirements]
         packages:
           - cmake>=3.23.1
+          - ninja
```
robertmaynard, line 23:
```suggestion
          - ninja
          - make
```

robertmaynard (changes requested):

dependencies.yaml
ajschmidt8: added in https://github.com/rapidsai/rapids-cmake/actions/runs/3431545485/jobs/5719970033

raydouglass (approved):

jakirkham (approved):

ajschmidt8:
hmmmm. something about the changes in this PR is making the test script take exponentially longer than it has in the past.

ajschmidt8:
I canceled the workflow for now to prevent wasting GPU resources since they're limited in GH Actions. we'll have to investigate why these changes are making things slower.

jakirkham:
Think `make` needs to be only Conda

dependencies.yaml
```diff
@@ -20,6 +20,8 @@ dependencies:
       - output_types: [conda, requirements]
         packages:
           - cmake>=3.23.1
+          - make
+          - ninja
       - output_types: conda
         packages:
```
jakirkham, line 26:
```suggestion
        packages:
          - make
```

dependencies.yaml
```diff
@@ -20,6 +20,8 @@ dependencies:
       - output_types: [conda, requirements]
         packages:
           - cmake>=3.23.1
+          - make
```
jakirkham, line 23:
```suggestion
```

ajschmidt8:
> hmmmm. something about the changes in this PR is making the test script take exponentially longer than it has in the past.
> 
> here's an old run for comparison:
> 
> * https://github.com/rapidsai/rapids-cmake/actions/runs/3411830012/jobs/5676747432

As it turns out, the slowdown was caused by the CI image changes below. Adding `ENV SCCACHE_BUCKET=rapids-sccache` to the CI images enables the S3 backend for `sccache`, but the proper AWS credentials  weren't available to this repository. I've added those credentials to the `rapids-cmake` repository and it seems that that has resolved the slowdown issue.

- https://github.com/rapidsai/ci-imgs/pull/19/files

robertmaynard (approved):

ajschmidt8:
@gpucibot merge

PR #303, opened 2022-11-15
Remove usage of FetchContent from tests to improve perf

robertmaynard:
## Description
Improve performance of rapids-cmake tests in CI by removing the usage of fetch content, and instead
using symlinks and tricking fetch content that it was used to get rapids-cmake.

This reduces the size of each test in the binary directory as they no longer have a copy of rapids-cmake. Plus still works
for tests that fetch rapids-cmake indirectly ( via cpm due to testing projects like rmm ). 

## Checklist

### DIFF
--- a/testing/utils/project_template.cmake.in
+++ b/testing/utils/project_template.cmake.in
@@ -18,18 +18,21 @@ cmake_minimum_required(VERSION 3.23.1)
 include(FetchContent)
 
 set(local-rapids-cmake-root "${rapids-cmake-dir}/..")
-set(rapids-cmake_SOURCE_DIR "${CMAKE_CURRENT_BINARY_DIR}/_deps/rapids-cmake-src")
-set(rapids-cmake_BINARY_DIR "${CMAKE_CURRENT_BINARY_DIR}/_deps/rapids-cmake-build")
+cmake_path(NORMAL_PATH local-rapids-cmake-root)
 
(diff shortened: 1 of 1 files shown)


ajschmidt8:
> Were tests particularly slow before? This PR seems like it introduces more boilerplate to speed up something that didn't appear to be a bottleneck, but I may have just been unaware of it. The new approach seems like a cleaner handling of local changes, though, which is a definite improvement over the old copying approach.

This PR was originally created in an attempt to resolve the issue described in the comment below. It turns out these changes were not the cause of the problem.

- https://github.com/rapidsai/rapids-cmake/pull/301#issuecomment-1315635528

robertmaynard:
While we found that `sccache` was causing the worst of slow-downs, this does solves a problem with rapids-cmake tests.

In general the inital configure of each test will include a full git clone of the local copy of rapids-cmake. Depending on the file-system, and size of the `.git` folder you can see a significant cost. I know locally I had to delete my src directory since initial configures of tests was taking ~20sec.

In testing locally with a freshly cloned rapids-cmake I see this shaving ~2sec from each test, which is really nice when most tests only take ~4-6sec.

vyasr (approved):

robertmaynard:
@gpucibot merge

robertmaynard:
@gpucibot merge

PR #308, opened 2022-11-17
Don't use CMake 3.25.0 as it has a show stopping FindCUDAToolkit bug

robertmaynard:
## Description
Don't use CMake 3.25.0 as it has a show stopping FindCUDAToolkit bug


ajschmidt8 (approved):

vyasr:
@gpucibot merge

PR #324, opened 2022-12-07
rapids_cuda_patch_toolkit: Better handle non-standard toolkits

### LINKED ISSUE #321: [BUG] rapids_cuda_patch_toolkit 'fixes' should be version guarded
**Describe the bug**
The fixes in rapids_cuda_patch_toolkit aren't version guarded so it potentially could cause problems if target names change

**Expected behavior**
rapids-cmake only applies the fixes to versions of FindCUDAToolkit as needed

robertmaynard:
## Description
Previously we presumed that `cublas`, `cusparse`, and `cusolver` would always be found and offer both static and shared versions. Now we only setup these rules if the correct targets exist.

Fixes #321 

## Checklist

### DIFF
--- a/rapids-cmake/cuda/patch_toolkit.cmake
+++ b/rapids-cmake/cuda/patch_toolkit.cmake
@@ -32,6 +32,7 @@ of cublas and cusolver targets are incorrect. This module must be called
 from the same CMakeLists.txt as the first `find_project(CUDAToolkit)` to
 patch the targets.
 
+For all version of Cmake the dependencies of cusparse are incorrect.
 .. note::
   :cmake:command:`rapids_cpm_find` will automatically call this module
   when asked to find the CUDAToolkit.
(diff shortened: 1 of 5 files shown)


rapids-cmake/cuda/patch_toolkit.cmake
```diff
@@ -32,6 +32,7 @@ of cublas and cusolver targets are incorrect. This module must be called
 from the same CMakeLists.txt as the first `find_project(CUDAToolkit)` to
 patch the targets.
 
+For all version of Cmake the dependencies of cusparse are incorrect.
```
bdice, line 35:
```suggestion
For all versions of CMake, the dependencies of cusparse are incorrect.
```

bdice: Is there an upstream fix needed to CMake? If an issue has been reported upstream, can it be cross-referenced here? "All versions are broken" sounds like a statement that will hopefully be changed in the future.

robertmaynard: I am in the process of figuring out if this line of docs is correct or not. Looking at the history it is unclear to me if this is needed for a subset of CUDA versions, or if this was done due to incorrect downstream CUDA linking and we 'fixed' it in rapids-cmake.

robertmaynard: I can't find an occurence where cusparse depends on cublas. Playing around with link lines using the static libraries I can't find any configuration order that causes a breakage ( 11.0, 11.4, 11.6. 11.8 ).

So I will drop the cusparse -> cublas linkage.

robertmaynard:
@gpucibot merge

PR #328, opened 2022-12-12
Add `rapids_test` allowing projects to run gpu tests in parallel

robertmaynard:
## Description
Introduces `rapids_test` functionality to allow tests executed via `ctest -j` to properly resource share GPUs. 

This is done by having tests state how many GPUs allocations they require, and uses CTest internal job scheduler to properly load balance.
## Checklist


docs/cpp_code_snippets/rapids_cmake_ctest_allocation.cpp
```diff
@@ -0,0 +1,101 @@
...
+  auto slots = gpu_resources.substr(slot_start);
+
+  return GPUAllocation{std::stoi(id), std::stoi(slots)};
+}
+
+std::vector<GPUAllocation> determineGPUAllocations() {
+  std::vector<GPUAllocation> allocations;
+  const auto *resource_count = std::getenv("CTEST_RESOURCE_GROUP_COUNT");
+  if (!resource_count) {
+    allocations.push_back(std::move(noGPUAllocation()));
```
bdice, line 56: Can this be an emplace?

docs/cpp_code_snippets/rapids_cmake_ctest_allocation.cpp
```diff
@@ -0,0 +1,101 @@
...
+  for (int index = 0; index < resource_max; ++index) {
+    std::string group_env = "CTEST_RESOURCE_GROUP_" + std::to_string(index);
+    std::string resource_group{std::getenv(group_env.c_str())};
+    std::transform(resource_group.begin(), resource_group.end(),
+                   resource_group.begin(), ::toupper);
+
+    if (resource_group == "GPUS") {
+      auto resource_env = group_env + "_" + resource_group;
+      auto &&allocation = parseCTestAllocation(resource_env);
+      allocations.push_back(std::move(allocation));
```
bdice, line 70: Can this be an emplace?

docs/cpp_code_snippets/rapids_cmake_ctest_allocation.cpp
```diff
@@ -0,0 +1,101 @@
...
+  const auto resource_max = std::stoi(resource_count);
+  for (int index = 0; index < resource_max; ++index) {
+    std::string group_env = "CTEST_RESOURCE_GROUP_" + std::to_string(index);
+    std::string resource_group{std::getenv(group_env.c_str())};
+    std::transform(resource_group.begin(), resource_group.end(),
+                   resource_group.begin(), ::toupper);
+
+    if (resource_group == "GPUS") {
+      auto resource_env = group_env + "_" + resource_group;
+      auto &&allocation = parseCTestAllocation(resource_env);
```
bdice, line 69: Are we using the same clang-format settings as cudf? I’d like to adopt those if not. It can be a separate PR, and I’m happy to do that if you agree it is worth doing.

docs/cpp_code_snippets/rapids_cmake_ctest_allocation.hpp
```diff
@@ -0,0 +1,87 @@
...
+struct GPUAllocation {
+  int device_id;
+  int slots;
+};
+
+/*
+ * Returns true when a CTest resource specification has been specified.
+ *
+ * Since the vast majority of tests should execute without a CTest resource
+ * spec ( e.g. when executed manually be developer ), callers of `rapids_cmake`
```
bdice, line 39:
```suggestion
 * spec (e.g. when executed manually by a developer), callers of `rapids_cmake`
```

docs/cpp_code_snippets/rapids_cmake_ctest_allocation.hpp
```diff
@@ -0,0 +1,87 @@
...
+ *     rapids_cmake::bind_to_first_gpu();
+ *     }
+ * ```
+*/
+bool using_resources();
+
+/*
+ * Returns all GPUAllocations allocated for a test
+ *
+ * To support multi-gpu tests the CTest resource specification allows a
```
bdice, line 54:
```suggestion
 * To support multi-GPU tests the CTest resource specification allows a
```

docs/cpp_code_snippets/rapids_cmake_ctest_allocation.hpp
```diff
@@ -0,0 +1,87 @@
...
+ * test to request multiple GPUs. As CUDA only allows binding to a
+ * single GPU at any time, this API allows tests to know what CUDA
+ * devices they should bind to.
+ *
+ * Note: The `device_id` of each allocation might not be unique.
+ * If a test says they need 50% of two gpu's they could be allocated
```
bdice, line 60:
```suggestion
 * If a test says it needs 50% of two GPUs, it could be allocated
```

docs/cpp_code_snippets/rapids_cmake_ctest_allocation.hpp
```diff
@@ -0,0 +1,87 @@
...
+ * the same physical GPU. If a test needs distinct / unique devices
+ * they must request 51%+ of a device.
```
bdice, line 62:
```suggestion
 * it must request 51%+ of a device.
```

docs/cpp_code_snippets/rapids_cmake_ctest_allocation.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+/*
+ * Have CUDA bind to a given GPUAllocation
+ *
+ * Have CUDA bind to the `device_id` specified in the CTest
+ * GPU allocation
+ *
+ * Note: Return value is the cudaError_t of `cudaSetDevice`
+ */
+int bind_to_gpu(GPUAllocation const &alloc);
```
bdice, line 77: Should the return type be `cudaError_t`?

docs/cpp_code_snippets/rapids_cmake_ctest_allocation.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+/*
+ * Convience method to bind to the first GPU that CTest has allocated
```
bdice, line 80:
```suggestion
 * Convenience method to bind to the first GPU that CTest has allocated
```

docs/cpp_code_snippets/rapids_cmake_ctest_allocation.hpp
```diff
@@ -0,0 +1,87 @@
...
+ * Provided as most RAPIDS tests only require a single GPU
+ *
+ * Note: Return value is the cudaError_t of `cudaSetDevice`
+ */
+bool bind_to_first_gpu();
```
bdice, line 85: Same as above, should return `cudaError_t`. Or at least int, not bool.

docs/hardware_resources_and_testing.rst
```diff
@@ -0,0 +1,122 @@
...
+
+Resource Allocation
+*******************
+
+CTest resource allocation frameworks allow tests to specify which hardware resources that they need, and for projects to specify the specific local/machine resources available.
+Combined together this ensures that tests are told which specific resources they should use, and ensures over-subscription won't occur no matter the requested testing parallel level.
+
+To get CTest resource allocation used by tests the following components are needed.
+
+  - A json per-machine resource specification file
```
bdice, line 16:
```suggestion
  - A JSON per-machine resource specification file
```

docs/hardware_resources_and_testing.rst
```diff
@@ -0,0 +1,122 @@
...
+  - The :cmake:variable:`CTEST_RESOURCE_SPEC_FILE` points to the json file
```
bdice, line 17:
```suggestion
  - The :cmake:variable:`CTEST_RESOURCE_SPEC_FILE` points to the JSON file
```

docs/hardware_resources_and_testing.rst
```diff
@@ -0,0 +1,122 @@
...
+  - Each :cmake:command:`add_test` records what resources it requires via test properties
+  - Each test reads the relevant environment variables to determine
+    what specific resources it should use
+
+
+These are steep requirements that require large amounts of infrastructure setup for each project.
+In addition the CTest resource allocation specification is very relaxed, allowing it to represent abitrary requirements such as CPUs, GPUs, and ASICs.
```
bdice, line 24:
```suggestion
In addition the CTest resource allocation specification is very relaxed, allowing it to represent arbitrary requirements such as CPUs, GPUs, and ASICs.
```

docs/hardware_resources_and_testing.rst
```diff
@@ -0,0 +1,122 @@
...
+rapids_test
+***********
+
+To help RAPIDS projects utilize all GPUs on a machine when running tests, the ``rapids-cmake`` project offers a suite of commands to simplify the process.
+These commands simplify GPU detection, setting up resource specification files, specifying test requirements, and setting the active CUDA GPU.
+
+Machine GPU Detection
+*********************
+
+The key component of CTest resource allocation is having an accurate representation of the hardware that exists on the developers machine.
```
bdice, line 35:
```suggestion
The key component of CTest resource allocation is having an accurate representation of the hardware that exists on the developer's machine.
```

docs/hardware_resources_and_testing.rst
```diff
@@ -0,0 +1,122 @@
...
+The :cmake:command:`rapids_test_init` function will do system introspection to determine the number of GPUs on the current machine and generate a resource allocation json file representation these GPUs.
```
bdice, line 36:
```suggestion
The :cmake:command:`rapids_test_init` function will do system introspection to determine the number of GPUs on the current machine and generate a resource allocation JSON file representing these GPUs.
```

docs/hardware_resources_and_testing.rst
```diff
@@ -0,0 +1,122 @@
...
+.. code-block:: cmake
+
+  include(${CMAKE_BINARY_DIR}/RAPIDS.cmake)
+
+  include(rapids-test)
+
+  enable_testing()
+  rapids_test_init()
+
+The CTest resource allocation specification isn't limited to representing GPUs a single unit.
```
bdice, line 47:
```suggestion
The CTest resource allocation specification isn't limited to representing GPUs as a single unit.
```

rapids-cmake/test/detail/detect_number_of_gpus.cmake
```diff
@@ -0,0 +1,81 @@
...
+
+      find_package(CUDAToolkit)
+      set(compiler ${CMAKE_CXX_COMPILER})
+      if(DEFINED CMAKE_CUDA_COMPILER)
+        set(compiler ${CMAKE_CUDA_COMPILER})
+      endif()
+      execute_process(COMMAND ${compiler} ${eval_file} -o ${eval_exe} -I${CUDAToolkit_INCLUDE_DIRS}
+                              -L${CUDAToolkit_LIBRARY_DIR} -lcudart ERROR_FILE ${error_file})
+      if(NOT EXISTS "${eval_exe}")
+        message(STATUS "rapids_test_detect_number_of_gpus failed to build detection executable, presuming 1 GPU."
```
bdice, line 70: Why are these not hard errors?

rapids-cmake/test/generate_resource_spec.cmake
```diff
@@ -0,0 +1,75 @@
...
+#=============================================================================
+include_guard(GLOBAL)
+
+#[=======================================================================[.rst:
+rapids_test_generate_resource_spec
+----------------------------------
+
+.. versionadded:: v23.02.00
+
+Generates a json resource specification file representing the machines GPUs
```
bdice, line 24:
```suggestion
Generates a JSON resource specification file representing the machine's GPUs
```

rapids-cmake/test/generate_resource_spec.cmake
```diff
@@ -0,0 +1,75 @@
...
+using system introspection.
+
+  .. code-block:: cmake
+
+    rapids_test_generate_resource_spec( DESTINATION filepath DETECT)
+
+Generates a json resource specification file representing the machines GPUs
```
bdice, line 31:
```suggestion
Generates a JSON resource specification file representing the machine's GPUs
```

rapids-cmake/test/generate_resource_spec.cmake
```diff
@@ -0,0 +1,75 @@
...
+using system introspection. This will allow CTest to schedule multiple
+single-gpu tests in parallel on multi-gpu machines.
```
bdice, line 33:
```suggestion
single-GPU tests in parallel on multi-GPU machines.
```

rapids-cmake/test/gpu_requirements.cmake
```diff
@@ -0,0 +1,78 @@
...
+#=============================================================================
+include_guard(GLOBAL)
+
+#[=======================================================================[.rst:
+rapids_test_gpu_requirements
+----------------------------
+
+.. versionadded:: v23.02.00
+
+States how many GPUs and what precent of each a test requires
```
bdice, line 24:
```suggestion
States how many GPUs and what percent of each a test requires.
```

rapids-cmake/test/gpu_requirements.cmake
```diff
@@ -0,0 +1,78 @@
...
+When combined with :cmake:command:`rapids_test_init` informs CTest what
+resources should be allocated to a test so that when testing in parallel
+oversubscription doesn't occur. Without this information user execution of
+CTest with high parallel levels will cause multiple tests to run on the
+same GPU and quickly exhaust all memory.
+
+
+``GPUS``
+  State how many GPUs this test requires. Allows CTest to not over-subscribe
+  a machines hardware.
```
bdice, line 43:
```suggestion
  a machine's hardware.
```

rapids-cmake/test/gpu_requirements.cmake
```diff
@@ -0,0 +1,78 @@
...
+
+``PERCENT``
+  State how much of each GPU this test requires. In general 100,50, and 20
```
bdice, line 46:
```suggestion
  State how much of each GPU this test requires. In general 100, 50, and 20
```

rapids-cmake/test/gpu_requirements.cmake
```diff
@@ -0,0 +1,78 @@
...
+  set(one_value GPUS PERCENT)
+  set(multi_value)
+  cmake_parse_arguments(_RAPIDS_TEST "${options}" "${one_value}" "${multi_value}" ${ARGN})
+
+  set(gpus ${_RAPIDS_TEST_GPUS})
+  if(NOT gpus)
+    message(FATAL_ERROR "rapids_test_gpu_requirements requires the GPUS option to be provided.")
+  endif()
+  if(gpus LESS 1 OR (NOT gpus MATCHES "^[0-9]+$"))
+    message(FATAL_ERROR "rapids_test_gpu_requirements GPUS requires a numeric value ( >= 1) provided ${gpus}."
```
bdice, line 62:
```suggestion
    message(FATAL_ERROR "rapids_test_gpu_requirements GPUS requires a numeric value (>= 1) provided ${gpus}."
```

rapids-cmake/test/gpu_requirements.cmake
```diff
@@ -0,0 +1,78 @@
...
+    )
+  endif()
+
+  set(percent 100)
+  if(DEFINED _RAPIDS_TEST_PERCENT)
+    set(percent ${_RAPIDS_TEST_PERCENT})
+  endif()
+
+  # verify that percent is inside the allowed bounds
+  if(percent GREATER 100 OR percent LESS 1 OR (NOT percent MATCHES "^[0-9]+$"))
```
bdice, line 72: Why is zero not legal? Are CPU-only tests supposed to be added in a different way?

rapids-cmake/test/init.cmake
```diff
@@ -0,0 +1,64 @@
...
+#=============================================================================
+include_guard(GLOBAL)
+
+#[=======================================================================[.rst:
+rapids_test_init
+----------------
+
+.. versionadded:: v23.02.00
+
+Establish necessary components for CTest gpu resource allocation to allow
```
bdice, line 24:
```suggestion
Establish necessary components for CTest GPU resource allocation to allow
```

rapids-cmake/test/init.cmake
```diff
@@ -0,0 +1,64 @@
...
+for parallel tests
```
bdice, line 25:
```suggestion
for parallel tests.
```

rapids-cmake/test/init.cmake
```diff
@@ -0,0 +1,64 @@
...
+
+  .. code-block:: cmake
+
+    rapids_test_init(  )
+
+Generates a json resource specification file representing the machines GPUs
```
bdice, line 31:
```suggestion
Generates a JSON resource specification file representing the machine's GPUs
```

rapids-cmake/test/init.cmake
```diff
@@ -0,0 +1,64 @@
...
+using system introspection. This will allow CTest to schedule multiple
+single-gpu or multi-gpu tests in parallel on multi-gpu machines.
```
bdice, line 33:
```suggestion
single-GPU or multi-GPU tests in parallel on multi-GPU machines.
```

rapids-cmake/test/init.cmake
```diff
@@ -0,0 +1,64 @@
...
+  rapids_test_init()
+
+  add_executable( test_example test.cu )
+  rapids_test_add(NAME single_gpu_alloc COMMAND test_example GPUS 1)
+  rapids_test_add(NAME two_gpu_alloc COMMAND test_example GPUS 2)
+
+Result Variables
+^^^^^^^^^^^^^^^^
+  :cmake:variable:`CTEST_RESOURCE_SPEC_FILE` will be set to the generated
+  json file if not already set
```
bdice, line 51:
```suggestion
  JSON file if not already set
```

testing/test/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
+#=============================================================================
+# Copyright (c) 2021, NVIDIA CORPORATION.
```
bdice, line 2:
```suggestion
# Copyright (c) 2022, NVIDIA CORPORATION.
```

testing/test/generate_resource_spec-invalid-mode.cmake
```diff
@@ -0,0 +1,18 @@
+#=============================================================================
+# Copyright (c) 2021, NVIDIA CORPORATION.
```
bdice, line 2:
```suggestion
# Copyright (c) 2022, NVIDIA CORPORATION.
```

testing/test/generate_resource_spec-no-gpus.cmake
```diff
@@ -0,0 +1,37 @@
+#=============================================================================
+# Copyright (c) 2021, NVIDIA CORPORATION.
```
bdice, line 2:
```suggestion
# Copyright (c) 2022, NVIDIA CORPORATION.
```

testing/test/generate_resource_spec-no-gpus.cmake
```diff
@@ -0,0 +1,37 @@
...
+if(NOT CTEST_RESOURCE_SPEC_FILE STREQUAL "sentinel file")
+  message(FATAL_ERROR "CTEST_RESOURCE_SPEC_FILE shouldn't be modified by calling rapids_test_generate_resource_spec")
+endif()
+
+if(NOT EXISTS "${CMAKE_CURRENT_BINARY_DIR}/spec.json")
+  message(FATAL_ERROR "rapids_test_generate_resource_spec failed to write out the requested spec file")
+endif()
+
+file(READ "${CMAKE_CURRENT_BINARY_DIR}/spec.json" content)
+if( content MATCHES 100)
```
bdice, line 35:
```suggestion
if(content MATCHES 100)
```

testing/test/generate_resource_spec-simple.cmake
```diff
@@ -0,0 +1,29 @@
+#=============================================================================
+# Copyright (c) 2021, NVIDIA CORPORATION.
```
bdice, line 2: And similarly for all files.
```suggestion
# Copyright (c) 2022, NVIDIA CORPORATION.
```

testing/test/init-activate_allocation_simple/main.cu
```diff
@@ -0,0 +1,18 @@
+
+#include <iostream>
+#include <vector>
+
+
+int main() {
+
+  // Very we only have a single GPU visible to us
```
bdice, line 8:
```suggestion
  // Verify we only have a single GPU visible to us
```

testing/test/init-activate_multi_allocations_same_gpu/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+rapids_test_init()
+
+add_executable(verify_alloc main.cu)
+
+enable_testing()
+rapids_test_add(NAME simple_test COMMAND verify_alloc GPUS 10 PERCENT 1)
+
+# Validate that the test runs correctly
+set_tests_properties(simple_test
+  PROPERTIES PASS_REGULAR_EXPRESSION "Seeing atleast a single GPU"
```
bdice, line 30:
```suggestion
  PROPERTIES PASS_REGULAR_EXPRESSION "Seeing at least a single GPU"
```

testing/test/init-activate_multi_allocations_same_gpu/main.cu
```diff
@@ -0,0 +1,17 @@
...
+int main() {
+
+  // Very we only have a single GPU visible to us
+  int nDevices = 0;
+  cudaGetDeviceCount(&nDevices);
+
+  if (nDevices == 0) {
+    return 1;
+  }
+  std::cout << "Seeing atleast a single GPU" << std::endl;
```
bdice, line 15:
```suggestion
  std::cout << "Seeing at least a single GPU" << std::endl;
```

testing/test/init-activate_multi_jobs_single_gpu/main.cpp
```diff
@@ -0,0 +1,38 @@
+
```
bdice, line 1: This empty line at the beginning seems unnecessary. Perhaps also missing a license?

bdice (changes requested):
This is a cool feature. I am guessing we want this to parallelize RAPIDS tests in CI? One concern is that percentages may not be the most useful metric. Some tests might allocate 8 GB of memory regardless of the GPU size, meaning that the percentage is different on a 32 GB GPU than a 48 GB GPU.

docs/cpp_code_snippets/rapids_cmake_ctest_allocation.cpp
robertmaynard: `rapids-cmake` doesn't really have a clang-format setup since it generally has no c++ code. With this being in the documents I am fine running a formatter over it once and comitting it.

testing/test/init-activate_multi_jobs_single_gpu/main.cpp
robertmaynard: I have updated all the the `.cpp` and `.cu` files in the PR to have licenses.

testing/test/init-activate_allocation_simple/main.cu
robertmaynard: Done

rapids-cmake/cmake/support_conda_env.cmake
```diff
@@ -30,6 +30,7 @@ Establish a target that holds the CONDA include and link directories.
 Creates a global interface target called `target_name` that holds
 the CONDA include and link directories, when executed.
 
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
robertmaynard: Yeah :(

rapids-cmake/test/gpu_requirements.cmake
robertmaynard: In my mind CPU only tests would not be using `rapids_test_add` and instead they would be using `add_test`. They would still be executed in parallel when run via `ctest -jN` but wouldn't be constrained by the job scheduler.

But I do see the argument that having everything use `rapids_test_add` makes the calling code easier. In that case we would need to only support `GPUS 0`. Since statement like `GPUS 4 PERCENT 0` make no sense at all ( to me )

docs/cpp_code_snippets/rapids_cmake_ctest_allocation.cpp
robertmaynard: yes, and updated

docs/cpp_code_snippets/rapids_cmake_ctest_allocation.cpp
robertmaynard: I have updated with a default formatting. If we are interested in making sure all cpp/cuda code anywhere in rapids-cmake matches the RAPIDS / libcudf style we should do that in a followup PR ( I am not opposed to that )

docs/cpp_code_snippets/rapids_cmake_ctest_allocation.hpp
robertmaynard: I didn't want to use a `cudaError_t` for this as it can return a failure state when no allocation is provided and we shouldn't represent that as a CUDA failure.

docs/cpp_code_snippets/rapids_cmake_ctest_allocation.hpp
robertmaynard: Yes it should. Updated

robertmaynard:
> I am guessing we want this to parallelize RAPIDS tests in CI? 

Yes, and other NVIDIA projects. This first step should allow some projects like Thrust to have correct parallel support now.
But we are missing a component for RAPIDS ( see below ) CI, but this will alllow local developers using `ctest`.

> One concern is that percentages may not be the most useful metric. Some tests might allocate 8 GB of memory regardless of the GPU size, meaning that the percentage is different on a 32 GB GPU than a 48 GB GPU.

>  detect_number_of_gpus.cmake `why not hard errors`?

So this comes back to the missing part needed for RAPIDS. There is no hard requirement when configuring and building RAPIDS projects that you have an actual GPU on the machine. Therefore we shouldn't expect that the test harness code to start hard erroring when it executes in these enviornments. RAPIDS CI runs the configure && build steps without any GPU devices so we want to gracefully fallback to presuming `1` GPU in those cases.

This comes back to the missing component. What rapids-cmake needs to offer is some executable / process for CI structures like RAPIDS to re-detect the number of GPUs on the machine before `ctest` is run. This way CI can build the binaries on machine A and ship to `B` and properly use the hardware.

I am currently trying to figure out an approach for this, but didn't want to hold up the initial PR

robertmaynard:
> Some tests might allocate 8 GB of memory regardless of the GPU size, meaning that the percentage is different on a 32 GB GPU than a 48 GB GPU.

I wanted to keep the usage as simple as possible. I am worried that we use `HIGH_WATER_MEMORY` instead of `PERCENT` it will make people think they need to be super accurate.

Another issue with using `MEMORY` is how we describe tests that use all GPU resources. We can't say each test requires 1TB of memory as that request can't be fulfilled. Likewise we can't use a magic placeholder like `ALL` since the GPUs that exist at CMake configure time might not match the GPUs that exist when we run the tests / update tool.

docs/cpp_code_snippets/rapids_cmake_ctest_allocation.hpp
bdice: Are the docs wrong, then? It says "Note: Return value is the `cudaError_t` of `cudaSetDevice`"

robertmaynard: The docs have been updated to ( when I corrected `bind_to_gpu` ): 
```
 * Will return `false` if no GPUs have been allocated, or if setting
 * the CUDA device failed for any reason.
 ```

File: docs/cpp_code_snippets/rapids_cmake_ctest_allocation.cpp (hunk 3 above)
Comment (robertmaynard) [reply docs/cpp_code_snippets/rapids_cmake_ctest_allocation.cpp]: Latest round of changes also brings in a `.clang-format` file to the root of rapids-cmake, since now the GPU detection + json file writing is done by a C++ executable we build

File: rapids-cmake/test/install_relocatable.cmake [hunk 41]
```diff
@@ -0,0 +1,110 @@
...
+  set(CTEST_SCRIPT_DIRECTORY ".")
+  set(CTEST_RESOURCE_SPEC_FILE "./resource_spec.json")
+  execute_process(COMMAND ./generate_ctest_json OUTPUT_FILE "${CTEST_RESOURCE_SPEC_FILE}")
+  ]==])
+
+  foreach(test IN LISTS tests_to_run)
+    get_test_property(${test} INSTALL_COMMAND command)
+    get_test_property(${test} RESOURCE_GROUPS resources)
+    string(APPEND content "add_test([=[${test}]=] ${command})\n")
+    string(APPEND content "set_tests_properties([=[${test}]=] PROPERTIES RESOURCE_GROUPS ${resources})\n")
```
robertmaynard, line 81: Need to export any labels on tests as well

bdice (commented):
@robertmaynard I started a review a while ago (probably a month ago) and didn't click submit. I am submitting those comments now -- forgive me if the comments are outdated.

ajschmidt8 (approved):

rapids-cmake/test/gpu_requirements.cmake
robertmaynard: I have updated `rapids_test_gpu_requirements` and `rapids_test_add` to better support the GPU workflow.

If you specify `rapids_test_add` without GPUS at all it presumes that you don't need any GPU allocation for that test. `rapids_test_gpu_requirements` still requires a GPUS argument but it support a `0` value.

This should make using the normal API easier ( `rapids_test_add` ) for projects that have a mixture of CPU and GPU tests

rapids-cmake/test/detail/generate_resource_spec.cpp
```diff
@@ -0,0 +1,86 @@
...
+  int id        = 0;
+  size_t memory = 0;
+  int slots     = 0;
+};
+
+struct local {
+  local()
+  {
+    int nDevices = 0;
+    cudaGetDeviceCount(&nDevices);
```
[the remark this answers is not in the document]
robertmaynard: It doesn't need CUDA error handling since it is outputting a json file on standard output. When the program can't use any of the CUDA runtime API's it will generate a json output that is equal to a machine with a valid CUDA runtime but no GPUs which is what we want ( no gpu support ).

rapids-cmake/test/install_relocatable.cmake
```diff
@@ -0,0 +1,112 @@
...
+include_guard(GLOBAL)
+
+#[=======================================================================[.rst:
+rapids_test_install_relocatable
+-------------------------------
+
+.. versionadded:: v23.04.00
+
+Install the needed `ctest` infrastructure to allow installed tests to be run
+by `ctest` in parallel with GPU awarness.
```
bdice, line 25:
```suggestion
by `ctest` in parallel with GPU awareness.
```

rapids-cmake/test/install_relocatable.cmake
```diff
@@ -0,0 +1,112 @@
...
+
+  .. code-block:: cmake
+
+    rapids_test_install_relocatable(INSTALL_COMPONENT_SET <component>
+                                    DESTINATION <relative_path>
+                                    [EXCLUDE_FROM_ALL])
+
+Add ...
```
bdice, line 33: What does this line mean? Is this an incomplete sentence?

.clang-format
```diff
@@ -0,0 +1,155 @@
+---
```
bdice, line 1: I would suggest adding this `.clang-format` configuration in a separate PR. It looks like it altered a handful of files that are unrelated to this PR's central focus.

bdice (approved):
This PR is pretty hefty. I skimmed it again (most code was familiar) and had only a few minor suggestions. I'll approve this and you can handle my suggestions as you wish.

rapids-cmake/test/install_relocatable.cmake
robertmaynard: Looks like I forgot to write all the documentation for what the function does. I will go ahead and actually write the document

.clang-format
robertmaynard: That makes sense I will go ahead and do that now

rapids-cmake/test/add.cmake
```diff
@@ -0,0 +1,149 @@
...
+                                 ${_RAPIDS_TEST_INSTALL_COMPONENT_SET})
+    else()
+      set(command_for_install ${command_or_target})
+    endif()
+    rapids_test_record_test_component(NAME ${_RAPIDS_TEST_NAME} COMPONENT
+                                      ${_RAPIDS_TEST_INSTALL_COMPONENT_SET})
+    rapids_test_record_test_command(NAME
+                                    ${_RAPIDS_TEST_NAME}
+                                    COMMAND
+                                    cmake
```
vyasr, line 143: Better to be `CMAKE_COMMAND`?
```suggestion
                                    ${CMAKE_COMMAND}
```

vyasr: Actually, even better might be that we save the command used for `add_test` and then reuse it here.

rapids-cmake/test/install_relocatable.cmake
```diff
@@ -0,0 +1,116 @@
...
+
+    rapids_test_install_relocatable(INSTALL_COMPONENT_SET <component>
+                                    DESTINATION <relative_path>
+                                    [EXCLUDE_FROM_ALL])
+
+Will install all tests created by :cmake:command:`rapids_test_add` that are
+part of the provided ``INSTALL_COMPONENT_SET``.
+
+The :cmake:command:`rapids_test_install_relocatable` presumes that all
+arguments provided to the tests are machine independent ( no absolute paths ).
```
vyasr, line 37:
```suggestion
arguments provided to the tests are machine independent (no absolute paths).
```

rapids-cmake/test/gpu_requirements.cmake
```diff
@@ -0,0 +1,92 @@
...
+  if(NOT gpus GREATER_EQUAL 0)
+    message(FATAL_ERROR "rapids_test_gpu_requirements requires a numeric GPUS value [0-N].")
+  endif()
+  if(NOT (percent GREATER_EQUAL 1 AND percent LESS_EQUAL 100))
+    message(FATAL_ERROR "rapids_test_gpu_requirements requires a numeric PERCENT value [1-100].")
+  endif()
+
+  if(gpus AND percent)
+    set_property(TEST ${test_name} PROPERTY RESOURCE_GROUPS "${gpus},gpus:${percent}")
+  endif()
```
vyasr, line 90: This conditional looks superfluous since we're setting them above.

rapids-cmake/test/add.cmake
```diff
@@ -0,0 +1,149 @@
...
+
+``GPUS``
+  State how many GPUs this test requires. Allows CTest to not over-subscribe
+  a machine's hardware.
+
+  If no value is provided, the test is considered to not use any GPUs
+
+``PERCENT``
+  By default if no percent is provided, 100 is used.
+  State how much of each GPU this test requires.
```
vyasr, line 55: I was initially thinking that it would make for a nicer API if we could use a floating point value for GPUs to indicate fractional usage, but seeing how you've implemented the translation to a resource group with 100 slots per GPU I see why you went with percent as a separate integral argument. Do you think it would introduce too much internal complexity to be worth implementing that API? I can see the complexities around needing to round and validate a float.

rapids-cmake/test/add.cmake
```diff
@@ -0,0 +1,149 @@
...
+When combined with :cmake:command:`rapids_test_init` informs CTest what
+resources should be allocated to a test so that when testing in parallel
+oversubscription doesn't occur. Without this information user execution of
+CTest with high parallel levels will cause multiple tests to run on the
+same GPU and quickly exhaust all memory.
+
+``COMMAND``
+  Specify the test command-line. If <command> specifies an executable target
+  (created by add_executable()) it will automatically be replaced by the location of the
+  executable created at build time.
```
vyasr, line 45: It would be helpful to indicate that COMMAND isn't just a single executable but can be the entire command. I don't know the best way to phrase this, but I mention it because I was initially confused by the way that args are separated out from the command below.

rapids-cmake/test/install_relocatable.cmake
```diff
@@ -0,0 +1,116 @@
...
+  endif()
+
+  get_target_property(targets_to_install rapids_test_install_${component} TARGETS_TO_INSTALL)
+  get_target_property(tests_to_run rapids_test_install_${component} TESTS_TO_RUN)
+
+  set(content
+      [==[
+  set(CTEST_SCRIPT_DIRECTORY ".")
+  set(CTEST_RESOURCE_SPEC_FILE "./resource_spec.json")
+  execute_process(COMMAND ./generate_ctest_json OUTPUT_FILE "${CTEST_RESOURCE_SPEC_FILE}")
```
vyasr, line 77: IIUC this code is necessary because while init.cmake would have generated the resource spec for the machine that the code was built on, this snippet is being embedded in the CMake code that is shipped with the built library so that it can be used to run tests via ctest on an installed binary. Is that right?

If that is right, then I supposed it's not that important for the filenames to match exactly in the two cases, but nonetheless if consistency is the goal could we store the filenames like `resource_spec.json` and `generate_ctest_json` in some centralized constants.json file? The current approach is a bit error-prone w.r.t. modifications needing to happen in both places.

rapids-cmake/test/add.cmake
vyasr: One advantage of the float approach is that it removes some ambiguity over what argument sets are valid. For instance, does it make sense to do `GPUS 2 PERCENT 25` as compared to `GPUS 1 PERCENT 50`? This may also be construed as a disadvantage if you intended to allow that kind of spec where a test requires multiple fractional GPUs, though.

vyasr (changes requested):
This is really cool!

I have a general question about properties that came up a few times in this review: can you `set_property` with any arbitrary property string, even if it doesn't already exist? Do you not need to `define_property` first in those cases?

I haven't reviewed the tests yet, will do that in a subsequent pass.

dependencies.yaml
```diff
@@ -41,23 +41,28 @@ dependencies:
             packages:
               - cudatoolkit=11.2
               - gcc<11.0.0
+              - sysroot_linux-64==2.17
```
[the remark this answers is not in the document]
robertmaynard: Just after we switched over to dependencies.yaml we ran into link issues with some of the tests that called cuda runtime funcitons. I will roll back these changes and see if other CI updates have made this unneeded

rapids-cmake/test/add.cmake
robertmaynard: We have to replicate the logic for mapping from a cmake target to actual executable name that `add_test` does internally. That is why we can't just re-use the same arguments provided to `add_test`.

( Needed as the target name and output executable name don't need to match ).

rapids-cmake/test/generate_resource_spec.cmake
```diff
@@ -0,0 +1,94 @@
...
+  if(NOT EXISTS "${eval_exe}")
+    find_package(CUDAToolkit REQUIRED)
+    file(MAKE_DIRECTORY "${PROJECT_BINARY_DIR}/rapids-cmake/")
+
+    set(compile_options "-I${CUDAToolkit_INCLUDE_DIRS}")
+    set(link_options ${CUDA_cudart_LIBRARY})
+    set(compiler "${CMAKE_CXX_COMPILER}")
+    if(NOT DEFINED CMAKE_CXX_COMPILER)
+      set(compiler "${CMAKE_CUDA_COMPILER}")
+    endif()
```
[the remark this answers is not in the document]
robertmaynard: Correct

rapids-cmake/test/gpu_requirements.cmake
robertmaynard: When `gpus` == 0 that evaluates as false and therefore we don't record a requirement in the resource group.

I couldn't decide on how to structure this function. Should things like `GPUS 0 PERCENT 200` produce an error message or early terminate due. I went with validate all inputs before termination for consistency

rapids-cmake/test/init.cmake
```diff
@@ -0,0 +1,64 @@
...
+^^^^^^^^^^^^^^^^
+  :cmake:variable:`CTEST_RESOURCE_SPEC_FILE` will be set to the generated
+  JSON file if not already set
+
+#]=======================================================================]
+function(rapids_test_init)
+  list(APPEND CMAKE_MESSAGE_CONTEXT "rapids.test.init")
+
+  include(${CMAKE_CURRENT_FUNCTION_LIST_DIR}/generate_resource_spec.cmake)
+  rapids_test_generate_resource_spec(DESTINATION "${PROJECT_BINARY_DIR}/resource_spec.json")
```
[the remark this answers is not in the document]
robertmaynard: I agree on the idempotency should be in the functon itself.

dependencies.yaml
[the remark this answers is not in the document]
robertmaynard: Removing these lines results in CI failures for `11.4.1, ubuntu20.04, amd64, 3.8, v100, 520` and `11.8.0, ubuntu22.04, amd64, 3.10, v100, 520`, so it looks like we need to pin glibc version so that we have the correct symbols for building the `rapids_test_generate_resource_spec` executable.

```
-- rapids_test_generate_resource_spec compile failure details are /opt/conda/envs/test/bin/../lib/gcc/x86_64-conda-linux-gnu/10.4.0/../../../../x86_64-conda-linux-gnu/bin/ld: /opt/conda/envs/test/lib/libcudart.so: undefined reference to `memcpy@GLIBC_2.14'
/opt/conda/envs/test/bin/../lib/gcc/x86_64-conda-linux-gnu/10.4.0/../../../../x86_64-conda-linux-gnu/bin/ld: /opt/conda/envs/test/lib/libcudart.so: undefined reference to `clock_gettime@GLIBC_2.17'
```

README.md
```diff
@@ -108,6 +108,16 @@ The most commonly used function are:
 - `rapids_find_package(<project_name> BUILD_EXPORT_SET <name> INSTALL_EXPORT_SET <name> )` Combines `find_package` and support to track dependencies for easy package exporting
 - `rapids_generate_module(<PackageName> HEADER_NAMES <paths...> LIBRARY_NAMES <names...> )` Generate a FindModule for the given package. Allows association to export sets so the generated FindModule can be shipped with the project
 
+### test
+
+The `rapids_test` functions simplify CTest resource allocation, allowing for
+tests to run in parallel without overallocating GPU resources.
+
+The most commonly used functions are:
+- `rapids_test_add(NAME <test_name> GPUS <N>)` State how many GPU resources a single
```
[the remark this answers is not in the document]
robertmaynard: updated to include `PERCENT`

rapids-cmake/test/add.cmake
robertmaynard: Made this clearer.

rapids-cmake/test/add.cmake
```diff
@@ -0,0 +1,149 @@
...
+  endif()
+
+  # Provide a copy of the test runner in the binary directory so that tests still can be executed if
+  # for some reason rapids-cmake src has been removed.
+  set(_rapids_run_gpu_test_script "${PROJECT_BINARY_DIR}/rapids-cmake/run_gpu_test.cmake")
+  set(_rapids_run_gpu_test_script_for_install "./run_gpu_test.cmake")
+  if(NOT EXISTS "${_rapids_run_gpu_test_script}")
+    file(COPY "${CMAKE_CURRENT_FUNCTION_LIST_DIR}/detail/run_gpu_test.cmake"
+         DESTINATION "${PROJECT_BINARY_DIR}/rapids-cmake/")
+  endif()
```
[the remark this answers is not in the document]
robertmaynard: The install rule is generated via `install_relocatable.cmake`

rapids-cmake/test/add.cmake
robertmaynard: We also can't re-use `CMAKE_COMMAND` since it encodes the absolute path to cmake on the build machine which might not be the same.

We also can't ask `CMAKE_COMMAND` to be evaluated when executing the test as that variable doesn't exist when ctest executes the test file ( No built-in cmake variables exist in this mode )

rapids-cmake/test/detail/record_install.cmake
```diff
@@ -0,0 +1,43 @@
...
+function(rapids_test_record_install)
+  list(APPEND CMAKE_MESSAGE_CONTEXT "rapids.test.record_install")
+
+  set(options)
+  set(one_value TARGET COMPONENT)
+  set(multi_value)
+  cmake_parse_arguments(_RAPIDS_TEST "${options}" "${one_value}" "${multi_value}" ${ARGN})
+
+  set(component ${_RAPIDS_TEST_COMPONENT})
+  set_property(TARGET rapids_test_install_${component} APPEND PROPERTY "TARGETS_TO_INSTALL"
```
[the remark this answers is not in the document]
robertmaynard: > Is TARGETS_TO_INSTALL an existing property, or something you've invented for this use case? If the latter, do we not need to define_property first?
It is new property that we will create and set via `set_property`. No need to use `define_property` first.

> why quote the literal TARGETS_TO_INSTALL?

Style only. Happy to change.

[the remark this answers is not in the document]
robertmaynard: Removed the quoted string

rapids-cmake/test/detail/run_gpu_test.cmake
```diff
@@ -0,0 +1,33 @@
...
+  # cmake-lint: disable=E1120
+  foreach(index RANGE 0 ${CTEST_RESOURCE_GROUP_COUNT})
+    set(allocation $ENV{CTEST_RESOURCE_GROUP_${index}_GPUS})
+    if(DEFINED allocation)
+      # strings look like "id:value,slots:value" so let's make a super lazy parser by deleting `id:`
+      # and replacing `,slots:` with `;` so we have a list with two items.
+      string(REPLACE "id:" "" allocation "${allocation}")
+      string(REPLACE ",slots:" ";" allocation "${allocation}")
+      list(GET allocation 0 device_ids)
+      list(GET allocation 1 percent)
```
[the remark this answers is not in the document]
robertmaynard: I had left it so other people understood what the item in the list represented. But will switch it to a comment so people know what it is, but doesn't `cost` anything

rapids-cmake/test/generate_resource_spec.cmake
```diff
@@ -0,0 +1,94 @@
...
+files to be generated.
+
+``DESTINATION``
+  Location that the JSON output from the detection should be written to
+
+.. note::
+    Unlike rapids_test_init this doesn't set CTEST_RESOURCE_SPEC_FILE
+
+#]=======================================================================]
+function(rapids_test_generate_resource_spec DESTINATION filepath)
```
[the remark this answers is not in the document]
robertmaynard: No we need to regenerate each time we invoke CMake since the user can execute CMake with `CUDA_VISIBLE_DEVICES` set to narrow the set of GPUs they want used.

rapids-cmake/test/add.cmake
robertmaynard: I prefer the percentage approach as it keeps the CMake logic simple. Mapping from decimals back to integer slots will cause edge cases around rounding and unreprenstable value ( e.g. 0.1  not mapping to a max of 10 jobs concurrently).

rapids-cmake/test/install_relocatable.cmake
robertmaynard: Updated with file names moved to a common location

rapids-cmake/test/detail/generate_resource_spec.cpp
```diff
@@ -0,0 +1,87 @@
...
+    buffer << "\n";
+  }
+  buffer << "\t]\n";
+  buffer << "}]\n";
+}
+
+int main(int, char**)
+{
+  version v{1, 0};
+  local l;
```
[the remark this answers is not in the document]
robertmaynard: removed the `local` class

rapids-cmake/test/detail/generate_resource_spec.cpp
```diff
@@ -0,0 +1,87 @@
...
+  buffer << "\"local\": [{\n";
+  buffer << "\t\"gpus\": [\n";
+  for (int i = 0; i < l.gpus.size(); ++i) {
+    to_json(buffer, l.gpus[i]);
+    if (i != (l.gpus.size() - 1)) { buffer << ","; }
+    buffer << "\n";
+  }
+  buffer << "\t]\n";
+  buffer << "}]\n";
+}
```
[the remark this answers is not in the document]
robertmaynard: I have updated the file to link to the resource spec file.

My end goal was that users didn't need to know about the resource-specification file format.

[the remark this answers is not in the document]
vyasr: Oh OK if that is the case then yes I think what you've done (adding links only to the source) is sufficient.

rapids-cmake/test/add.cmake
```diff
@@ -0,0 +1,149 @@
...
+
+  add_test(NAME ${_RAPIDS_TEST_NAME}
+           COMMAND ${CMAKE_COMMAND} "-Dcommand_to_run=${command}" "-Dcommand_args=${args}" -P
+                   "${_rapids_run_gpu_test_script}"
+           WORKING_DIRECTORY "${_RAPIDS_TEST_WORKING_DIRECTORY}")
```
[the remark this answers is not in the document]
vyasr: We discussed this further offline and decided that the tradeoff is worthwhile. The main benefits of using `gtest_discover_tests` can also be attained in other ways, mostly by leveraging environment variables that gtest supports for doing things like filtering.

rapids-cmake/test/generate_resource_spec.cmake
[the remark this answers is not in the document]
vyasr: I didn't consider that option, that is a nice use case to support.

rapids-cmake/test/gpu_requirements.cmake
vyasr: I think when I first reached this point in the review I didn't realize that the way to indicate a fractional GPU was to do GPUS 0 PERCENT X. With that context I think this result is a bit awkward but a reasonable choice.

vyasr (approved):
LGTM! It would be nice if we could somehow template the testing files with preprocessor macros or something since there's so much duplication (50% of this PR seems like it's just the license headers) but other than that everything looks good.

PR #329, opened 2022-12-12
Update cuco git tag to fetch bug fixes and cleanups

PointKernel:
This PR fetches the latest changes in cuco to:

- Inline `cuco::sentinel` namespace (https://github.com/NVIDIA/cuCollections/pull/247)
- Fix memory leak in `static_map::retrieve_all` (https://github.com/NVIDIA/cuCollections/pull/257)
- Use public `cuco::murmurhash3_32` (https://github.com/NVIDIA/cuCollections/pull/253)

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/rapids-cmake/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.
- [x] The `cmake-format.json` is up to date with these changes.
- [ ] I have added new files under rapids-cmake/
   - [ ] I have added include guards (`include_guard(GLOBAL)`)
   - [ ] I have added the associated docs/ rst file and update the api.rst

PR #331, opened 2022-12-16
Branch 23.02 merge 22.12

vyasr:
## Description

Replaces #326 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,40 @@
 
 Please see https://github.com/rapidsai/rapids-cmake/releases/tag/v23.02.00a for the latest changes to this development branch.
 
-# rapids-cmake 22.12.00 (Date TBD)
+# rapids-cmake 22.12.00 (8 Dec 2022)
 
-Please see https://github.com/rapidsai/rapids-cmake/releases/tag/v22.12.00a for the latest changes to this development branch.
+## 🐛 Bug Fixes
+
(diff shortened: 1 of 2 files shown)


vyasr:
Something seems to have changed in conda that's triggering an installation of gcc 11.x instead of 10.x in the CTK 11.2 environment. We have other PRs currently open that select the correct gcc version (e.g. #329) so I'm not sure why this is happening. @ajschmidt8 do you know if it's safe to include c-compiler _and_ include a gcc < 11 pin in the same env? If not, then we may need to investigate why the solve changed.

ajschmidt8:
> Something seems to have changed in conda that's triggering an installation of gcc 11.x instead of 10.x in the CTK 11.2 environment. We have other PRs currently open that select the correct gcc version (e.g. #329) so I'm not sure why this is happening. @ajschmidt8 do you know if it's safe to include c-compiler _and_ include a gcc < 11 pin in the same env? If not, then we may need to investigate why the solve changed.

I know very little about the inner workings of `rapids-cmake`. And even less about compiler compatibility. I don't think I'm the best person to consult here.

robertmaynard:
> Something seems to have changed in conda that's triggering an installation of gcc 11.x instead of 10.x in the CTK 11.2 environment. We have other PRs currently open that select the correct gcc version (e.g. #329) so I'm not sure why this is happening. @ajschmidt8 do you know if it's safe to include c-compiler _and_ include a gcc < 11 pin in the same env? If not, then we may need to investigate why the solve changed.

@vyasr This approach works and would be how I would do this

PR #336, opened 2023-01-04
Use pre-commit for style checks

bdice:
## Description
This PR adopts pre-commit for style checks in rapids-cmake. In addition to cmake-format and cmake-lint, this will help with frequently encountered PR review notes, like formatting C++ code, standardizing copyright notices, and catching typos via codespell.

For now, codespell and whitespace linting are disabled to reduce PR size.

## Checklist


ci/checks/run-cmake-format.sh
```diff
@@ -0,0 +1,61 @@
...
+  echo "${DEFAULT_RAPIDS_CMAKE_FORMAT_FILE}"
+  echo ""
+  echo "Try setting the environment variable RAPIDS_CMAKE_FORMAT_FILE to the path to the config file."
+  exit 0
+else
+  echo "Using format file ${RAPIDS_CMAKE_FORMAT_FILE}"
+fi
+
+if [[ $1 == "cmake-format" ]]; then
+  for cmake_file in "${@:2}"; do
```
bdice, line 47:
```suggestion
  # We cannot pass multiple input files because of a bug in cmake-format.
  # See: https://github.com/cheshirekow/cmake_format/issues/284
  for cmake_file in "${@:2}"; do
```

ci/checks/run-cmake-format.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
+
+# Copyright (c) 2021-2023, NVIDIA CORPORATION.
+
+# This script is a wrapper for cmakelang that may be used with pre-commit. The
```
bdice, line 5: Should we use the upstream pre-commit hook (https://github.com/cheshirekow/cmake-format-precommit) instead of copying the `run-cmake-format.sh` file, since this repo contains the original config file and it doesn't need to be fetched?

robertmaynard: I am happy to defer this to a follow up PR

robertmaynard (approved):

ci/check_style.sh
```diff
@@ -17,58 +17,5 @@ set +eu
...
-  echo -e "\n\n>>>> PASSED: cmake lint check\n\n"
-fi
-
-RETVALS=($CMAKE_FORMAT_RETVAL $CMAKE_LINT_RETVAL)
-IFS=$'\n'
-RETVAL=`echo "${RETVALS[*]}" | sort -nr | head -n1`
-
-exit $RETVAL
+# Run pre-commit checks
+pre-commit run --all-files --show-diff-on-failure
```
ajschmidt8, line 21: I can't make a suggestion for this via the web interface, but can you remove the `+e` on line 16 of this file?

we can probably remove the `set -u` flags throughout this file now too since it's so succinct. on lines 4, 16, and 18.

ajschmidt8 (approved):

ci/checks/run-cmake-format.sh
```diff
@@ -0,0 +1,61 @@
...
+  echo ""
+  echo "Try setting the environment variable RAPIDS_CMAKE_FORMAT_FILE to the path to the config file."
+  exit 0
+else
+  echo "Using format file ${RAPIDS_CMAKE_FORMAT_FILE}"
+fi
+
+if [[ $1 == "cmake-format" ]]; then
+  for cmake_file in "${@:2}"; do
+    cmake-format --in-place --first-comment-is-literal --config-files ${RAPIDS_CMAKE_FORMAT_FILE} ${RAPIDS_CMAKE_ROOT}/ci/checks/cmake_config_format.json -- ${cmake_file}
```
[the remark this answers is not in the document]
vyasr: I have not seen this before either. Your diagnosis seems reasonable, though, and as you say below switching to a direct invocation should bypass the issue.

ci/checks/run-cmake-format.sh
vyasr: I think we should switch, yes. The fetch was the only reason I implemented the script in the first place. Hopefully switching to the hook will also address the issue you observed with `--first-comment-is-literal`.

bdice: I attempted to switch to the official pre-commit hook, but failed. There is no option to require `pre-commit` to run one file at a time, so the use of `--first-comment-is-literal` means we always run into the bug mentioned above. We have to use an intermediate shell script to split the work and execute `cmake-format` on individual files, so I guess we're stuck with using `run-cmake-format.sh`.

PR #342, opened 2023-01-19
Update spdlog to 1.11 ( latest version )

robertmaynard:
## Description
RAPIDS is moving to the latest spdlog version to help minimize integration issues with the conda community

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/rapids-cmake/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #346, opened 2023-01-23
Revert "Update spdlog to 1.11 ( latest version ) (#342)"

bdice:
## Description
This reverts commit 3d5c0a1e0d94b0882f714cce05cb9a3bd1b0e3c7 / PR #342.

The update to `spdlog 1.11.0` was tested for `cuml` (and seemed fine) but broke builds for `cugraph` both locally and in CI:
- https://github.com/rapidsai/cugraph/actions/runs/3980866464/jobs/6824225844
- https://github.com/rapidsai/rapids-cmake/pull/342#issuecomment-1399183315

I am reverting that change until a diagnosis can be performed and solution can be found.

See also this packaging change for rmm (not yet merged): https://github.com/rapidsai/rmm/pull/1177

## Checklist

### DIFF
--- a/rapids-cmake/cpm/versions.json
+++ b/rapids-cmake/cpm/versions.json
@@ -32,7 +32,7 @@
       "git_tag" : "branch-${version}"
     },
     "spdlog" : {
-      "version" : "1.11.0",
+      "version" : "1.8.5",
       "git_url" : "https://github.com/gabime/spdlog.git",
       "git_tag" : "v${version}"
     },


robertmaynard (approved):

robertmaynard:
This reads to me as a namespace resolution lookup issue.

Looking over the includes the compile issue looks to occur if either of the following includes occur:
```cpp
#include "Static/KTruss/KTruss.impl.cuh"
#include "Static/KTruss/KTrussWeighted.impl.cuh"
```
Looking into both of those files they both do:
```cpp
using namespace std;
using namespace rmm;
```

Which is really bad for a header file as it can cause global lookup issues when trying to resolve unspecified namespace types.

Locally if I can remove all 4 of those `using namespace` commands I get cugraph to compile with spdlog 1.11

bdice:
Thanks @robertmaynard. I have identified that `using namespace rmm` is only present in the cuhornet repository, one of cugraph's dependencies. I will refactor cuhornet to avoid `using namespace rmm` and `using namespace std`.

https://github.com/search?q=repo%3Arapidsai%2Fcuhornet%20%22using%20namespace%20rmm%22&type=code

PR #358, opened 2023-02-01
Make NAMESPACE property truly optional in rapids_export

agirault:
## Description

This tackles two issues, broken down in two commits:

1. If no `NAMESPACE` is given (optional), the configured `config.cmake` file had invalid CMake code which would fail on import:
    ```
    CMake Error at <project>-config.cmake:90 (list):
      list sub-command TRANSFORM, action PREPEND expects 1 argument(s).
    ```
2. Even with (1) fixed, `NAMESPACE` is not a true opt-in property (as per [guidelines](https://github.com/rapidsai/rapids-cmake/blob/branch-23.04/CONTRIBUTING.md#code-contributions)) even if optional, as it uses a default value when not set (`${project_name}::`) which can conflict with the property `EXPORT_NAME` set on targets, outside of `rapids_export`. The second commit of this MR changes the default behavior to set no additional namespace (see commit message for details).

## Checklist
    - ~~Looking for current tests. Unlikely since (1) is a bug with the previous behavior.~~
    - ~~Can't figure what the semantics of this file are. `"NAMESPACE": 1` ?~~
- [ ] ~~I have added new files under rapids-cmake/~~


vyasr:
@agirault thanks for the contribution! Sorry it took so long to look at this.

Regarding the JSON file, I don't think there's any change that you need to make here. The format file already looks to be configured to allow a missing NAMESPACE, even if the code would have raise an error before. If you're curious, [here's the spec for that format file](https://cmake-format.readthedocs.io/en/latest/custom_parsers.html). `NAMESPACE` basically falls into the simplest default case where there is a keyword argument that only has a single positional argument of length one, so you get `"NAMESPACE": 1`.

It looks like your file does need some reformatting. You should be able to get the formatting locally using `pre-commit`. If you're not familiar, just running `pip install pre-commit; pre-commit install; pre-commit run --all-files` in the root of your local repo should take care of it.

robertmaynard (approved):
Agreeing with @vyasr thanks for finding this issue.

As far as testing, please add the following test:

1. A test simillar to `export-verify-build-namespace`  which is a two phase test that generates a project-config.cmake file and loads it to verify the targets exist.

rapids-cmake/export/export.cmake
```diff
@@ -182,9 +185,9 @@ function(rapids_export type project_name)
     rapids_export_parse_version(${_RAPIDS_VERSION} rapids_orig rapids_project_version)
   endif()
 
-  set(_RAPIDS_PROJECT_NAMESPACE "${project_name}::")
-  if(DEFINED _RAPIDS_NAMESPACE)
-    set(_RAPIDS_PROJECT_NAMESPACE ${_RAPIDS_NAMESPACE})
+  if(NOT DEFINED _RAPIDS_NAMESPACE)
+    message(NOTICE "rapids-cmake EXPORT: no NAMESPACE was provided. `${project_name}::` is recommended \
```
robertmaynard, line 189: I think this should be a `WARNING`  or maybe `VERBOSE`. We haven't used `NOTICE` before in rapids-cmake as that outputs in the error style and this use could be intentional.

rapids-cmake/export/export.cmake
```diff
@@ -80,11 +80,14 @@ calls to :cmake:command:`find_dependency`, or :cmake:command:`CPMFindPackage`.
...
+  A recommended namespace could be `<project_name>::`.
+  If not provided, no namespace is used.
 
-  Note: When exporting with `BUILD` type, only `GLOBAL_TARGETS` will
-  be placed in the namespace.
+  Note:
+  - When exporting with `BUILD` type, only `GLOBAL_TARGETS` will be placed
+  in the namespace.
+  - The namespace can be configured on a per-target basis instead using the
+  EXPORT_NAME property.
```
robertmaynard, line 90:
```suggestion
  :cmake:prop_tgt:`EXPORT_NAME <cmake:prop_tgt:EXPORT_NAME>`  property.
```
So we cross reference the official docs for `EXPORT_NAME`

robertmaynard (changes requested):

vyasr:
@agirault were you planning on finishing this PR up for 23.04? Let us know if you need any additional guidance!

robertmaynard:
retargetting this for 23.06

rapids-cmake/export/export.cmake
agirault: I don't love the `WARNING` since users could have done that on purpose and used one directly with `EXPORT_NAME`. `VERBOSE` then?

robertmaynard: Lets go with `VERBOSE`

robertmaynard (changes requested):
We need to revise the original statement that no explicit usage of `NAMESPACE` will generate a bad `config.cmake` file. That no longer is the case since the merging of the `COMPONENTS` PR.

That means the only change that exists is the removal of the automatic `NAMESPACE` which would be a behavior change. Given that `rapids_export` is widely used we should use the new `rapids_cmake_policy` feature to warn people about the new behavior, and remove the implicit support in 23.08.

vyasr:
I've rebased this PR on 23.08 and fixed a couple issues that I found. The full test suite now passes for me locally. I'll address the review feedback next.

vyasr:
Actually @robertmaynard I'm not quite sure what you are suggesting that we do in your last review. After the components PR we can still associate a namespace with each component's export, which I have now fixed. I agree that removing the implicit namespace is a behavior change needing a policy change. Were you saying that one of the original bugs had now been fixed by the merge of the components PR?

rapids-cmake/export/detail/component.cmake
```diff
@@ -51,18 +51,28 @@ function(rapids_export_component type project_name component_name export_set uni
...
+              DESTINATION "${install_location}" COMPONENT ${component_name})
+    endif()
 
   else()
     set(install_location "${PROJECT_BINARY_DIR}")
     set(deps_destination "${install_location}/")
 
-    export(EXPORT ${export_set} NAMESPACE ${namespace}
-           FILE "${install_location}/${project_name}-${unique_name}-targets.cmake")
+    if(namespace STREQUAL "")
```
robertmaynard, line 69: The contents of both the `if` and `else` here are the same.

rapids-cmake/export/detail/component.cmake
```diff
@@ -51,18 +51,28 @@ function(rapids_export_component type project_name component_name export_set uni
     file(MAKE_DIRECTORY "${deps_destination}")
     install(DIRECTORY "${deps_destination}" DESTINATION "${install_location}"
             COMPONENT ${component_name})
-    install(EXPORT ${export_set}
-            FILE ${project_name}-${unique_name}-targets.cmake
-            NAMESPACE ${namespace}
-            DESTINATION "${install_location}"
-            COMPONENT ${component_name})
+    if(namespace STREQUAL "")
```
robertmaynard, line 54: I think this `if` / `else` is inverted. If `namespace` is empty we shouldn't specify a `NAMESPACE` argument to `install(EXPORT`

robertmaynard (changes requested):

rapids-cmake/export/detail/component.cmake
vyasr: You are correct looks like I made an error or two during conflict resolution.

vyasr:
I snuck in the change to run ctests in parallel, which we haven't been so far.

robertmaynard (approved):

PR #368, opened 2023-02-09
Bump spdlog to 1.11, add fmt as dependency for spdlog

kkraus14:
## Description

Bumps spdlog to 1.11.0 and adds `fmt` as a dependency of spdlog and sets the `SPDLOG_FMT_EXTERNAL_HO` cmake option for spdlog. Happy to change that to be controlled by an argument to the `rapids_cpm_spdlog` function if desired. I believe the existing testing should cover the changes made here as the `spdlog` and `rmm` related tests were failing until adding `fmt` to the `spdlog` `BUILD_EXPORT_SET` and `INSTALL_EXPORT_SET`.

## Checklist

### DIFF
--- a/rapids-cmake/cpm/spdlog.cmake
+++ b/rapids-cmake/cpm/spdlog.cmake
@@ -39,10 +39,11 @@ across all RAPIDS projects.
   option only controls the behavior when spdlog is fetched and built, NOT when an installed spdlog is found on the
   system.
 
-  This option can be set to one of three values: `BUNDLED`, `EXTERNAL_FMT`, or `EXTERNAL_FMT_HO`. If set to
+  This option can be set to: `BUNDLED`, `EXTERNAL_FMT`, `EXTERNAL_FMT_HO`, or `STD_FORMAT`. If set to
   `BUNDLED`, then spdlog will use its own bundled version of fmt. If set to `EXTERNAL_FMT` then spdlog will use the
   `fmt::fmt` target and be linked with the fmt library. If set to `EXTERNAL_FMT_HO` then spdlog will use the
-  `fmt::fmt-header-only` target and be linked with a header only fmt library.
+  `fmt::fmt-header-only` target and be linked with a header only fmt library. If set to `STD_FORMAT` then spdlog
+  will use `std::format` instead of the fmt library.
 
   Defaults to `EXTERNAL_FMT_HO`.
 
@@ -94,23 +95,26 @@ function(rapids_cpm_spdlog)
     set(_RAPIDS_FMT_OPTION "EXTERNAL_FMT_HO")
   endif()
 
-  if("${_RAPIDS_FMT_OPTION}" STREQUAL "BUNDLED")
+  if(_RAPIDS_FMT_OPTION STREQUAL "BUNDLED")
     set(spdlog_fmt_option "")
-  elseif("${_RAPIDS_FMT_OPTION}" STREQUAL "EXTERNAL_FMT" OR "${_RAPIDS_FMT_OPTION}" STREQUAL
-                                                            "EXTERNAL_FMT_HO")
+  elseif(_RAPIDS_FMT_OPTION STREQUAL "EXTERNAL_FMT" OR _RAPIDS_FMT_OPTION STREQUAL
+                                                       "EXTERNAL_FMT_HO")
     include("${rapids-cmake-dir}/cpm/fmt.cmake")
 
     # Using `spdlog_ROOT` needs to cause any internal find calls in `spdlog-config.cmake` to first
     # search beside it before looking globally.
     list(APPEND fmt_ROOT ${spdlog_ROOT})
 
-    rapids_cpm_fmt(BUILD_EXPORT_SET spdlog INSTALL_EXPORT_SET spdlog)
-    set(spdlog_fmt_option "SPDLOG_${fmt_target_name} ON")
-    if("${_RAPIDS_FMT_OPTION}" STREQUAL "EXTERNAL_FMT")
+    rapids_cpm_fmt(BUILD_EXPORT_SET ${_RAPIDS_BUILD_EXPORT_SET}
+                   INSTALL_EXPORT_SET ${_RAPIDS_INSTALL_EXPORT_SET})
+    set(spdlog_fmt_option "SPDLOG_${_RAPIDS_FMT_OPTION} ON")
+    if(_RAPIDS_FMT_OPTION STREQUAL "EXTERNAL_FMT")
       set(spdlog_fmt_target fmt::fmt)
-    elseif("${_RAPIDS_FMT_OPTION}" STREQUAL "EXTERNAL_FMT_HO")
+    elseif(_RAPIDS_FMT_OPTION STREQUAL "EXTERNAL_FMT_HO")
       set(spdlog_fmt_target fmt::fmt-header-only)
     endif()
+  elseif(_RAPIDS_FMT_OPTION STREQUAL "STD_FORMAT")
+    set(spdlog_fmt_option "SPDLOG_USE_${_RAPIDS_FMT_OPTION} ON")
   else()
     message(FATAL_ERROR "Invalid option used for FMT_OPTION, got: ${_RAPIDS_FMT_OPTION}, expected one of: 'BUNDLED', 'EXTERNAL_FMT', 'EXTERNAL_FMT_HO'"
     )
--- a/rapids-cmake/cpm/spdlog.cmake
+++ b/rapids-cmake/cpm/spdlog.cmake
@@ -35,6 +35,8 @@ across all RAPIDS projects.
                      [<CPM_ARGS> ...])
 
 ``FMT_OPTION``
+.. versionadded:: v23.04.00
+
   Spdlog depends on the fmt library and offers multiple ways of handling this dependency when spdlog is built. This
   option only controls the behavior when spdlog is fetched and built, NOT when an installed spdlog is found on the
   system.
@@ -71,7 +73,8 @@ function(rapids_cpm_spdlog)
   set(multi_value)
   cmake_parse_arguments(_RAPIDS "${options}" "${one_value}" "${multi_value}" ${ARGN})
 
-  # Fix up _RAPIDS_UNPARSED_ARGUMENTS to have EXPORT_SETS as this is need for rapids_cpm_find
+  # Fix up _RAPIDS_UNPARSED_ARGUMENTS to have EXPORT_SETS as this is need for rapids_cpm_find. Also
+  # propagate the user provided build and install export sets.
   if(_RAPIDS_INSTALL_EXPORT_SET)
     list(APPEND _RAPIDS_UNPARSED_ARGUMENTS INSTALL_EXPORT_SET ${_RAPIDS_INSTALL_EXPORT_SET})
   endif()
@@ -105,8 +108,8 @@ function(rapids_cpm_spdlog)
     # search beside it before looking globally.
     list(APPEND fmt_ROOT ${spdlog_ROOT})
 
-    rapids_cpm_fmt(BUILD_EXPORT_SET ${_RAPIDS_BUILD_EXPORT_SET}
-                   INSTALL_EXPORT_SET ${_RAPIDS_INSTALL_EXPORT_SET})
+    rapids_cpm_fmt(${_RAPIDS_UNPARSED_ARGUMENTS})
+
     set(spdlog_fmt_option "SPDLOG_${_RAPIDS_FMT_OPTION} ON")
     if(_RAPIDS_FMT_OPTION STREQUAL "EXTERNAL_FMT")
       set(spdlog_fmt_target fmt::fmt)
--- a/rapids-cmake/cpm/spdlog.cmake
+++ b/rapids-cmake/cpm/spdlog.cmake
@@ -29,9 +29,22 @@ across all RAPIDS projects.
 
 .. code-block:: cmake
 
-  rapids_cpm_spdlog( [BUILD_EXPORT_SET <export-name>]
+  rapids_cpm_spdlog( [FMT_OPTION <fmt-option-name>]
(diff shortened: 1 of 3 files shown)


rapids-cmake/cpm/spdlog.cmake
```diff
@@ -62,6 +62,10 @@ function(rapids_cpm_spdlog)
   include("${rapids-cmake-dir}/cpm/detail/generate_patch_command.cmake")
   rapids_cpm_generate_patch_command(spdlog ${version} patch_command)
 
+  # Add fmt in the case where we need to fetch spdlog from source
+  include("${rapids-cmake-dir}/cpm/fmt.cmake")
+  rapids_cpm_fmt(BUILD_EXPORT_SET spdlog INSTALL_EXPORT_SET spdlog)
```
kkraus14, line 67: I'm not sure if this is the proper way to be doing this, but I figure it would have the following behavior:
- In the case where we're finding spdlog from a conda package, this will find fmt as well
- In the case where we're finding spdlog not from a conda package, this will find fmt if it is already in the system, otherwise it will fetch it and build it (as far as I can tell there isn't a CMake option to avoid building a library for fmt)
- In the case where we're fetching spdlog, this will find fmt if it is already in the system, otherwise it will fetch it and build it

This generally seems like the correct behavior that we want.

I'm not sure if setting the exports as I've done here for `spdlog` is correct, but without these tests would fail from what looked like trying to export spdlog targets.

robertmaynard: The problem with this approach is that it won't obey `spdlog_ROOT` and the associated search behavior. If a project has extenerally built spdlog and fmt. Using `spdlog_ROOT` would cause any internal find calls in `spdlog-config.cmake` to first search beside it before looking globally.

I think we need to do the following:

1. Have the current call to `rapids_cpm_fmt` be opt-in.
2. Do `list(APPEND fmt_ROOT ${spdlog_ROOT})` before calling `rapids_cpm_fmt`
3. Set `SPDLOG_FMT_EXTERNAL_HO` only when opting in

kkraus14: > The problem with this approach is that it won't obey `spdlog_ROOT` and the associated search behavior. If a project has extenerally built spdlog and fmt. Using `spdlog_ROOT` would cause any internal find calls in `spdlog-config.cmake` to first search beside it before looking globally.

Are you saying that if someone sets `spdlog_ROOT` that it should search for the `fmt` dependency of spdlog beside `spdlog_ROOT` before using the standard search paths? That feels a bit funky to me, but if that's standard CMake behavior so be it.

Will tackle those 3 bullets.

robertmaynard: > Are you saying that if someone sets spdlog_ROOT that it should search for the fmt dependency of spdlog beside spdlog_ROOT before using the standard search paths?

Yes when the `find_package` call for fmt is inside the `spd` config.cmake file. This is the purpose of the `<PackageName>_ROOT` variables they are meant as a way to re-root the search behavior so that all dependency package queries first search 'beside' the primary project.

kkraus14: Should be addressed in the latest commits, PTAL

rapids-cmake/cpm/spdlog.cmake
```diff
@@ -62,16 +89,43 @@ function(rapids_cpm_spdlog)
...
+    set(spdlog_fmt_option "")
+  elseif("${_RAPIDS_FMT_OPTION}" STREQUAL "EXTERNAL_FMT" OR "${_RAPIDS_FMT_OPTION}" STREQUAL
+                                                            "EXTERNAL_FMT_HO")
+    include("${rapids-cmake-dir}/cpm/fmt.cmake")
+
+    # Using `spdlog_ROOT` needs to cause any internal find calls in `spdlog-config.cmake` to first
+    # search beside it before looking globally.
+    list(APPEND fmt_ROOT ${spdlog_ROOT})
+
+    rapids_cpm_fmt(BUILD_EXPORT_SET spdlog INSTALL_EXPORT_SET spdlog)
```
robertmaynard, line 107: I think we want to propagate the user provided build and install export sets. If the user hasn't enabled an install export set for spdlog we shouldn't have one for fmt. ( Would be wrong to install fmt and not spdlog ).

rapids-cmake/cpm/spdlog.cmake
robertmaynard: Added some comments, but overall looks good

rapids-cmake/cpm/spdlog.cmake
```diff
@@ -29,10 +29,23 @@ across all RAPIDS projects.
 
 .. code-block:: cmake
 
-  rapids_cpm_spdlog( [BUILD_EXPORT_SET <export-name>]
+  rapids_cpm_spdlog( [FMT_OPTION <fmt-option-name>]
+                     [BUILD_EXPORT_SET <export-name>]
                      [INSTALL_EXPORT_SET <export-name>]
                      [<CPM_ARGS> ...])
 
+``FMT_OPTION``
```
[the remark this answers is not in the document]
robertmaynard: Wouldn't a value of `CXX_FMT` for the `FMT_OPTION` be sufficient for the `std::format` case?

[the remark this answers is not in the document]
kkraus14: Yes, I could add another option I suppose, but calling it `FMT_OPTION` is maybe a bit misleading at that point?

I would also probably call it `STD_FORMAT` instead of `CXX_FMT` just to be a little bit more consistent in mapping the spdlog CMake options to our options here.

rapids-cmake/cpm/spdlog.cmake
kkraus14: My intention with this was to attach fmt as a build and install export to the spdlog target where then if the user added spdlog to a build or install export, fmt would be added transitively. I'm not sure if that's possible, but if it was I'm guessing I should be doing it against the `spdlog::spdlog` and `spdlog::spdlog_header_only` targets instead.

robertmaynard: That would be possible by passing the same BUILD and EXPORT set down to `fmt` that was provided to `spdlog`.

That will ensure that fmt is installed when spdlog is installed ( and vice versa ) plus will ensure `find_package(fmt)` and `find_package(spdlog)` are performed by the project that is associated with the export set.

rapids-cmake/cpm/spdlog.cmake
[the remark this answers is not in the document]
robertmaynard: I was hand waving away the ambgious shortform of `FMT` to be the formatting engine used by spdlog and not to mean the fmt package specifically.

rapids-cmake/cpm/spdlog.cmake
kkraus14: I believe I addressed this in the way you suggested, but I'm not 100% sure if I did what you were thinking. PTAL

rapids-cmake/cpm/spdlog.cmake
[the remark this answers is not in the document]
kkraus14: Added an option of `STD_FORMAT`

rapids-cmake/cpm/spdlog.cmake
```diff
@@ -29,10 +29,24 @@ across all RAPIDS projects.
...
```
robertmaynard, line 37:
```suggestion
``FMT_OPTION``
.. versionadded:: v23.04.00
```

robertmaynard (changes requested):

rapids-cmake/cpm/spdlog.cmake
robertmaynard: You will need to do the following:

```cmake
  set(user_export_set_args)
  if(_RAPIDS_INSTALL_EXPORT_SET)
    list(APPEND user_export_set_args INSTALL_EXPORT_SET ${_RAPIDS_INSTALL_EXPORT_SET})
  endif()
  if(_RAPIDS_BUILD_EXPORT_SET)
    list(APPEND user_export_set_args BUILD_EXPORT_SET ${_RAPIDS_BUILD_EXPORT_SET})
  endif()
  list(APPEND _RAPIDS_UNPARSED_ARGUMENTS ${user_export_set_args})

...

rapids_cpm_fmt(${user_export_set_args})
  ```
  
  That way we properly handle cases where only a BUILD_EXPORT_SET or INSTALL_EXPORT_SET is provided.

Comment (kkraus14) [reply rapids-cmake/cpm/spdlog.cmake]: I had already been updating `_RAPIDS_UNPARSED_ARGUMENTS` like this, so I just reused that variable for passing to `rapids_cpm_fmt`. Should be addressed in the latest commit.

I wasn't sure if the `_RAPIDS_UNPARSED_ARGUMENTS` is used elsewhere, but it didn't look like it.

Comment (robertmaynard) [reply rapids-cmake/cpm/spdlog.cmake]: That is an interesting solution. Using `_RAPIDS_UNPARSED_ARGUMENTS` might actually be better as now any CPM options passed to `rapids_cpm_spdlog` are also propagated to `rapids_cpm_fmt` which could be useful.

**Review by robertmaynard — APPROVED** [review]

File: rapids-cmake/cpm/spdlog.cmake [hunk 5]
```diff
@@ -62,16 +93,46 @@ function(rapids_cpm_spdlog)
...
+    set(spdlog_fmt_option "SPDLOG_${_RAPIDS_FMT_OPTION} ON")
+    if(_RAPIDS_FMT_OPTION STREQUAL "EXTERNAL_FMT")
+      set(spdlog_fmt_target fmt::fmt)
+    elseif(_RAPIDS_FMT_OPTION STREQUAL "EXTERNAL_FMT_HO")
+      set(spdlog_fmt_target fmt::fmt-header-only)
+    endif()
+  elseif(_RAPIDS_FMT_OPTION STREQUAL "STD_FORMAT")
+    set(spdlog_fmt_option "SPDLOG_USE_${_RAPIDS_FMT_OPTION} ON")
+  else()
+    message(FATAL_ERROR "Invalid option used for FMT_OPTION, got: ${_RAPIDS_FMT_OPTION}, expected one of: 'BUNDLED', 'EXTERNAL_FMT', 'EXTERNAL_FMT_HO'"
```
bdice, line 122:
```suggestion
    message(FATAL_ERROR "Invalid option used for FMT_OPTION, got: ${_RAPIDS_FMT_OPTION}, expected one of: 'BUNDLED', 'EXTERNAL_FMT', 'EXTERNAL_FMT_HO', 'STD_FORMAT'"
```

rapids-cmake/cpm/spdlog.cmake
```diff
@@ -29,10 +29,26 @@ across all RAPIDS projects.
...
+.. versionadded:: v23.04.00
+
+  Spdlog depends on the fmt library and offers multiple ways of handling this dependency when spdlog is built. This
+  option only controls the behavior when spdlog is fetched and built, NOT when an installed spdlog is found on the
+  system.
+
+  This option can be set to: `BUNDLED`, `EXTERNAL_FMT`, `EXTERNAL_FMT_HO`, or `STD_FORMAT`. If set to
+  `BUNDLED`, then spdlog will use its own bundled version of fmt. If set to `EXTERNAL_FMT` then spdlog will use the
+  `fmt::fmt` target and be linked with the fmt library. If set to `EXTERNAL_FMT_HO` then spdlog will use the
+  `fmt::fmt-header-only` target and be linked with a header only fmt library. If set to `STD_FORMAT` then spdlog
```
bdice, line 47: “linked with a header only fmt library”

Is that the right wording? Can you link to a header-only library?

bdice (changes requested):
Requesting changes for now with two small comments. I am going to wait to merge this until I can babysit the RMM PR through to completion, to minimize potential downtime / downstream breakage. Probably next Monday.

rapids-cmake/cpm/spdlog.cmake
robertmaynard: From a CMake perspective you can. You use `target_link_libraries` on interface targets ( header only ) to get any include directories, defintions, etc that are need to use that library.

kkraus14: In my brain I was thinking of the cmake `target_link_libraries` call which uses the `fmt::fmt-header-only` target. 🤷

bdice: Thanks for the clarification!

kkraus14:
This doesn't seem to be working properly in the case where spdlog and fmt aren't installed already locally when trying to build RMM:

```
CMake Error in CMakeLists.txt:
  export called with target "rmm" which requires target "fmt-header-only"
  that is not in this export set, but in multiple other export sets:
  /home/keith/git/rmm/build/_deps/fmt-build/fmt-targets.cmake,
  /home/keith/git/rmm/build/fmt-targets.cmake.

  An exported target cannot depend upon another target which is exported
  multiple times.  Consider consolidating the exports of the
  "fmt-header-only" target to a single export.
 ```
 
I think this is because we're depending on both `spdlog` and `fmt` and we transitively export `fmt` when exporting `spdlog` which causes `fmt` to be double exported? I would assume the solution to this is to pass on the export arguments into the `rapids_cpm_fmt` call in the `rapids_cpm_spdlog` call conditionally, but I'm not quite sure. @robertmaynard maybe you have an idea?

robertmaynard:
> I think this is because we're depending on both `spdlog` and `fmt` and we transitively export `fmt` when exporting `spdlog` which causes `fmt` to be double exported? I would assume the solution to this is to pass on the export arguments into the `rapids_cpm_fmt` call in the `rapids_cpm_spdlog` call conditionally, but I'm not quite sure. @robertmaynard maybe you have an idea?

We are doing multiple exports of the `fmt` package since it is also doing a call to `export(`. The correct approach it to remove our export in `get_fmt` ( https://github.com/rapidsai/rmm/pull/1177/files#r1104528086 ).

PR #370, opened 2023-02-10
Use trap to handle errors in test scripts

AjayThorve:
This PR adds a less verbose [trap method](https://github.com/rapidsai/cugraph/blob/f2b081075704aabc789603e14ce552eac3fbe692/ci/test.sh#L19), for error handling to help ensure that we capture all potential error codes in our test scripts, and works as follows:

- setting an environment variable, EXITCODE, with a default value of 0
- setting a trap statement triggered by ERR signals which will set EXITCODE=1 when any commands return a non-zero exit code

cc @ajschmidt8

### DIFF
--- a/ci/test_cpp.sh
+++ b/ci/test_cpp.sh
@@ -31,6 +31,7 @@ cd build
 EXITCODE=0
 trap "EXITCODE=1" ERR
 set +e
+
 ctest --schedule-random --output-on-failure
 
 rapids-logger "Test script exiting with value: $EXITCODE"
--- a/ci/test_cpp.sh
+++ b/ci/test_cpp.sh
@@ -23,11 +23,15 @@ rapids-print-env
 rapids-logger "Check GPU usage"
 nvidia-smi
 
+EXITCODE=0
+trap "EXITCODE=1" ERR
+set +e
+
 rapids-logger "Begin cpp tests"
 cmake -S testing -B build
 
 cd build
 ctest --schedule-random --output-on-failure
-exitcode=$?
 
-exit ${exitcode}
+rapids-logger "Test script exiting with value: $EXITCODE"
+exit ${EXITCODE}
--- a/ci/test_cpp.sh
+++ b/ci/test_cpp.sh
@@ -23,14 +23,14 @@ rapids-print-env
 rapids-logger "Check GPU usage"
 nvidia-smi
 
-EXITCODE=0
-trap "EXITCODE=1" ERR
-set +e
-
 rapids-logger "Begin cpp tests"
 cmake -S testing -B build
 
 cd build
+
+EXITCODE=0
+trap "EXITCODE=1" ERR
+set +e
 ctest --schedule-random --output-on-failure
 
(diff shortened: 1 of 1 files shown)


ci/test_cpp.sh
```diff
@@ -23,11 +23,15 @@ rapids-print-env
...
 
+EXITCODE=0
+trap "EXITCODE=1" ERR
+set +e
+
 rapids-logger "Begin cpp tests"
 cmake -S testing -B build
 
 cd build
 ctest --schedule-random --output-on-failure
```
ajschmidt8, line 34:
```suggestion
rapids-logger "Begin cpp tests"
cmake -S testing -B build

cd build

EXITCODE=0
trap "EXITCODE=1" ERR
set +e
ctest --schedule-random --output-on-failure
```

I think we should move this to just above the `ctest` invocation.

The other parts should fail immediately if there's an error.

ajschmidt8 (changes requested):

robertmaynard (commented):
I am confused.

The PR title is `Reduce error handling verbosity in CI tests scripts` but the scripts are now longer. What verbosity is this intending to reduce?

ajschmidt8:
> I am confused.
> 
> The PR title is `Reduce error handling verbosity in CI tests scripts` but the scripts are now longer. What verbosity is this intending to reduce?

Good point. This change doesn't have much effect on `rapids-cmake` since it's test scripts are pretty succinct.

In other repositories, these changes were meant to replace the many conditional code blocks that existed previously for capturing error codes (like [this file](https://github.com/rapidsai/cudf/pull/12738/files#diff-12c1ee18bba80b27398d5173148eaba86ea281dea120e3f055989287cbd7763c) for cudf).

@robertmaynard, if you want us to close this PR since `rapids-cmake` has a small test script that's fine. I initially just created a list of all the GHAs repositories that should be checked which is why this PR was opened.

AjayThorve:
> I am confused.
> 
> The PR title is `Reduce error handling verbosity in CI tests scripts` but the scripts are now longer. What verbosity is this intending to reduce?

Hey @robertmaynard, you are right. This PR was part of a larger effort across all repos, and the title applied to all the repos except rapids-cmake. Infact in this case, as you mentioned it's adding more lines. Apologies for the confusion, but the aim of the change is to use the `trap` method to catch ERR, and while in most repositories across rapidsai, it significantly reduced the verbosity, in the case of rapids-cmake, we can may be use `trap` to keep error handling consistent across projects?

robertmaynard:
Absolutely fine moving over to using trap. Let's just update the pr title

ci/test_cpp.sh
```diff
@@ -27,7 +27,11 @@ rapids-logger "Begin cpp tests"
 cmake -S testing -B build
 
 cd build
+
+EXITCODE=0
+trap "EXITCODE=1" ERR
+set +e
```
bdice, line 33: Add newline before calling `ctest`.
```suggestion
set +e

```

bdice (approved):

ajschmidt8 (approved):

PR #372, opened 2023-02-14
Support static builds of fmt

robertmaynard:
## Description
By default `fmt` doesn't set the `POSITION_INDEPENDENT_CODE` value on targets when built statically. This causes link errors when the static library is consumed by a shared library.

## Checklist

### DIFF
--- a/rapids-cmake/cpm/fmt.cmake
+++ b/rapids-cmake/cpm/fmt.cmake
@@ -71,7 +71,7 @@ function(rapids_cpm_fmt)
                   GIT_SHALLOW ${shallow}
                   PATCH_COMMAND ${patch_command}
                   EXCLUDE_FROM_ALL ${exclude}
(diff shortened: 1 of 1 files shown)


robertmaynard:
Need to add a test that uses fmt statically.

testing/cpm/CMakeLists.txt
```diff
@@ -48,6 +48,8 @@ add_cmake_config_test( cpm_cuco-libcudacxx-no-install-export.cmake )
 
 add_cmake_config_test( cpm_fmt-export.cmake )
 add_cmake_config_test( cpm_fmt-simple.cmake )
+add_cmake_build_test( cpm_fmt-static-lib.cmake )
+
```
bdice, line 52: Extra newline not needed.
```suggestion
```

bdice (approved):

PR #375, opened 2023-02-17
rapids_cpm_spdlog properly specify usage of external fmt library

robertmaynard:
## Description
Previously we specified the wrong options for spdlog to use an the external fmt library. This caused spdlog to use the internal version while we had other non-spdlog components uses the upstream fmt.

## Checklist

### DIFF
--- a/rapids-cmake/cpm/spdlog.cmake
+++ b/rapids-cmake/cpm/spdlog.cmake
@@ -132,7 +132,7 @@ function(rapids_cpm_spdlog)
                   GIT_SHALLOW ${shallow}
                   PATCH_COMMAND ${patch_command}
                   EXCLUDE_FROM_ALL ${exclude}
-                  OPTIONS "SPDLOG_INSTALL ${to_install}" ${spdlog_fmt_option})
+                  OPTIONS "SPDLOG_INSTALL ${to_install}" "${spdlog_fmt_option}")
 
   include("${rapids-cmake-dir}/cpm/detail/display_patch_status.cmake")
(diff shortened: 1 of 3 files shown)


rapids-cmake/cpm/spdlog.cmake
```diff
@@ -100,27 +100,27 @@ function(rapids_cpm_spdlog)
...
+  if(_RAPIDS_FMT_OPTION STREQUAL "EXTERNAL_FMT" OR _RAPIDS_FMT_OPTION STREQUAL "EXTERNAL_FMT_HO")
     include("${rapids-cmake-dir}/cpm/fmt.cmake")
 
     # Using `spdlog_ROOT` needs to cause any internal find calls in `spdlog-config.cmake` to first
     # search beside it before looking globally.
     list(APPEND fmt_ROOT ${spdlog_ROOT})
 
     rapids_cpm_fmt(${_RAPIDS_UNPARSED_ARGUMENTS})
-
-    set(spdlog_fmt_option "SPDLOG_${_RAPIDS_FMT_OPTION} ON")
```
vyasr, line 113: Should we consider changing the valid rapids-cmake options so that they match the built-in spdlog options, i.e. make this set command work?

robertmaynard: I don't want to use the spdlog option as they aren't consistent ( `SPDLOG_USE_STD_FORMAT` vs `SPDLOG_FMT_EXTERNAL[_HO]` ) and am willing to have some complexity here to make a nicer user interface.

vyasr (approved):

PR #376, opened 2023-02-17
rapids_cpm_nvbench properly specify usage of external fmt library

robertmaynard:
## Description
When we are inside a conda env the linker will be set to ld.bfd which will try to resolve all undefined symbols at link time.

Since we could be using a shared library version of fmt we need it on the final link line of consumers. So patch nvbench to understand this requirement.

## Checklist

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -45,6 +45,7 @@ dependencies:
               cuda: "11.4"
             packages:
               - cudatoolkit=11.4
+              - gcc<11.0.0
           - matrix:
               cuda: "11.5"
             packages:
--- a/rapids-cmake/cpm/patches/nvbench/public_fmt_dep_in_conda.diff
+++ b/rapids-cmake/cpm/patches/nvbench/public_fmt_dep_in_conda.diff
@@ -1,0 +1,31 @@
+diff --git a/nvbench/CMakeLists.txt b/nvbench/CMakeLists.txt
+index f86bd41..ba6418f 100644
+--- a/nvbench/CMakeLists.txt
++++ b/nvbench/CMakeLists.txt
+@@ -76,10 +76,25 @@ target_link_libraries(nvbench
+   PUBLIC
+     ${ctk_libraries}
+   PRIVATE
+-    fmt::fmt
+     nvbench_json
+     nvbench_git_revision
+ )
++
++# ##################################################################################################
++# * conda environment -----------------------------------------------------------------------------
++rapids_cmake_support_conda_env(conda_env MODIFY_PREFIX_PATH)
++if(TARGET conda_env)
++  # When we are inside a conda env the linker will be set to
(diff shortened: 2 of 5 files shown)


vyasr:
I'm afraid I don't understand this change. Won't any linker require that symbols be resolved at link-time (and if linking dynamically, _again_ at runtime)? IIUC the change in this PR is propagating the fmt requirement from nvbench to its consumers, but I don't understand how the resulting change in terms of generated compile/link commands is related to the choice of linker.

robertmaynard:
> I'm afraid I don't understand this change. Won't any linker require that symbols be resolved at link-time (and if linking dynamically, the requirement will be enforced again at runtime by the loader)? IIUC the change in this PR is propagating the fmt requirement from nvbench to its consumers, but I don't understand how the resulting change in terms of generated compile/link commands is related to the choice of linker.


Good question. This PR is resolving an interacation between hidden symbols, undefined symbols and the conda linker ld.bfd.

When compiling things like the libcudf benchmarks we run into the following scenario:
  1. Usage of rmm which requests using the fmt header only library
  2. Usage of nvbench which has used the fmt via shared library 

This results in the final executable having a link error like the following:

```bash
/x86_64-conda-linux-gnu/bin/ld: uses_fmt: hidden symbol `_ZN3fmt2v96detail18throw_format_errorEPKc' in CMakeFiles/uses_fmt.dir/use_fmt.cpp.o is referenced by DSO
```

Since we compiled nvbench against the shared library version of fmt, it has recorded that ```_ZN3fmt2v96detail18throw_format_errorEPKc``` is
an undefined symbol. When we compile the executable translation unit it records that ```_ZN3fmt2v96detail18throw_format_errorEPKc``` is weak and PRIVATE,
due to conda injecting `-fvisibility-inlines-hidden` on the compilation line.



So when we go to link the final executable the linker notices that we have inconsistent definitions for this symbol. One library says it is PRIVATE and the other says it is PUBLIC and undefined. Since `ld.bfd` behavior is to reject when we have inconsistent specifications it errors out.

To fix this I have identified two solutions:

1. Ensure that any translation unit compiled under conda which ingests FMT via both shared lib and header has `-fno-visibility-inlines-hidden` 
on the compile line.
2. Ensure that any translation unit compiled under conda which ingests FMT via both shared lib and header has `-DFMT_SHARED` on the compile line


I went with option 2 as it has the smallest impact on other functions / libraries being used. So to implement #2 I made the `nvbench` dependency on `fmt` public so we propagate the `FMT_SHARED` flag.

vyasr:
Got it, that makes sense. What happens in the case where a TU links to both rmm (which links to the `fmt::fmt-header-only` target) and the patched nvbench (which links to `fmt::fmt`)? Will the two not end up conflicting?

vyasr (approved):

PR #378, opened 2023-02-23
Introduce clang-format to rapids-cmake to format C++ code examples

robertmaynard:
## Description
In preparation of more C++ code being added to rapids-cmake introduce a .clang-format file to the project to keep everything with a consistent style and formatting.

## Checklist

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,155 @@
+---
+# Refer to the following link for the explanation of each params:
+#   http://releases.llvm.org/8.0.0/tools/clang/docs/ClangFormatStyleOptions.html
+Language: Cpp
+# BasedOnStyle: Google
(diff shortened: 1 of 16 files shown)


vyasr (commented):
How did you choose the settings? Are they copied from another project?

I was going to recommend adding a pre-commit hook, but it looks like we've already got one. Have we been running it without a format file?

robertmaynard:
I copied the format file from cudf. 

Yeah we have been running format without any file so it was using the defaults. But I don't think we ever ran it over the existing c++ files

bdice (approved):

bdice:
Thanks! This simplifies #328.

vyasr (approved):

PR #384, opened 2023-03-07
support_conda_env only add rpath-link flag to linkers that support it.

### LINKED ISSUE #383: [BUG] rapids_cmake_support_conda_env adds -rpath-link flag even for Mac builds
**Describe the bug**
`rapids_cmake_support_conda_env` now adds `-rpath-link` for all platforms, but this flag is not valid for the Mac linker.

Build a library in a Conda environment and call `rapids_cmake_support_conda_env` on that library.

**Expected behavior**
`-rpath-link` should not be added when the platform is Mac? at least not for the default Mac linker

**Environment details (please complete the following information):**
CMake 3.24, Mac OS 12.2, default Mac linker

robertmaynard:
## Description
Fixes #383

## Checklist


trxcllnt (changes requested):
Doesn't this also need to be in the `elseif(in_conda_prefix)` block too?

trxcllnt (approved):

PR #388, opened 2023-03-14
rapids_test_install_relocatable EXCLUDE_FROM_ALL is now the default

robertmaynard:
## Description
Provides better default behavior to `rapids_test_install_relocatable` so that projects don't install test files all the time.

## Checklist

### DIFF
--- a/cmake-format-rapids-cmake.json
+++ b/cmake-format-rapids-cmake.json
@@ -346,7 +346,7 @@
       "rapids_test_install_relocatable": {
         "pargs": {
           "nargs": "0",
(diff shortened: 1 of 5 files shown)


robertmaynard:
> Is it really worth having both options? I presume that you did so for backwards compatibility, but since nothing is using this functionality yet perhaps it would be simpler to remove `EXCLUDE_FROM_ALL` entirely and just update the rmm/cudf PRs accordingly.

You are correct it was out of concern that `EXCLUDE_FROM_ALL` was released. But since nobody is using the feature yet lets yank the option now.

vyasr (approved):
LGTM

PR #393, opened 2023-03-27
rapids_cpm cccl packages cmake files are now relocated to not clash with upstream

### LINKED ISSUE #367: [BUG] Thrust (and cub) cmake files aren't installed in a rapids specific location
**Describe the bug**
Currently we install Thrust (and cub) headers into a RAPIDS specific location in `include/rapids`: https://github.com/rapidsai/rapids-cmake/blob/branch-23.04/rapids-cmake/cpm/thrust.cmake#L64

We don't do the same for the cmake files, so we're effectively hijacking `find_package(Thrust)` calls to point towards our RAPIDS installation when we probably shouldn't be.

Run a build + install of librmm, where the cmake files for Thrust (and cub) get put into `lib/cmake/thrust` (and `lib/cmake/cub`).

**Expected behavior**

I assume we should put the cmake files in a `rapids` subdirectory of sorts and then handle adding that subdirectory to the cmake module / prefix path as part of the rapids-cmake initialization or something of the sort?

robertmaynard:
## Description
This way rapids-cmake versions of these projects don't clash with the upstream versions.

Fixes #367

## Checklist


rapids-cmake/cpm/libcudacxx.cmake
```diff
@@ -103,14 +103,18 @@ function(rapids_cpm_libcudacxx)
...
     include(GNUInstallDirs)
     set(CMAKE_INSTALL_INCLUDEDIR "${CMAKE_INSTALL_INCLUDEDIR}/rapids/libcudacxx")
+    set(CMAKE_INSTALL_LIBDIR "${CMAKE_INSTALL_LIBDIR}/rapids/")
+
+    include("${rapids-cmake-dir}/export/find_package_root.cmake")
+    rapids_export_find_package_root(INSTALL libcudacxx
+                                    [=[${CMAKE_CURRENT_LIST_DIR}/../../rapids/cmake/libcudacxx]=]
+                                    ${_RAPIDS_INSTALL_EXPORT_SET})
 
     # libcudacxx 1.8 has a bug where it doesn't generate proper exclude rules for the
```
bdice, line 113: Can we remove this whole block when we bump to the latest libcudacxx? I want to know since I'll be updating CCCL soon.

bdice (approved):
Looks good to me. Should we add tests?

rapids-cmake/cpm/libcudacxx.cmake
robertmaynard: I have removed all of the patching here. We will still need to patch libcudacxx ( even 2.0+ ) but the match set will be fairly minimal. The libcudacxx header search wasn't set up like thrust/cub and doesn't work with custom CMAKE_INSTALL_INCLUDEDIR values.

bdice (approved):

rapids-cmake/cpm/libcudacxx.cmake
bdice: Okay, thanks! I approved the PR.

vyasr:
I do plan to review this PR, just haven't gotten around to it quite yet.

bdice:
@vyasr Would you be able to review this PR? I'd like to have this merged and then update the CCCL versions in https://github.com/rapidsai/rapids-cmake/pull/399.

rapids-cmake/cpm/patches/libcudacxx/install_rules.diff
```diff
@@ -0,0 +1,66 @@
...
+ install(DIRECTORY "${libcudacxx_SOURCE_DIR}/lib/cmake/libcudacxx"
+   DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake"
+-  PATTERN libcudacxx-header-search EXCLUDE
++  PATTERN *.cmake.in EXCLUDE
+ )
+ 
+ # Need to configure a file to store CMAKE_INSTALL_INCLUDEDIR
+ # since it can be defined by the user. This is common to work around collisions
+ # with the CTK installed headers.
++set(install_location "${CMAKE_INSTALL_LIBDIR}/cmake/libcudacxx")
```
vyasr, line 33: Why was it important to have this variable set before the below install commands rather than hardcoding it in the install?

rapids-cmake/cpm/patches/libcudacxx/install_rules.diff
```diff
@@ -0,0 +1,66 @@
...
+ )
+ install(DIRECTORY "${libcudacxx_SOURCE_DIR}/include/nv"
+   DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
++  PATTERN CMakeLists.txt EXCLUDE
+ )
+ 
+ # Libcudacxx cmake package
+ install(DIRECTORY "${libcudacxx_SOURCE_DIR}/lib/cmake/libcudacxx"
+   DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake"
+-  PATTERN libcudacxx-header-search EXCLUDE
```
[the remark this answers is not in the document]
vyasr: Ah I see that file is also a *.cmake.in so the new pattern is more general and will capture it in addition to other files.

rapids-cmake/cpm/thrust.cmake
```diff
@@ -56,17 +56,18 @@ Result Variables
 function(rapids_cpm_thrust NAMESPACE namespaces_name)
   list(APPEND CMAKE_MESSAGE_CONTEXT "rapids.cpm.thrust")
 
+  include("${rapids-cmake-dir}/cpm/detail/package_details.cmake")
+  rapids_cpm_package_details(Thrust version repository tag shallow exclude)
```
vyasr, line 60: Kinda confused as to why this moved, this information isn't being used anywhere new is it? The new lines outside of the conditional scope don't seem to use them, what am I missing?

vyasr (commented):
I would recommend changing the title to more clearly indicate that what this PR is doing is relocating where rapids-cmake installs CCCL packages. Vendor sounds like the files are embedded in some way, and tbh I was expecting this PR to be about creating a rapids-cmake conda package containing the rapids-specific CCCL dependencies before I read through it.

rapids-cmake/cpm/thrust.cmake
robertmaynard: Nice eye. The issue is that it should be used in the logic to compute `to_install` and wasn't.
Update the PR to correct this bug.

rapids-cmake/cpm/libcudacxx.cmake
```diff
@@ -54,6 +54,32 @@ function(rapids_cpm_libcudacxx)
   include("${rapids-cmake-dir}/cpm/detail/package_details.cmake")
   rapids_cpm_package_details(libcudacxx version repository tag shallow exclude)
 
+  set(to_install OFF)
+  if(INSTALL_EXPORT_SET IN_LIST ARGN AND NOT exclude)
```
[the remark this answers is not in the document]
robertmaynard: Parsing before hand comes with some complexity as you need to re-construct the `BUILD_EXPORT_SET` and `INSTALL_EXPORT_SET` command so they propagate to `rapids_cpm_find` so I try to avoid it as much as possbile.

rapids-cmake/cpm/libcudacxx.cmake
```diff
@@ -82,41 +109,11 @@ function(rapids_cpm_libcudacxx)
...
-      file(READ "${libcudacxx_SOURCE_DIR}/cmake/libcudacxxInstallRules.cmake" contents)
-      string(REPLACE "PATTERN cub-header-search EXCLUDE" "REGEX cub-header-search.* EXCLUDE"
-                     contents "${contents}")
-      string(REPLACE "PATTERN libcudacxx-header-search EXCLUDE"
-                     "REGEX libcudacxx-header-search.* EXCLUDE" contents "${contents}")
-      file(WRITE "${libcudacxx_BINARY_DIR}/cmake/libcudacxxInstallRulesForRapids.cmake" ${contents})
-    endif()
-    set(libcudacxx_ENABLE_INSTALL_RULES ON)
-    include("${libcudacxx_BINARY_DIR}/cmake/libcudacxxInstallRulesForRapids.cmake")
+  if(to_install)
```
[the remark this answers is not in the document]
robertmaynard: You are correct. Re-added the check for `libcudacxx_SOURCE_DIR`

rapids-cmake/cpm/patches/libcudacxx/install_rules.diff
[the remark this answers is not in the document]
robertmaynard: The old pattern doesn't work at all since `PATTERN` only matches at the end of names.

rapids-cmake/cpm/patches/libcudacxx/install_rules.diff
robertmaynard: The variable is also expanded by the `configure_file` of `libcudacxx-header-search.cmake"`.

rapids-cmake/cpm/libcudacxx.cmake
```diff
@@ -54,6 +54,32 @@ function(rapids_cpm_libcudacxx)
...
+    # CMake's default of system include. Compilers when provided the same include as both user and
+    # system always goes with system.
+
+    # Now while rmm could also mark `install/include` as system this just pushes the issue to
+    # another dependency which isn't built by RAPIDS and comes by and marks `install/include` as
+    # system.
+
+    # Instead the more reliable option is to make sure that we get libcudacxx to be placed in an
+    # unique include path that the other project will use. In the case of rapids-cmake we install
+    # the headers to `include/rapids/libcudacxx`
```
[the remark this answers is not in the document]
robertmaynard: removed `install/`

vyasr (approved):

rapids-cmake/cpm/libcudacxx.cmake
```diff
@@ -54,6 +54,32 @@ function(rapids_cpm_libcudacxx)
...
+
+    # This is a problem for CMake+NVCC due to the rules around import targets, and user/system
+    # includes. In this case both rmm and libcudacxx will specify an include path of
+    # `include`, while libcudacxx tries to mark it as an user include, since rmm uses
+    # CMake's default of system include. Compilers when provided the same include as both user and
+    # system always goes with system.
+
+    # Now while rmm could also mark `include` as system this just pushes the issue to
+    # another dependency which isn't built by RAPIDS and comes by and marks `include` as
+    # system.
```
bdice, line 73: Fix style for CI:
```suggestion
    # includes. In this case both rmm and libcudacxx will specify an include path of `include`,
    # while libcudacxx tries to mark it as an user include, since rmm uses CMake's default of system
    # include. Compilers when provided the same include as both user and system always goes with
    # system.

    # Now while rmm could also mark `include` as system this just pushes the issue to another
    # dependency which isn't built by RAPIDS and comes by and marks `include` as system.
```

vyasr:
@robertmaynard can we merge this now? Then we can also finalize #13235.

PR #396, opened 2023-03-30
Revert "Define Cython language_level explicitly. (#394)"

vyasr:
## Description

Current builds of RAPIDS are not compatible with this language level setting yet.

## Checklist

### DIFF
--- a/rapids-cmake/cython/init.cmake
+++ b/rapids-cmake/cython/init.cmake
@@ -1,5 +1,5 @@
 # =============================================================================
-# Copyright (c) 2022, NVIDIA CORPORATION.
+# Copyright (c) 2022-2023, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
 # in compliance with the License. You may obtain a copy of the License at


vyasr:
I suspect that setting a value of 3 will also work, but I would want to test that locally on a few repos before pushing a change like this during code freeze.

robertmaynard:
> I suspect that setting a value of 3 will also work, but I would want to test that locally on a few repos before pushing a change like this during code freeze.

This and the original PR target 23.06 so we shouldn't need to worry about code freeze.

vyasr:
Ah yes of course, you are right. I am experimenting with the new setting locally and should at least be able to validate a few repos today. That said, I would still advocate reverting the change for now and then making the fix in a subsequent PR because the current state of rapids-cmake blocks development of other repos on 23.06. Once I have the time to do enough local builds I can make a follow-up, I just don't know that I can promise that today.

bdice:
Thanks @vyasr for picking this up. We should use `3str` and not `3`, but only because `3str` is the default language level in Cython and we have no reason to differ from the default. This setting is only meant to squash warnings about the fact that the default value has changed in Cython 3 (the warnings go away when setting it manually).

bdice (approved):

vyasr:
> Thanks @vyasr for picking this up. We should use `3str` and not `3`, but only because `3str` is the default language level in Cython and we have no reason to differ from the default. This setting is only meant to squash warnings about the fact that the default value has changed in Cython 3 (the warnings go away when setting it manually).

Is there any reason not to try switching to 3? We shouldn't be relying on Python 2 string semantics in any way.

vyasr:
Based on the Cython documentation 3str and 3 should be identical when running on Python 3 instead of Python 2. I suspect that the current issue is a Cython bug in 0.29.x that is fixed in Cython 3.

vyasr:
I figured out why this was causing issues. We are in fact already using `3`, not `3str`, because [that is the scikit-build default](https://github.com/scikit-build/scikit-build/blob/master/skbuild/resources/cmake/UseCython.cmake#L186). Since that's being provided on the command line, it's overriding the Cython default. The cudf builds failed because as I mentioned before we are already doing things that are incompatible with Python 2 strings, namely using `object` types inside nogil blocks where they must be contextually convertible to C/C++ types. The lack of consistency in string types in Python 2 makes code [like this](https://github.com/rapidsai/cudf/blob/branch-23.08/python/cudf/cudf/_lib/datetime.pyx#L420) invalid with Python 2 strings.

Cython defaults to `3str`, I assume, because strings are the single largest change from Python 2 to Python 3, and they didn't want to break backward compatibility that extensively because Cython is so widely used. We should be comfortable moving to `3`.

PR #399, opened 2023-04-14
Update CCCL to 2.1.0.

bdice:
## Description
This PR updates rapids-cmake to use CCCL version 2.1.0 (thrust, cub, libcudacxx). I will test this downstream with other repositories before opening this draft for review.

## Checklist


bdice:
Downstream testing PRs (will edit this comment as I test more repos):
- https://github.com/rapidsai/rmm/pull/1247

RAPIDS.cmake
```diff
@@ -1,5 +1,5 @@
 #=============================================================================
```
robertmaynard, line 1: You shouldn't need to modify the `RAPIDS.cmake` to test your PR. The testing infrastructure will do this automatically since we don't need to fetch rapids-cmake.

For downstream PR's to verify these changes what you can do is:
```
if(NOT EXISTS ${CMAKE_CURRENT_BINARY_DIR}/RMM_RAPIDS.cmake)
  set(rapids-cmake-repo bdice/rapids-cmake)
  set(rapids-cmake-branch cccl-update-2.1.0)
  file(DOWNLOAD https://raw.githubusercontent.com/rapidsai/rapids-cmake/branch-23.06/RAPIDS.cmake
endif()
include(${CMAKE_CURRENT_BINARY_DIR}/RMM_RAPIDS.cmake)
```
That will cause rmm to use the `23.06` version of the `RAPIDS.cmake` file, but that file will download your custom fork/branch of rapids-cmake ( weird but works ).

bdice: I looked at this and recognized that your suggestion was the intended path — but I was really eager to have a simple sed replacement for the downstream repo tests. I saw that this method of testing (modifying the upstream repo default) let me get by with fewer lines of downstream changes. As a result, I can easily use r3 (an automation tool from @ajschmidt8) to open up testing PR en masse for all repos. Is there anything that could go wrong with my approach? I would revert my changes in this PR before merging, of course.

robertmaynard: The change in `rmm` doesn't work on local builds when the build directory already exists. The `RMM_RAPIDS.cmake` already exists in the build directory so the new version will never be downloaded.

bdice: I see. For now, I'm only testing in CI, so I won't hit that case. I can try to build a more complex sed command to do this the other way.

bdice: I found a better sed command so I'll be using this for subsequent testing PRs.

```bash
find . -name "CMakeLists.txt" -type f -exec sed -i 's/\(^[[:space:]]*\)\(.*branch-23.06\/RAPIDS\.cmake.*\)/\1set(rapids-cmake-repo bdice\/rapids-cmake)\n\1set(rapids-cmake-branch cccl-update-2.1.0)\n\1\2/g' {} \;
```

bdice:
I have now updated libcudacxx and it appears that builds/tests are passing for rmm. I verified in the logs that Thrust/CUB/libcudacxx versions 2.1.0 are being used.

I will open testing PRs for other RAPIDS repositories.

(cross-posted from https://github.com/rapidsai/rmm/pull/1247#issuecomment-1522731678)

bdice:
Several packages (cuDF, cuML, cuGraph) are complaining about problems with device lambdas. I will need to investigate this further. We'll likely need to write proper functors/callable structs for these lambdas.

ajschmidt8 (dismissed):

EDIT---

whoops, didn't realize this was still draft. I'll probably have to re-approve when it's ready, but that's fine.

bdice:
Update: there was a minor issue with `cuda::proclaim_return_type` which is fixed in https://github.com/NVIDIA/libcudacxx/pull/448. I have a smaller patch in this PR that I'll update once that upstream PR is merged. We _could_ use an untagged commit hash instead of a patch once that is merged... not sure which approach we tend to prefer in `versions.json`.

rapids-cmake/rapids-version.cmake
```diff
@@ -1,5 +1,5 @@
 #=============================================================================
-# Copyright (c) 2021, NVIDIA CORPORATION.
+# Copyright (c) 2021-2023, NVIDIA CORPORATION.
```
bdice, line 2:
```suggestion
# Copyright (c) 2021, NVIDIA CORPORATION.
```

bdice:
I rebased and force-pushed this PR so that all commits are signed. Now, future changes should not require `/ok to test` for CI to run.

bdice:
Closing this, now that #495 is merged. We will leave `rapids_cpm_thrust` and `rapids_cpm_libcudacxx` in place pinned to their current versions, and eventually deprecate them in favor of `rapids_cpm_cccl`.

PR #403, opened 2023-04-20
Use @loader_path instead of $ORIGIN on MacOS

manopapad:
## Description

The dynamic libraries currently produced by rapids-cmake on MacOS for cython modules don't work, because the wrong "ORIGIN" marker is used. The typical suggestion appears to be to use `@loader_path` in place of `$ORIGIN` on MacOS, although there is one more option that I can't currently recall.

Please excuse my ignorance around your processes. Please feel free to make/request changes.

## Checklist
- [X] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/rapids-cmake/blob/HEAD/CONTRIBUTING.md).


vyasr:
That solution is consistent with what I've seen used on Mac before, but I'll defer to @robertmaynard since he has far more experience than I do.

rapids-cmake/cython/add_rpath_entries.cmake
```diff
@@ -72,14 +72,19 @@ function(rapids_cython_add_rpath_entries)
     list(APPEND cleaned_paths "${path}")
   endforeach()
 
+  if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
+    set(platform_rpath_origin "\$ORIGIN")
```
bdice, line 76: Is this supposed to be escaped? (`$ORIGIN` vs. `\$ORIGIN`) It isn't escaped here in the original code, but it is escaped in the `install_rpath` in the other file. I'm not 100% sure how variable interpolation works in CMake so escaping may be needed.

rapids-cmake/cython/add_rpath_entries.cmake
```diff
@@ -72,14 +72,19 @@ function(rapids_cython_add_rpath_entries)
     list(APPEND cleaned_paths "${path}")
   endforeach()
 
+  if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
```
bdice, line 75: Could we make this `@loader_path` behavior specific to Darwin and use `$ORIGIN` in an "else"? I'd like to avoid the string check for `"Linux"` as a defensive behavior.

bdice (commented):
Like @vyasr, I think this looks generally right but I would defer to @robertmaynard.

rapids-cmake/cython/add_rpath_entries.cmake
robertmaynard: I agree, we should explicit check for Darwin and otherwise use $ORIGIN

rapids-cmake/cython/add_rpath_entries.cmake
robertmaynard: No need to escape $ORIGIN with CMake 3.16+

manopapad:
Comments addressed, PTAL

manopapad:
@bdice is there anything else to do on this one? I updated against latest 23.06, but I am unable to set labels, as required by the "Label Checker" action.

robertmaynard:
> @bdice is there anything else to do on this one? I updated against latest 23.06, but I am unable to set labels, as required by the "Label Checker" action.

Added labels

bdice (approved):
All looks fine to me. Are we good to merge @robertmaynard @vyasr ?

bdice:
I bumped this to 23.08 since 23.06 is in code freeze.

Issue #427, opened 2023-06-09
[FEA] Implement a function to link a target appropriately to the CUDA runtime


vyasr:
**Is your feature request related to a problem? Please describe.**
rapids-cmake offers `rapids_cuda_init_runtime`, which implicitly changes the (default) CUDA linking behavior for all subsequent targets. This implicit behavior makes it difficult for consumers to reason about the CUDA linking behavior of targets since the resulting code is order-dependent and may be overridden by settings on particular targets.

**Describe the solution you'd like**
We should consider implementing a new function that accepts a target and adds a linkage to the appropriate CUDA library (dynamic or static) based on a function parameter. We could implement this either via a direct `target_link_libraries` call under the hood, or by creating a helper target (e.g. `add_library(rapids_cmake_cuda_lib...)`) that encodes the linkage, similar to how we handle conda in `rapids_cmake_conda_env`. If we wanted to incorporate math libraries we could include those as well, but we need to ensure that those are not constrained to the same type as the CUDA runtime (since the most common long-term behavior we want to encourage is statically linking cudart but dynamically linking the math libraries). We may instead benefit more from creating a separate function to handle the math library linkage.

robertmaynard:
This does open the question if we also need to offer controls over the usage behavior of the linking.

So do we actually need
```cmake
add_library(uses_cuda SHARED <.....>)

rapids_cuda_set_runtime(uses_cuda PUBLIC USE_STATIC_RUNTIME ${PROJ_CUDA_STATIC_RUNTIME})
```

PR #429, opened 2023-06-13
Add rapids_cuda_set_runtime

robertmaynard:
## Description
Add `rapids_cuda_set_runtime` which is the per target version of `rapids_cuda_init_runtime`.
Fixes https://github.com/rapidsai/rapids-cmake/issues/427

## Checklist


rapids-cmake/cuda/set_runtime.cmake
```diff
@@ -0,0 +1,65 @@
...
+.. versionadded:: v23.08.00
+
+Establish what CUDA runtime library should be used by a single target
+
+  .. code-block:: cmake
+
+    rapids_cuda_set_runtime( target USE_STATIC (TRUE|FALSE) )
+
+  Establishes what CUDA runtime will be used for a target, via
+  the :cmake::prop_tgt:`CUDA_RUNTIME_LIBRARY <cmake::prop_tgt:CUDA_RUNTIME_LIBRARY>`
```
bdice, line 31:
```suggestion
  the :cmake:prop_tgt:`CUDA_RUNTIME_LIBRARY <cmake:prop_tgt:CUDA_RUNTIME_LIBRARY>`
```

rapids-cmake/cuda/set_runtime.cmake
```diff
@@ -0,0 +1,65 @@
...
+  and by linking to `CUDA::cudart` or `CUDA::cudart_static`.
+
+ .. note::
+  If using the deprecated `FindCUDA.cmake` you must use the
+  :cmake:command:`rapids_init_runtime` method to properly establish the default
```
bdice, line 36:
```suggestion
  :cmake:command:`rapids_cuda_init_runtime` method to properly establish the default
```

rapids-cmake/cuda/set_runtime.cmake
```diff
@@ -0,0 +1,65 @@
...
+rapids_cuda_set_runtime
+-------------------------------
+
+.. versionadded:: v23.08.00
+
+Establish what CUDA runtime library should be used by a single target
+
+  .. code-block:: cmake
+
+    rapids_cuda_set_runtime( target USE_STATIC (TRUE|FALSE) )
```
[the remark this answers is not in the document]
vyasr: Rather than having this be a true/false, should we instead have `USE_STATIC` be an option rather than a one-valued parameter?

Also, if we want to encourage static linking of cudart perhaps we should make it an option `USE_SHARED` and otherwise default to use static. What do you think?

rapids-cmake/cuda/set_runtime.cmake
```diff
@@ -0,0 +1,65 @@
...
+
+  Establishes what CUDA runtime will be used for a target, via
+  the :cmake:prop_tgt:`CUDA_RUNTIME_LIBRARY <cmake:prop_tgt:CUDA_RUNTIME_LIBRARY>`
+  and by linking to `CUDA::cudart` or `CUDA::cudart_static`.
+
+ .. note::
+  If using the deprecated `FindCUDA.cmake` you must use the
```
vyasr, line 35: Perhaps clarify that you're contrasting with the newer FindCUDAToolkit, which is supported?

rapids-cmake/cuda/set_runtime.cmake
```diff
@@ -0,0 +1,65 @@
...
+  get_target_property(type ${target} TYPE)
+  if(type STREQUAL "INTERFACE_LIBRARY")
+    set(mode INTERFACE)
+  else()
+    set(mode PUBLIC)
+  endif()
+
+  if(${value})
+    set_target_properties(${target} PROPERTIES CUDA_RUNTIME_LIBRARY STATIC)
+    target_link_libraries(${target} ${mode} $<TARGET_NAME_IF_EXISTS:CUDA::cudart_static>)
```
vyasr, line 59: Does it make sense for this to do nothing silently rather than error if the CUDA target doesn't exist? I guess the benefit of using the generator here is to make it independent of ordering on when the CTK is found?

vyasr (changes requested):

rapids-cmake/cuda/set_runtime.cmake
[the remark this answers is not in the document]
robertmaynard: The current mode of having it be `USE_STATIC` was selected so that it reduced the amount of calling code since RAPIDS projects generally have an option to control this called `CUDA_STATIC_RUNTIME`. 

So people can do:
```cmake
rapids_cuda_set_runtime(<target> USE_STATIC ${CUDA_STATIC_RUNTIME})
```

instead of:
```cmake
if(CUDA_STATIC_RUNTIME)
  rapids_cuda_set_runtime(<target> USE_STATIC)
else()
  rapids_cuda_set_runtime(<target>)
endif()
```


Your suggestion of having a default value of `TRUE` does make sense though.

rapids-cmake/cuda/set_runtime.cmake
robertmaynard: Updated the docs to clarify behavior around the FindCUDAToolkit targets

rapids-cmake/cuda/set_runtime.cmake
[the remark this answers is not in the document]
vyasr: Makes sense, reducing friction here is probably more valuable than purity.

robertmaynard:
> Do we need to address [the interface/private question you raised](https://github.com/rapidsai/rapids-cmake/issues/427#issuecomment-1587572641) before we finalize?

I think we should address this before merging.

Currently the approach is that we us PUBLIC linking for all targets (except INTERFACE). After thinking about this more I think that `PRIVATE` by default might be the best approach as it has the least amount of suprises. By that I mean if you call `rapids_cuda_set_runtime` on a target I don't think users expect it would have an impact on anything that links to that target.

Note: I am ignoring `INTERFACE` targets from the above. Since INTERFACE targets can only have `INTERFACE` linking the behavior with those is pre-determined. I do need to update the docs no matter what to explicitly state what behavior you will get based on the target type you provide.

PR #433, opened 2023-06-29
Support scikit-build-core

vyasr:
## Description

This PR adds a new cython-core module of rapids-cmake that can be used with scikit-build-core. `cython-core` is designed as a drop-in replacement for `cython`, allowing consumers to simply change the `include(cython)` to `include(cython-core)` and have things generally work as expected when switching Python build backends.

## Checklist


rapids-cmake/cython-core/cython.cmake
```diff
@@ -0,0 +1,213 @@
...
+  endif()
+
+  set(compile_time_env)
+  if(DEFINED _RAPIDS_COMPILE_COMPILE_TIME_ENV)
+    list(JOIN _RAPIDS_COMPILE_COMPILE_TIME_ENV "," compile_time_env)
+    set(compile_time_env "-X ${compile_time_env}")
+  endif()
+
+  # Combine arguments into a usable string
+  set(args)
```
robertmaynard, line 168: The two foreach loops can be simplified to:
```cmake
set(args)
foreach(arg IN LISTS _rapids_cython_option_vars _rapids_cython_one_value_vars)
  if(${arg})
    list(APPEND args "${${arg}}")
  endif()
endforeach()
```

vyasr: Good call, thanks!

rapids-cmake/cython-core/add_rpath_entries.cmake
```diff
@@ -0,0 +1,93 @@
...
+# Unless required by applicable law or agreed to in writing, software distributed under the License
+# is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+# or implied. See the License for the specific language governing permissions and limitations under
+# the License.
+# =============================================================================
+
+include_guard(GLOBAL)
+
+#[=======================================================================[.rst:
+rapids_cython_add_rpath_entries
```
robertmaynard, line 18: Can you go over the change of this being in `cython-core` instead of `cython`. Is the plan here to remove the `cython` function entirely?

rapids-cmake/cython-core/add_rpath_entries.cmake
```diff
@@ -0,0 +1,93 @@
...
+-------------------------------
+.. versionadded:: v22.12.00
```
robertmaynard, line 20: All the version added lines are wrong

robertmaynard (changes requested):

rapids-cmake/cython-core/add_rpath_entries.cmake
vyasr: For the moment I've separated the two for prototyping purposes because I anticipated needing to make breaking changes in Cython. I was hoping to just create a new set of CMake files that would be mostly drop-in replacements, i.e. for a rapids-cmake consumer to switch to using scikit-build-core they would just change `include(rapids-cython)` to `include(rapids-cython-core)` and then the rest of the function calls would "just work". That would allow a piecemeal migration of RAPIDS libraries rather than introducing a breaking point. This seems especially important since we now have consumers outside of RAPIDS alone (e.g. legate libraries) and so providing a softer off ramp would be valuable.

robertmaynard: That makes sense. As we move out of draft we should encode this information somewhere in the docs so that people understand the differences between `rapids-cython` and `rapids-cython-core` and how you can easily switch between them.

We might also to think about how we can write some smoke tests that verify that both modules share a common subset of API

vyasr: Definitely, I'll keep that in mind. JFYI it'll probably be a while before this PR comes out of draft, there's a lot of upstream work around our CI workflows that needs to get resolved before I think we'll be ready for the scikit-build-core switch. Hopefully 23.10.

rapids-cmake/cython-core/cython.cmake
vyasr, line 1: This file should be obsoleted once https://github.com/vyasr/cython-cmake is migrated to the scikit-build org and published.

rapids-cmake/cython-core/create_modules.cmake
```diff
@@ -0,0 +1,158 @@
...
+rapids_cython_create_modules
+----------------------------
+
+.. versionadded:: v24.02.00
+
+Generate C(++) from Cython and create Python modules.
+
+.. code-block:: cmake
+
+  rapids_cython_create_modules([CXX] [SOURCE_FILES <src1> <src2> ...] [LINKED_LIBRARIES <lib1> <lib2> ... ]  [INSTALL_DIR <install_path>] [MODULE_PREFIX <module_prefix>] )
```
bdice, line 27: Should this be nicely wrapped [like other example docstrings](https://github.com/rapidsai/rapids-cmake/blob/2969d42a155d6045a17cc7e79c83783befd3b7d5/rapids-cmake/cpm/gbench.cmake#L32)? Also there's some inconsistent spacing.
```suggestion
  rapids_cython_create_modules([CXX]
                               [SOURCE_FILES <src1> <src2> ...]
                               [LINKED_LIBRARIES <lib1> <lib2> ...]
                               [INSTALL_DIR <install_path>]
                               [MODULE_PREFIX <module_prefix>])
```

rapids-cmake/cython-core/create_modules.cmake
```diff
@@ -0,0 +1,158 @@
...
+
+.. note::
+  Requires :cmake:command:`rapids_cython_init` to be called before usage.
+
+``CXX``
+  Flag indicating that the Cython files need to generate C++ rather than C.
+
+``SOURCE_FILES``
+  The list of Cython source files to be built into Python extension modules.
+  Note that this function assumes that Cython source files have a one-one
```
bdice, line 41:
```suggestion
  Note that this function assumes that Cython source files have a one-to-one
```

rapids-cmake/cython-core/create_modules.cmake
```diff
@@ -0,0 +1,158 @@
...
+  rapids_cython_create_modules([CXX] [SOURCE_FILES <src1> <src2> ...] [LINKED_LIBRARIES <lib1> <lib2> ... ]  [INSTALL_DIR <install_path>] [MODULE_PREFIX <module_prefix>] )
+
+Creates a Cython target for each provided source file, then adds a
+corresponding Python extension module. Each built library has its RPATH set to
+$ORIGIN.
+
+.. note::
+  Requires :cmake:command:`rapids_cython_init` to be called before usage.
+
+``CXX``
```
bdice, line 36: Is there a reason why we'd ever want this to be false? I'd just force this to true and remove the parameter.

rapids-cmake/cython-core/create_modules.cmake
```diff
@@ -0,0 +1,158 @@
...
+  to CMAKE_CURRENT_SOURCE_DIR relative to PROJECT_SOURCE_DIR.
+
+``MODULE_PREFIX``
+  A prefix used to name the generated library targets. This functionality is
+  useful when multiple Cython modules in different subpackages of the the same
+  project have the same name. The default prefix is the empty string.
+
+``ASSOCIATED_TARGETS``
+  A list of targets that are associated with the Cython targets created in this
+  function. The target<-->associated target mapping is stored and may be
```
bdice, line 62:
```suggestion
  function. The target to associated target mapping is stored and may be
```

rapids-cmake/cython-core/create_modules.cmake
vyasr: When I originally wrote rapids-cython, I wanted to allow for C projects in case we had something like ucx-py or dask-cuda writing C (not C++) extensions. In practice, that hasn't turned out to be the case AFAIK. That being said, I'd like to leave this in place for now because I've endeavored to make the new functions drop-in replacements for the old ones and this would be an API break. I plan future changes that will be API breaking, so I can make this change then. For now, the scikit-build-core PRs in other repos are made much cleaner by avoiding to have to change any CMake other than the `include(rapids-cython)`.

vyasr:
> Missing documentation and cmake-format updates

I added docs. The new functions are identical to the old ones, just included differently (`include(rapids-cython-core)` instead of `include(rapids-cython)`), so there are no new additions required for cmake-format at this stage.

rapids-cmake/cython/init.cmake
```diff
@@ -38,6 +38,11 @@ Result Variables
 #]=======================================================================]
 macro(rapids_cython_init)
   list(APPEND CMAKE_MESSAGE_CONTEXT "rapids.cython.init")
+  include("${rapids-cmake-dir}/cmake/detail/policy.cmake")
+  rapids_cmake_policy(DEPRECATED_IN 24.02
+                      REMOVED_IN 24.08
```
bdice, line 43: Removing in 24.06 would also be fine by me.

bdice (approved):

rapids-cmake/cython/init.cmake
vyasr: I'm fine with either. @robertmaynard you can make the final call.

robertmaynard: Given that not all consumers of rapids are using the same release frequency I think we can wait to 24.08 to yank this

robertmaynard (approved):

PR #439, opened 2023-07-20
Conditionally modify envvar vs. global CMAKE_PREFIX_PATH in `rapids_cmake_support_conda_env`

trxcllnt:
## Description
This PR updates the logic when `rapids_cmake_support_conda_env()` is called with `MODIFY_PREFIX_PATH`.

The current logic prepends the `$PREFIX`/`$BUILD_PREFIX`/`$CONDA_PREFIX` paths to the global `CMAKE_PREFIX_PATH` variable.

This PR proposes the following changes:
1. Append instead of prepend the  `$PREFIX`/`$BUILD_PREFIX`/`$CONDA_PREFIX` paths
2. Prefer modifying the `CMAKE_PREFIX_PATH` _environment variable_ instead of the global `CMAKE_PREFIX_PATH` variable
3. Only modify the global `CMAKE_PREFIX_PATH` variable if it's already populated (preserving paths in `ENV{CMAKE_PREFIX_PATH}`)

## Checklist


rapids-cmake/cmake/support_conda_env.cmake
```diff
@@ -76,12 +76,45 @@ function(rapids_cmake_support_conda_env target)
...
+      list(APPEND CMAKE_PREFIX_PATH ${_PATHS})
+      list(REMOVE_DUPLICATES CMAKE_PREFIX_PATH)
+      set(CMAKE_PREFIX_PATH ${CMAKE_PREFIX_PATH} PARENT_SCOPE)
+      message(VERBOSE "CMAKE_PREFIX_PATH set to: ${CMAKE_PREFIX_PATH}")
+    endmacro()
+
+    macro(modify_cmake_prefix_path_envvar)
+      cmake_parse_arguments("" "" "" "PATHS" ${ARGN})
+      cmake_path(CONVERT "$ENV{CMAKE_PREFIX_PATH}" TO_CMAKE_PATH_LIST _paths NORMALIZE)
+      list(APPEND _paths ${_PATHS})
```
robertmaynard, line 98: Can we make the variable usage between `modify_cmake_prefix_path_global` and `modify_cmake_prefix_path_envvar` consistent. It looks like in `modify_cmake_prefix_path_envvar` we append to `_paths` instead of `_PATHS`.

robertmaynard (changes requested):

rapids-cmake/cmake/support_conda_env.cmake
trxcllnt: The local `_paths` is the version of the `$ENV{CMAKE_PREFIX_PATH}` converted to a list and normalized. `_PATHS` is the argument list of paths to append.

robertmaynard: Ahhhhhh.

rapids-cmake/cmake/support_conda_env.cmake
```diff
@@ -76,12 +76,49 @@ function(rapids_cmake_support_conda_env target)
 
   if(in_conda_build OR in_conda_prefix)
 
+    #
+    macro(modify_cmake_prefix_path_global)
+      cmake_parse_arguments(_RAPIDS "" "" "PATHS" ${ARGN})
+
+      if(NOT ("$ENV{CMAKE_PREFIX_PATH}" STREQUAL ""))
```
robertmaynard, line 83:
```suggestion
      if(DEFINED ENV{CMAKE_PREFIX_PATH})
```
I think that makes it clearer

rapids-cmake/cmake/support_conda_env.cmake
```diff
@@ -76,12 +76,49 @@ function(rapids_cmake_support_conda_env target)
...
+      list(APPEND _paths ${_RAPIDS_PATHS})
+      list(REMOVE_DUPLICATES _paths)
+      cmake_path(CONVERT "${_paths}" TO_NATIVE_PATH_LIST _paths NORMALIZE)
+      set(ENV{CMAKE_PREFIX_PATH} ${_paths})
+      message(VERBOSE "ENV{CMAKE_PREFIX_PATH} set to: $ENV{CMAKE_PREFIX_PATH}")
+    endmacro()
+
+    #
+    macro(modify_cmake_prefix_path)
+      if(NOT ("${CMAKE_PREFIX_PATH}" STREQUAL ""))
```
robertmaynard, line 110:
```suggestion
      if(DEFINED CMAKE_PREFIX_PATH)
```

robertmaynard (changes requested):
We should prefer using`DEFINED` since it makes it clearer.

If some user some how defines `CMAKE_PREFIX_PATH` to a empty string, they win an award :)

robertmaynard (approved):

PR #446, opened 2023-07-25
Ensure tests the modify same git repo don't execute at the same time

robertmaynard:
## Description
A subset of the rapids_cpm tests modify the same git repo active branch which isn't safe. We need to serialize those tests to ensure everything works robustly.

## Checklist


bdice (approved):

testing/utils/cmake_test.cmake
```diff
@@ -141,6 +142,17 @@ function(add_cmake_test mode source_or_dir)
       message(FATAL_ERROR "${mode} mode not one of the valid modes (config|build|install) by add_cmake_build_test")
     endif()
 
+    if(RAPIDS_TEST_SERIAL)
+      set_tests_properties(${test_name} PROPERTIES RUN_SERIAL ON)
+      if(TEST ${test_name}_configure)
+        message(STATUS "${test_name}_configure is serial")
```
vyasr, line 148: Slightly inconsistent with a status message for the configure but not the build.

vyasr (approved):
LGTM (one minor comment). Thanks!

testing/utils/cmake_test.cmake
robertmaynard: Ohh that was a debug line, will remove it.

PR #458, opened 2023-09-13
cpm overrides don't occur when `CPM_<pkg>_SOURCE` exists

robertmaynard:
## Description
Since the rapids-cmake cpm overrides hook into fetch content they previously would take priority over `CPM_<pkg>_SOURCE`. Now we don't add overrides when `CPM_<pkg>_SOURCE` exists.

## Checklist

### DIFF
--- a/rapids-cmake/cpm/package_override.cmake
+++ b/rapids-cmake/cpm/package_override.cmake
@@ -49,11 +49,14 @@ projects.
 
 .. note::
 
+  .. versionadded:: v23.10.00
+    When the variable `CPM_<package_name>_SOURCE` exists, any override entries
+    for `package_name` will be ignored.
+
   If the override file doesn't specify a value or package entry the default
   version will be used.
 
   Must be called before any invocation of :cmake:command:`rapids_cpm_find`.
-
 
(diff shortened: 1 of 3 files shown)


robertmaynard:
@trxcllnt This PR should fix the bugs you have encountered with rapids-cmake overrides and cpm pinning via `CPM_<pkg>_SOURCE`.

rapids-cmake/cpm/package_override.cmake
```diff
@@ -49,12 +49,15 @@ projects.
 
 .. note::
 
+  .. versionadded:: v23.10.00
```
bdice, line 52: Do you want this to be inside the `.. note::`? It looks a little funny when rendered. Maybe put this above the note block.

bdice (approved):

rapids-cmake/cpm/package_override.cmake
robertmaynard: I moved the element to a separate note. I want to explicitly call this behavior change out so people understand what is happening so I am using a note element.

bdice (approved):

PR #463, opened 2023-09-22
Quote the list of patch files in case they have spaces in their paths

ericniebler:
## Description
The rapids cpm scripts fail when populating a dependency with patches if the build directory contains spaces in its path name. The reason is because the paths are not properly quoted in the cmake patch script generated from `command_template.cmake.in`.

This PR properly quotes the list of patch file names.

## Checklist


bdice (commented):
Thanks @ericniebler! Could you add a test for this case?

bdice (approved):

PR #467, opened 2023-09-28
Thrust when exported now automatically calls `thrust_create_target`

robertmaynard:
## Description
Instead of having each consuming project inject a code block into the `project-config.cmake` we instead have rapids-cmake call `thrust_create_target` right after `find_package(Thrust)`, only when thrust was found.

This fixes two existing issues:
  1. It removes the need for `rapids_cpm_thrust` to remove the the custom Thrust target from the global target set. Now that target will exist when rapids-cmake tries to promote it to global
  2. It removes issues when multiple calls to `find_package(thrust)` occur before any calls to `thrust_create_target`. This creates some vexing issues when the thrust find calls resolve to different major versions of thrust

## Checklist


bdice (approved):
LGTM

rapids-cmake/cpm/thrust.cmake
```diff
@@ -90,18 +97,27 @@ function(rapids_cpm_thrust NAMESPACE namespaces_name)
...
 
   if(Thrust_SOURCE_DIR AND _RAPIDS_INSTALL_EXPORT_SET AND to_install)
     include("${rapids-cmake-dir}/export/find_package_root.cmake")
+    include("${rapids-cmake-dir}/export/detail/post_find_package_code.cmake")
     rapids_export_find_package_root(INSTALL Thrust
                                     [=[${CMAKE_CURRENT_LIST_DIR}/../../rapids/cmake/thrust]=]
                                     ${_RAPIDS_INSTALL_EXPORT_SET})
+    rapids_export_post_find_package_code(INSTALL Thrust "${post_find_code}"
+                                         ${_RAPIDS_INSTALL_EXPORT_SET})
   endif()
```
vyasr, line 121: Maybe combine these two blocks? Setting a variable to `set(BUILD_TYPE BUILD|INSTALL)` and using `${RAPIDS_${BUILD_TYPE}_EXPORT_SET}` should make all but the paths for `find_package_root` the same.

testing/export/CMakeLists.txt
vyasr, line 1: I'm confused, why did the alphabetization (ordering) change in this file when adding the new post-find tests?

testing/export/export_package-install-post-find-code.cmake
```diff
@@ -0,0 +1,39 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#=============================================================================
+include(${rapids-cmake-dir}/export/package.cmake)
+include(${rapids-cmake-dir}/export/detail/post_find_package_code.cmake)
+
+rapids_export_package( INSTALL FAKE_PACKAGE test_export_set VERSION 22.08)
+rapids_export_post_find_package_code(INSTALL FAKE_PACKAGE "set(a ON)" test_export_set)
```
vyasr, line 20: nit: I notice some different spacing in a couple places in these files w.r.t the space after the open parentheses.

vyasr (changes requested):

rapids-cmake/cpm/thrust.cmake
```diff
@@ -90,18 +97,27 @@ function(rapids_cpm_thrust NAMESPACE namespaces_name)
   set(multi_value)
   cmake_parse_arguments(_RAPIDS "${options}" "${one_value}" "${multi_value}" ${ARGN})
 
+  set(post_find_code "if(NOT TARGET ${namespaces_name}::Thrust)"
+                     "  thrust_create_target(${namespaces_name}::Thrust FROM_OPTIONS)" "endif()")
```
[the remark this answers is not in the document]
robertmaynard: I don't believe that this should break any downstream consumer that has a proper `<pkg>-config.cmake`. Since `<pkg>-config.cmake` can be called numerous times calls to `thrust_create_target` are wrapped inside a check for the target e.g.:
```cmake
if(NOT TARGET <pkg>::Thrust)
  thrust_create_target(<pkg>::Thrust FROM_OPTIONS)
endif()
```

So if we do break some downstream consumers, those consumer are already writing out a bad config file and need a change.

testing/export/CMakeLists.txt
robertmaynard: I realized that the tests aren't in sorted order and moved them around. I am happy to revert this.

testing/export/export_package-build-post-find-code.cmake
```diff
@@ -0,0 +1,50 @@
...
+
+set(to_match [=[if(FAKE_PACKAGE_FOUND)_set(a ON)_set(b ON)_set(c ON)_set(d ON)_set(e ON)__endif()]=])
+
+file(STRINGS "${CMAKE_CURRENT_BINARY_DIR}/install_export_set.cmake" contents NEWLINE_CONSUME)
+string(REPLACE "\n" "_" contents "${contents}")
+
+string(FIND "${contents}" "${to_match}" is_found)
+if(is_found EQUAL -1)
+  message(FATAL_ERROR "rapids_export_write_dependencies(BUILD) failed to record rapids_export_post_find_package_code")
+endif()
```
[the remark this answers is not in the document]
robertmaynard: You are correct, dropped the first half.

rapids-cmake/cpm/thrust.cmake
robertmaynard: Done.

testing/export/CMakeLists.txt
vyasr: I'm fine either way, on first read I thought the original ordering was correct but I see now.

rapids-cmake/cpm/thrust.cmake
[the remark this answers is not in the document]
vyasr: Got it.

rapids-cmake/export/find_package_root.cmake
```diff
@@ -29,7 +29,8 @@ needs to be set to the provided path.
   rapids_export_find_package_root( (BUILD|INSTALL)
                                    <PackageName>
                                    <directory_path>
-                                   <ExportSet>
+                                   (<ExportSetName> | EXPORT_SET [ExportSetName])
```
vyasr, line 32: Is this supporting both uses for backwards compatibility? If so, can we make a plan to deprecate the former usage? Maybe using `rapids_cmake_policy`?

rapids-cmake/export/detail/post_find_package_code.cmake
```diff
@@ -0,0 +1,88 @@
...
+include_guard(GLOBAL)
+
+#[=======================================================================[.rst:
+rapids_export_post_find_package_code
+------------------------------------
+
+.. versionadded:: v23.12.00
+
+Record for <PackageName> a set of CMake instructions to be executed after the package
+has ben found successfully.
```
vyasr, line 25:
```suggestion
has been found successfully.
```

rapids-cmake/export/detail/post_find_package_code.cmake
```diff
@@ -0,0 +1,88 @@
...
+
+.. code-block:: cmake
+
+  rapids_export_post_find_package_code((BUILD|INSTALL)
+                                       <PackageName>
+                                       <code>
+                                       (<ExportSetName> | EXPORT_SET [ExportSetName])
```
vyasr, line 32: Assuming the other function has both options for both primarily for backwards compat, could we just require the `EXPORT_SET` keyword in this new function? Both calls I see are using it anyway.
```suggestion
                                       EXPORT_SET [ExportSetName]
```

rapids-cmake/export/detail/post_find_package_code.cmake
```diff
@@ -0,0 +1,88 @@
...
+  list(APPEND CMAKE_MESSAGE_CONTEXT "rapids.export.post_find_package_code")
+
+  set(options "")
+  set(one_value EXPORT_SET CONDITION)
+  set(multi_value "")
+  cmake_parse_arguments(_RAPIDS "${options}" "${one_value}" "${multi_value}" ${ARGN})
+  # handle when we are given just an export set name and not `EXPORT_SET <name>`
+  if(_RAPIDS_UNPARSED_ARGUMENTS AND NOT _RAPIDS_COMPONENTS_EXPORT_SET)
+    set(_RAPIDS_EXPORT_SET ${_RAPIDS_UNPARSED_ARGUMENTS})
+  endif()
```
vyasr, line 70: If we accept my prior suggestion
```suggestion
```

rapids-cmake/export/write_dependencies.cmake
```diff
@@ -88,14 +89,12 @@ endif()\n")
...
 
-  if(find_root_dirs)
-    foreach(package IN LISTS find_root_dirs)
-      get_property(root_dir_path TARGET rapids_export_${type}_${export_set}
-                   PROPERTY "FIND_ROOT_FOR_${package}")
-      set(dep_content "set(${package}_ROOT \"${root_dir_path}\")")
-      string(APPEND _RAPIDS_EXPORT_CONTENTS "${dep_content}\n")
-    endforeach()
-  endif()
+  foreach(package IN LISTS find_root_dirs)
```
vyasr, line 92: What changed that `find_root_dirs` is guaranteed to be defined?

vyasr (changes requested):

rapids-cmake/export/find_package_root.cmake
robertmaynard: Yes it is for backwards compatibility, as RAFT, cudf, cuml, etc use this function. 

Deprecation does sound reasonable. Depreacte in 23.12 and remove in the second release of 24.X feels reasonable to me.

rapids-cmake/export/detail/post_find_package_code.cmake
robertmaynard: Done ( Had to revise tests as well ).

rapids-cmake/export/detail/post_find_package_code.cmake
robertmaynard: Done

rapids-cmake/export/write_dependencies.cmake
robertmaynard: `foreach( IN LISTS)` will ignore empty variable, so the if condition was never needed. Just noticed now that I could remove it.

rapids-cmake/export/find_package_root.cmake
robertmaynard: Looking at keeping symetry across the `rapids_export` functions I am also going to update `rapids_export_find_package_root`

vyasr (approved):

PR #470, opened 2023-10-04
Express Python version in dependencies.yaml.

bdice:
## Description
This PR adds `py_version` to the `dependencies.yaml` file so that `rapids-cmake` expresses this in the same way as every other RAPIDS repository. The primary reason this might matter is for ensuring that compatible Python versions are used for style checks.

## Checklist


robertmaynard (approved):

vyasr (approved):
We should also

PR #474, opened 2023-10-16
target from write_git_revision now works with export sets

robertmaynard:
## Description
The git revision target previously wouldn't work when you needed to export targets. Those prohibited projects from producing static targets while also using this function.

## Checklist


bdice (commented):
Do we need a test for this?

robertmaynard:
> Do we need a test for this?

Yeah we do. Writing one now

bdice (approved):

PR #483, opened 2023-11-13
Remove invalid argument to find_package_root

robertmaynard:
## Description
Remove an invalid argument to rapids_export_find_package_root

## Checklist


bdice (approved):

bdice:
I wonder if this is the root cause of all the deprecation warnings I've been seeing in build logs (even after I thought we had fixed the calls to deprecated functions).

robertmaynard:
> I wonder if this is the root cause of all the deprecation warnings I've been seeing in build logs (even after I thought we had fixed the calls to deprecated functions).

Most likely yes. The other cases I am seeing are all inside rapids-cmake tests that verified the deprecated API

PR #484, opened 2023-11-13
Remove deprecated function usages

robertmaynard:
## Description
Remove usage of function signatures that are no longer supported in 23.04

## Checklist


bdice (approved):

bdice:
image
Is this referencing the previous gpuCI system? Is something set wrong in the branch protections?

cryos:
This looks like it should have also made it into the 23.12 branch, I am seeing deprecation warnings if I use 23.12 that this patch would silence when calling `rapids_find_generate_module`.

robertmaynard:
> This looks like it should have also made it into the 23.12 branch, I am seeing deprecation warnings if I use 23.12 that this patch would silence when calling `rapids_find_generate_module`.

Yeah it should have. Since it is just a diagnostic bug I think it isn't worth releasing a 23.12 patch release of rapids-cmake

PR #496, opened 2023-12-04
Add missing nvcomp targets

vyasr:
## Description

The list of targets marked as global when finding nvcomp is incomplete and should include the gdeflate and bitcomp libs as well.

## Checklist

### DIFF
--- a/rapids-cmake/cpm/nvcomp.cmake
+++ b/rapids-cmake/cpm/nvcomp.cmake
@@ -155,7 +155,11 @@ function(rapids_cpm_nvcomp)
   # provide consistent targets between a found nvcomp and one building from source
   if(NOT TARGET nvcomp::nvcomp AND TARGET nvcomp)
     add_library(nvcomp::nvcomp ALIAS nvcomp)
+  endif()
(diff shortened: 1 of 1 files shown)


robertmaynard (approved):

rapids-cmake/cpm/nvcomp.cmake
```diff
@@ -152,6 +155,8 @@ function(rapids_cpm_nvcomp)
   # provide consistent targets between a found nvcomp and one building from source
   if(NOT TARGET nvcomp::nvcomp AND TARGET nvcomp)
     add_library(nvcomp::nvcomp ALIAS nvcomp)
+    add_library(nvcomp::nvcomp_gdeflate ALIAS nvcomp_gdeflate)
```
robertmaynard, line 158: We should have these in separate logic blocks to guard against them going away in the future

robertmaynard (changes requested):

rapids-cmake/cpm/nvcomp.cmake
vyasr: Done.

PR #504, opened 2023-12-12
Disable NVBench CUPTI support by default.

bdice:
## Description
This PR disables CUPTI support in NVBench by default.

In essentially all of RAPIDS, we have to set `NVBench_ENABLE_CUPTI OFF`. By default, this is enabled, but it leads to several problems. First, when CUPTI support is enabled, the benchmark executables require the same CUDA major.minor version at runtime, meaning that (for instance) we can't build with CUDA 11.8 and expect that the benchmarks can be run with CUDA 11.5. Generally RAPIDS wants to produce builds that follow rules for CUDA minor version compatibility within a CUDA major version. Secondly, developers (especially new developers building from source) frequently run into issues where CUPTI is not available. We don't make `cuda-cupti-dev` a conda dependency in cudf, specifically because we disable CUPTI support most of the time. However, users building from source or using their own build invocations won't have CUPTI disabled, and run into problems compiling because of this missing dependency.

Examples:
- https://github.com/rapidsai/cudf/blob/d1ba966536e30dc728c09726f6ec595100bfea68/conda/recipes/libcudf/build.sh#L9
- https://github.com/rapidsai/cudf/blob/d1ba966536e30dc728c09726f6ec595100bfea68/.github/workflows/pr.yaml#L140

## Checklist

### DIFF
--- a/rapids-cmake/cpm/nvbench.cmake
+++ b/rapids-cmake/cpm/nvbench.cmake
@@ -43,6 +43,10 @@ across all RAPIDS projects.
   Will build nvbench statically. No local searching for a previously
   built version will occur.
 
+.. versionchanged:: v24.02.00
+
+NVBench_ENABLE_CUPTI is set to OFF by default.
+
 Result Targets
 ^^^^^^^^^^^^^^
   nvbench::nvbench target will be created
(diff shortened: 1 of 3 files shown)


robertmaynard (approved):

robertmaynard (commented):
Please add a note to the documentation for this function to say by default the cupti support for nvbench will be disabled

PR #505, opened 2023-12-13
Manually invoke install rules for components

vyasr:
## Description

CCCL's CMake is currently set up to exit early whenever it's added via `add_subdirectory`, which inadvertently prevents the install rules of its components (Thrust/CUB/libcudacxx) from being called. We don't currently have a way to tell CCCL not to do this because preventing that early exit leads to other undesirable outcomes (such as the CCCL::CCCL targets not being created correctly). Therefore, to work around this we must include the various install rule modules directly until this can be fixed upstream in CCCL.

In addition, we require a couple of patches to CCCL's CMake so that 1) the correct config files are installed, and 2) so that multiple invocations of `find_package(CCCL)` will work. These patches have already been fixed on the latest CCCL, see NVIDIA/cccl#298 and NVIDIA/cccl#1157.

## Checklist


bdice (approved):
Approving based on https://github.com/rapidsai/rmm/pull/1404 passing CI.

bdice:
@vyasr I'm going to merge this - we already discussed adding a test after the fact, which would be helpful to verify the future bug fixes in CCCL.

PR #513, opened 2023-12-20
rapids_cpm_cccl preserve install location details from first invocation.

robertmaynard:
## Description
Previously we would run the install rules each time rapids_cpm_cccl was called, since we checked for `CCCL_SOURCE_DIR` and that could have been defined by an earlier invocation at the same scope.

This causes issue when the first invocation of `rapids_cpm_cccl` used a custom install location.
CMake's install rules would use the first location, but CCCL install files would contain the second location since they got re-configured on each execution.

## Checklist

### DIFF
--- a/rapids-cmake/cpm/cccl.cmake
+++ b/rapids-cmake/cpm/cccl.cmake
@@ -92,9 +92,10 @@ function(rapids_cpm_cccl)
                   EXCLUDE_FROM_ALL ${exclude}
                   OPTIONS "CCCL_ENABLE_INSTALL_RULES ${to_install}")
 
-  # rapids_cpm_cccl can be called multiple times from the same scope such as from cudf/CMakeLists.txt
-  # and cudf's call to find_package(rmm). In these situations, subsequent invocations will still have
-  # `CCCL_SOURCE_DIR` set due to how `rapids_cpm_find` early termination sets up variables
+  # rapids_cpm_cccl can be called multiple times from the same scope such as from
+  # cudf/CMakeLists.txt and cudf's call to find_package(rmm). In these situations, subsequent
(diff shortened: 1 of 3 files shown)


rapids-cmake/cpm/cccl.cmake
```diff
@@ -92,10 +92,19 @@ function(rapids_cpm_cccl)
                   EXCLUDE_FROM_ALL ${exclude}
                   OPTIONS "CCCL_ENABLE_INSTALL_RULES ${to_install}")
 
-  if(CCCL_SOURCE_DIR AND to_install)
+  # rapids_cpm_ccl can be called multiple times from the same scope such as from cudf/CMakeLists.txt
+  # and cudfs call to find_package(rmm) In these situations the second+ invocations will still have
```
bdice, line 96:
```suggestion
  # and cudf's call to find_package(rmm). In these situations, subsequent invocations will still have
```

rapids-cmake/cpm/cccl.cmake
```diff
@@ -92,10 +92,19 @@ function(rapids_cpm_cccl)
                   EXCLUDE_FROM_ALL ${exclude}
                   OPTIONS "CCCL_ENABLE_INSTALL_RULES ${to_install}")
 
-  if(CCCL_SOURCE_DIR AND to_install)
+  # rapids_cpm_ccl can be called multiple times from the same scope such as from cudf/CMakeLists.txt
```
bdice, line 95:
```suggestion
  # rapids_cpm_cccl can be called multiple times from the same scope such as from cudf/CMakeLists.txt
```

bdice (approved):
Thanks, this looks good to me.

PR #516, opened 2024-01-10
Add link libraries to generate_resource_spec cmake

davidwendt:
## Description
Adds `-lpthread -lrt -ldl` link library/options to the `generate_resource_spec.cmake` to resolve link errors when building the resource_spec generator on some systems.

## Checklist


rapids-cmake/test/generate_resource_spec.cmake
```diff
@@ -73,7 +73,7 @@ function(rapids_test_generate_resource_spec DESTINATION filepath)
     if(CUDAToolkit_FOUND)
       set(compile_options "-I${CUDAToolkit_INCLUDE_DIRS}" "-DHAVE_CUDA")
     endif()
-    set(link_options ${CUDA_cudart_LIBRARY})
+    set(link_options ${CUDA_cudart_LIBRARY} "-lpthread -lrt -ldl")
```
KyleFromNVIDIA, line 76: This shouldn't be quoted - `execute_process()` will treat `"-lpthread -lrt -ldl"` as a single argument.

KyleFromNVIDIA (changes requested):
Looks good overall minus the quoting issue. In the long run, I'd like to see the resource spec generator built by CMake's build graph if possible - that's what CMake is for :)

rapids-cmake/test/generate_resource_spec.cmake
davidwendt: So should it be this?
```
set(link_options ${CUDA_cudart_LIBRARY} -lpthread -lrt -ldl)
```

KyleFromNVIDIA: Yes, exactly.

davidwendt: I was following the pattern from https://github.com/rapidsai/rapids-cmake/pull/516/files#diff-92f51ddffa5e81afb760f384afc6f2249d2b75ed468d9f4fbaa77bc8560a8ed1R74
```
set(compile_options "-I${CUDAToolkit_INCLUDE_DIRS}" "-DHAVE_CUDA")
```

KyleFromNVIDIA: Note that in that example there are quotes surrounding each argument, not the whole argument list. The equivalent here would be:

```cmake
set(link_options ${CUDA_cudart_LIBRARY} "-lpthread" "-lrt" "-ldl")
```

robertmaynard (approved):

KyleFromNVIDIA (approved):

robertmaynard:
> Looks good overall minus the quoting issue. In the long run, I'd like to see the resource spec generator built by CMake's build graph if possible - that's what CMake is for :)

We are due this year to bump the minimum required CMake version, and moving this to the build graph and leveraging the ability to generate the ctest spec at runtime is part of it :)

PR #517, opened 2024-01-10
refactor CUDA versions in dependencies.yaml

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/7.

Proposes splitting the `cuda-version` dependency in `dependencies.yaml` out to its own thing, separate from the bits of the CUDA Toolkit this project needs.

### Benefits of this change

* prevents accidental inclusion of multiple `cuda-version` version in environments
* reduces update effort (via enabling more use of globs like `"12.*"`)
* improves the chance that errors like "`conda` recipe is missing a dependency" are caught in CI

## Checklist

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -4,7 +4,8 @@ files:
     output: none
     includes:
       - build
-      - cudatoolkit
+      - cuda
+      - cuda_version
       - docs
       - py_version
       - test
@@ -17,7 +18,7 @@ files:
   docs:
     output: none
     includes:
-      - cudatoolkit
+      - cuda_version
       - docs
 channels:
   - rapidsai
@@ -80,9 +81,8 @@ dependencies:
           - matrix:
               cuda: "12.0"
             packages:
-              - cuda-version=12.0
               - cuda-nvcc
-  cudatoolkit:
+  cuda_version:
     specific:
       - output_types: conda
         matrices:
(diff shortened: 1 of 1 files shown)


vyasr (approved):

AyodeAwe (approved):

dependencies.yaml
```diff
@@ -80,51 +81,72 @@ dependencies:
...
+              - cuda-version=12.0
+  cuda:
+    specific:
+      - output_types: conda
+        matrices:
+          - matrix:
+              cuda: "11.2"
+            packages:
               - cudatoolkit
               - gcc<11.0.0
```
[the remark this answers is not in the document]
jameslamb: ok yeah that makes sense to me, thanks

dependencies.yaml
```diff
@@ -34,7 +35,22 @@ dependencies:
           - c-compiler
           - cxx-compiler
           - make
+          - sysroot_linux-64==2.17
     specific:
+      - output_types: conda
+        matrices:
+          - matrix:
+              cuda: "11.[245]"
```
bdice, line 43: @jameslamb What do you think of this syntax? This is valid glob syntax and should work since it's handled by `fnmatch.fnmatch`. Just wanted to point it out explicitly to you since it's different than what we've been writing elsewhere.

dependencies.yaml
```diff
@@ -35,6 +36,30 @@ dependencies:
...
+          - matrix:
+              cuda: "12.*"
+            packages:
+              - gcc<13.0.0
+      - output_types: conda
+        matrices:
+          - matrix:
+              arch: x86_64
+            packages:
+              - sysroot_linux-64==2.17
```
bdice, line 58: It looks like we were requiring `sysroot_linux-64` on ARM builds before this PR. I fixed that.

bdice (changes requested):
Blocking merge since I made a few significant changes (improvements, hopefully).

dependencies.yaml
jameslamb: I thought this could lead to surprises if there were ever double-digit releases (`11.10+`), but I guess that isn't true and `[]` matches a single character.

On Python 3.10.12:

```python
import fnmatch

fnmatch.fnmatch("11.12", "11.[1245]")
# False

fnmatch.fnmatch("11.12", "11.[1245][12]")
# True

fnmatch.fnmatch("11.12", "11.[245][12]")
# False
```

Given that, I support this and think it's a nice simplification. Thanks for the `@`!

vyasr:
Changes look good to me.

bdice (approved):

PR #520, opened 2024-01-11
Fix message context

### LINKED ISSUE #514: [BUG] `rapids_cython_init()` does not clear its modification to `CMAKE_MESSAGE_CONTEXT`
There is no matching `list(POP_BACK CMAKE_MESSAGE_CONTEXT)` for https://github.com/rapidsai/rapids-cmake/blob/branch-24.02/rapids-cmake/cython/init.cmake#L40. Since `rapids_cython_init()` is a macro, this doesn't reset on scope-exit.

vyasr:
## Description

Resolves #514 

## Checklist

### DIFF
--- a/rapids-cmake/cython-core/init.cmake
+++ b/rapids-cmake/cython-core/init.cmake
@@ -1,5 +1,5 @@
 # =============================================================================
-# Copyright (c) 2023, NVIDIA CORPORATION.
+# Copyright (c) 2023-2024, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
 # in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


KyleFromNVIDIA (approved):
+1

I wonder if there would be a way to test this to make sure future macros don't have this problem...

vyasr:
That's a good question. I'm not sure if there's an easy, semantically-aware solution. The simplest purely textual solution I could think of would be writing a pre-commit hook that just counts instances of the insertion and pop_back. There's probably an ideal middle ground that we could find.

robertmaynard:
> That's a good question. I'm not sure if there's an easy, semantically-aware solution. The simplest purely textual solution I could think of would be writing a pre-commit hook that just counts instances of the insertion and pop_back. There's probably an ideal middle ground that we could find.

I can't think of a nice way that doesn't include parsing the source code.

PR #523, opened 2024-01-16
Mark all cccl and cuco kernels with hidden visibility

robertmaynard:
## Description
Patch both CCCL and CUCO to have only internal linkage.

For cuco I am working on upstreaming these changes ( https://github.com/NVIDIA/cuCollections/pull/422 ). Once that is accepted and we have validated that moving cuco is stable ( e.g. changes around `cuco::experimental::static_set` ) we can drop this patch set.

For cccl the long term fix is to move to CCCL 2.3+, but due to issues ( https://github.com/NVIDIA/cccl/pull/1249, maybe others ) that isn't viable for the 24.02 timeframe.
Since the CCCL changes mean C++ and CUDA sources have non compatible ABI's, we need to specify `THRUST_DISABLE_ABI_NAMESPACE` and `THRUST_IGNORE_ABI_NAMESPACE_ERROR` so that we don't change ABI in rapids-cmake consumers since they expect 2.2 behavior.

## Checklist

### DIFF
--- a/rapids-cmake/cpm/cccl.cmake
+++ b/rapids-cmake/cpm/cccl.cmake
@@ -1,5 +1,5 @@
 #=============================================================================
-# Copyright (c) 2023, NVIDIA CORPORATION.
+# Copyright (c) 2023-2024, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


bdice:
Looks good to me. We can merge this after testing at least a few of RAFT, cuML, cuGraph, etc.

robertmaynard:
I think the valid builds in raft/cudf/cugraph are sufficient that this PR won't break anything. 
That will also unblock the cudf CI external linkage checks as well.

PR #525, opened 2024-01-16
Always download repos when they are being patched

vyasr:
## Description

This PR updates `rapids_cpm_find` to force downloading of a package whenever a nonempty patch command exists.

## Checklist

### DIFF
--- a/testing/cpm/CMakeLists.txt
+++ b/testing/cpm/CMakeLists.txt
@@ -23,6 +23,7 @@ add_cmake_config_test( cpm_find-existing-target )
 add_cmake_config_test( cpm_find-existing-target-to-export-sets )
 add_cmake_config_test( cpm_find-gtest-no-gmock )
 add_cmake_config_test( cpm_find-options-escaped )
+add_cmake_config_test( cpm_find-patch-command)
 add_cmake_config_test( cpm_find-restore-cpm-vars )
 add_cmake_config_test( cpm_find-version-explicit-install.cmake )
 
--- a/testing/cpm/CMakeLists.txt
+++ b/testing/cpm/CMakeLists.txt
@@ -23,7 +23,7 @@ add_cmake_config_test( cpm_find-existing-target )
 add_cmake_config_test( cpm_find-existing-target-to-export-sets )
 add_cmake_config_test( cpm_find-gtest-no-gmock )
 add_cmake_config_test( cpm_find-options-escaped )
-add_cmake_config_test( cpm_find-patch-command)
+add_cmake_config_test( cpm_find-patch-command )
 add_cmake_config_test( cpm_find-restore-cpm-vars )
 add_cmake_config_test( cpm_find-version-explicit-install.cmake )
 
--- a/testing/cpm/cpm_find-patch-command/CMakeLists.txt
+++ b/testing/cpm/cpm_find-patch-command/CMakeLists.txt
@@ -1,5 +1,5 @@
 #=============================================================================
-# Copyright (c) 2023, NVIDIA CORPORATION.
+# Copyright (c) 2023-2024, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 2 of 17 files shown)


vyasr:
@robertmaynard this is close, but there is an issue with passing an empty patch command. The issue comes from passing empty strings to `rapids_cpm_find` and how they are parsed by `cmake_parse_arguments`, which seems to drop that string even if it is quoted in the `rapids_cpm_find` function call. I tried using the `cmake_parse_arguments(parse_argv 0...)` syntax instead, but that causes the `cpm_find-options-escaped-*` tests to fail because quotes around CPM args are no longer preserved. I'm not sure what the best approach is here and could use some advice.

robertmaynard:
> @robertmaynard this is close, but there is an issue with passing an empty patch command. The issue comes from passing empty strings to `rapids_cpm_find` and how they are parsed by `cmake_parse_arguments`, which seems to drop that string even if it is quoted in the `rapids_cpm_find` function call. I tried using the `cmake_parse_arguments(parse_argv 0...)` syntax instead, but that causes the `cpm_find-options-escaped-*` tests to fail because quotes around CPM args are no longer preserved. I'm not sure what the best approach is here and could use some advice.

@vyasr I tried playing around as well and I don't see a great solution to this. I think we are going to do is change around `rapids_cpm_generate_patch_command` to generate the entire CPM/FetchContent string of `PATCH_COMMAND <....>` or an empty string. Therefore we only pass a `PATCH_COMMAND` arg when we truly have a patch.

That would mean a change to each `rapids_cpm_find` invocation in `rapids-cmake/cpm` but it should be pretty easy:
```PATCH_COMMAND ${patch_command}``` becomes `${patch_command}`

**Review by robertmaynard — APPROVED** [review]

File: testing/cpm/CMakeLists.txt [hunk 1]
```diff
@@ -23,6 +23,7 @@ add_cmake_config_test( cpm_find-existing-target )
 add_cmake_config_test( cpm_find-existing-target-to-export-sets )
 add_cmake_config_test( cpm_find-gtest-no-gmock )
 add_cmake_config_test( cpm_find-options-escaped )
+add_cmake_config_test( cpm_find-patch-command)
```
bdice, line 26: Consistency in spacing:
```suggestion
add_cmake_config_test( cpm_find-patch-command )
```

testing/cpm/cpm_find-patch-command/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
+#=============================================================================
+# Copyright (c) 2023-2024, NVIDIA CORPORATION.
```
bdice, line 2:
```suggestion
# Copyright (c) 2024, NVIDIA CORPORATION.
```

testing/cpm/cpm_find-patch-command/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
...
+#=============================================================================
+cmake_minimum_required(VERSION 3.23.1)
+project(rapids-cpm_find-patch-command-project LANGUAGES CXX)
+
+set(deps_dir "${CMAKE_CURRENT_BINARY_DIR}/_cccl_dep")
+if(NOT EXISTS "${deps_dir}")
+  file(MAKE_DIRECTORY "${deps_dir}")
+  find_package(Git)
+  execute_process(
+    COMMAND ${GIT_EXECUTABLE} clone --depth 1 --branch v2.2.0 https://github.com/NVIDIA/cccl.git
```
bdice, line 24: I expect that this test will fail unless we keep the `v2.2.0` in sync with the `versions.json`, because any other version will fail to apply the patches when it calls `rapids_cpm_cccl()`. Is that correct? Is there anything we can do to avoid needing to modify this test's CCCL version in the future?

vyasr: Yes, thanks for the reminder. I meant to fix this earlier, it's updated now.

bdice (approved):

jakirkham:
Thanks all! 🙏

PR #530, opened 2024-01-30
rapids-cmake can generate pinned versions file

robertmaynard:
## Description
Add rapids_cpm_generate_pinned_versions to support reproducible builds

rapids-cmake can now generate a `versions.json` that contains the exact git SHA1 used by depdendencies which can be re-used
to construct reproducible builds.

## Checklist


rapids-cmake/cpm/detail/gpv_root_dir_hook.cmake
```diff
@@ -0,0 +1,31 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#=============================================================================
+include_guard(GLOBAL)
+
+# Make sure we always have CMake 3.23 policies when executing this file Since we can be executing in
```
bdice, line 18:
```suggestion
# Make sure we always have CMake 3.23 policies when executing this file since we can be executing in
```

rapids-cmake/cpm/detail/gpv_root_dir_hook.cmake
```diff
@@ -0,0 +1,31 @@
...
+# directories of users of rapids-cmake which have a lower minimum cmake version and therefore
+# different policies
+#
+cmake_policy(PUSH)
+cmake_policy(VERSION 3.23)
+
+# Include the needed functions that write out the the pinned versions file
+include("${rapids-cmake-dir}/cpm/detail/gpv_write_file.cmake")
```
bdice, line 26: Could we use a name like `pinning` instead of `gpv`? `pinning_write_file`, `rapids_cpm_pinning_write_file`, etc. are clearer to me.

Alternatively, should we call this a "lockfile" and call the features "write lockfile", etc?

rapids-cmake/cpm/detail/gpv_write_file.cmake
```diff
@@ -0,0 +1,252 @@
...
+                    WORKING_DIRECTORY ${source_dir}
+                    ERROR_QUIET
+                    OUTPUT_VARIABLE _RAPIDS_URL
+                    OUTPUT_STRIP_TRAILING_WHITESPACE)
+    # Need to handle when we have applied N patch sets to the git repo and therefore the latest
+    # commit is just local
+    #
+    # Find all commits on our branch back to the common parent ( what we cloned )
+    #
+    execute_process(COMMAND ${GIT_EXECUTABLE} show-branch --current --sha1-name
```
bdice, line 41: Can we use `git rev-parse HEAD` to get the SHA? I don't understand the need for branch information here.

rapids-cmake/cpm/detail/gpv_write_file.cmake
```diff
@@ -0,0 +1,252 @@
...
+  endif()
+  # We start with a default template, and only add members that don't exist
+  string(CONFIGURE [=[{
+  "version" : "${CPM_PACKAGE_${package_name}_VERSION}",
+  ${url_string}
+  ${sha_string}
+  "git_shallow" : "false",
+  "always_download" : "true",
+  "trailing_place_holder" : "true"
+  },
```
bdice, line 106: Typically, JSON does not have whitespace before the colon. Linters would complain about this. This applies in several places.

Also we probably don't want the leading spaces before the closing `}`.
```suggestion
  string(CONFIGURE [=[{
  "version": "${CPM_PACKAGE_${package_name}_VERSION}",
  ${url_string}
  ${sha_string}
  "git_shallow": "false",
  "always_download": "true",
  "trailing_place_holder": "true"
},
```

rapids-cmake/cpm/detail/gpv_write_file.cmake
```diff
@@ -0,0 +1,252 @@
...
+    set(git_sha)
+    rapids_cpm_gpv_extract_source_git_info(${package} git_url git_sha)
+    rapids_cpm_gpv_add_json_entry(_rapids_entry ${package} git_url git_sha)
+    string(APPEND _rapids_json "${_rapids_entry}")
+  endforeach()
+
+  # To make everything easier we add a fake package to the end so we don't have any trailing ','
+  set(post_amble
+      [=[
+  "trailing_place_holder" : {}
```
bdice, line 240: This feels like a hack. Is this written to disk?

rapids-cmake/cpm/generate_pinned_versions.cmake
```diff
@@ -0,0 +1,107 @@
...
+  {
+    "version" : "<CPM_PACKAGE_<package_name>_VERSION>",
+    "git_url" : "<deduced>",
+    "git_tag" : "<deduced>",
+    "git_shallow" : "false",
+    "always_download" : "true"
+  }
+
+
+If the original package ( or override ) also had any `patches`, or `proprietary_binary`
```
bdice, line 64:
```suggestion
If the original package (or override) also had any `patches`, or `proprietary_binary`
```

rapids-cmake/cpm/generate_pinned_versions.cmake
```diff
@@ -0,0 +1,107 @@
...
+fields those will be propagated to the generated entry.
+
+.. note::
+  The git SHA1 computed for each package is found be finding the most recent
```
bdice, line 68:
```suggestion
  The git SHA1 computed for each package is found by finding the most recent
```

rapids-cmake/cpm/generate_pinned_versions.cmake
```diff
@@ -0,0 +1,107 @@
...
+  commit that can be cloned from the url.
```
bdice, line 69: I'm not sure if this is a good design. I think the better behavior would be to write out a pinning with the actual commit used, regardless of whether it exists on the remote. Pushing commits to the remote to ensure reproducibility is a reasonable responsibility for the user and would simplify the implementation of this feature. This might also be important to support multiple remotes, where a commit might only exist on a private fork or user's fork but not the upstream URL. Perhaps we can emit a warning if that SHA doesn't exist on the remote.

rapids-cmake/cpm/generate_pinned_versions.cmake
```diff
@@ -0,0 +1,107 @@
...
+    message(FATAL_ERROR "rapids_cpm_generate_pinned_versions requires an `OUTPUT` argument")
+  endif()
+
+  # Append the requested write path for `detail/write_pinned_versions.cmake`
+  set_property(GLOBAL APPEND PROPERTY rapids_cpm_generate_pin_files "${_RAPIDS_OUTPUT}")
+
+  get_property(already_hooked GLOBAL PROPERTY rapids_cpm_generate_pin_hook SET)
+  if(NOT already_hooked)
+    # install a hook that writes out the pinned versions at the end of the root level
+    # CMakeLists.cmake execution so we get all CPM packages added. Plus we can compute the paths
```
bdice, line 99: Is this right?
```suggestion
    # CMakeLists.txt execution so we get all CPM packages added. Plus we can compute the paths
```

rapids-cmake/cpm/init.cmake
```diff
@@ -25,13 +25,21 @@ Establish the `CPM` and preset package infrastructure for the project.
...
 
 The CPM module will be downloaded based on the state of :cmake:variable:`CPM_SOURCE_CACHE` and
 :cmake:variable:`ENV{CPM_SOURCE_CACHE}`. This allows multiple nested projects to share the
 same download of CPM. If those variables aren't set the file will be cached
 in the build tree of the calling project
 
+.. versionadded:: v24.04.00
+  As part of establishing rapids-cmake CPM the :cmake:command:`rapids_cpm_init` command
+  will call :cmake:command:`rapids_cpm_generate_pinned_versions` to automatically generate
+  `<CMAKE_BINARY_DIR>/rapids-cmake/pinned_versions.json` which will contain all the exact GIT SHAs
```
bdice, line 40:
```suggestion
  `<CMAKE_BINARY_DIR>/rapids-cmake/pinned_versions.json` which will contain all the exact git SHAs
```

rapids-cmake/cpm/init.cmake
```diff
@@ -40,14 +48,22 @@ in the build tree of the calling project
   If the override file doesn't specify a value or package entry the default
   version will be used.
 
+.. versionadded:: v24.04.00
+  ```
+  GENERATE_PINNED_VERSIONS
+  ```
+  Generates a json file with all cpm dependencies with pinned version values.
+  This allows for reproducible builds using the exact same state.
+  The follow will be located at `<CMAKE_BINARY_DIR>/rapids-cmake/pinned_versions.json`
```
bdice, line 57:
```suggestion
  The pinning file will be located at `<CMAKE_BINARY_DIR>/rapids-cmake/pinned_versions.json`
```

testing/cpm/cpm_generate_pins-format-patches/override.json
```diff
@@ -0,0 +1,33 @@
...
+          "fixed_in" : ""
+        }
+      ]
+    },
+    "rmm" : {
+      "version" : "22.02",
+      "git_tag" : "v22.02.00",
+      "patches" : [
+        {
+          "file" : "${current_json_dir}/patches/0001-Move-GIT-SHA1.patch",
```
bdice, line 21: Write `git` in lowercase everywhere.

testing/cpm/cpm_generate_pins-format-patches/patches/hide_kernels.patch
bdice, line 1: Can we use a patch that is less than 558 lines?

testing/export/write_language-multiple-nested-enables/CMakeLists.txt
```diff
@@ -23,11 +23,11 @@ project(write_language-multiple-nested-enable LANGUAGES NONE)
 # only installs one hook in `A` and `CMAKE_SOURCE_DIR`
 add_subdirectory(A)
 
-#Verify number of hooks is equal to 2
+#Verify number of hooks is equal to 3 (2 langs, 1 rapids init hook)
```
bdice, line 26:
```suggestion
# Verify number of hooks is equal to 3 (2 langs, 1 rapids init hook)
```

rapids-cmake/cpm/generate_pinned_versions.cmake
robertmaynard: We can't use the latest git commit. If the project has patches with the `.patch` suffix they are applied as git commits ( `git am` ). Therefore these patches will move `HEAD` and therefore that will not be clone-able when trying to reproduce the project.

> This might also be important to support multiple remotes, where a commit might only exist on a private fork or user's fork but not the upstream URL. Perhaps we can emit a warning if that SHA doesn't exist on the remote.

When using rapids-cmake, cpm, fetch_content, etc from CMake you only have one remote in a git cloned project. But even in the case the local user has changed that the `rapids_cpm_gpv_extract_source_git_info ` extracts the remote details for the current branch via the 'tracking' details

rapids-cmake/cpm/detail/gpv_write_file.cmake
robertmaynard: As mentioned in the above comment. We can't use `git rev-parse HEAD` since rapids-cmake supports applying `git format-patch` patches which cause local only commits to exist.

rapids-cmake/cpm/detail/gpv_write_file.cmake
robertmaynard: It is, since it makes `rapids_cpm_gpv_pretty_format_json` significantly easier to implement.
We could remove it but would need to do another string removal pass IIRC

rapids-cmake/cpm/detail/gpv_root_dir_hook.cmake
robertmaynard: I wanted to make it clear what user level function ( `generate_pinned_versions` ) that these two files are related too.
But felt using `generate_pinned_versions` in each file and function signature was too verbose. Maybe we just accept that cost...

Or we go with a shorter word like `pinning` for the detail side. @bdice thoughts?

testing/cpm/cpm_generate_pins-format-patches/patches/hide_kernels.patch
robertmaynard: Removed, I can re-use one of the patches being applied to rmm also to cuco to validate that we still compute the correct SHA1

rapids-cmake/cpm/detail/gpv_root_dir_hook.cmake
bdice: I'm okay with `pinning` for the detail side. That's more meaningful than `gpv`, in my opinion.

rapids-cmake/cpm/generate_pinned_versions.cmake
bdice: Okay. I'm not sure if I entirely understand this, but I trust your reasoning.

rapids-cmake/cpm/detail/gpv_write_file.cmake
robertmaynard: Updated the implementation to not need the `trailing_place_holder` for packages.

rapids-cmake/cpm/detail/pinning_write_file.cmake
```diff
@@ -0,0 +1,263 @@
...
+  if(NOT CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
+    return()
+  endif()
+
+  find_package(Git QUIET)
+
+  set(_rapids_json
+      [=[
+{
+"packages" : {
```
bdice, line 231: Can we remove the spaces that precede colons? This happens quite a few places.
```suggestion
"packages": {
```

rapids-cmake/cpm/detail/pinning_write_file.cmake
```diff
@@ -0,0 +1,263 @@
...
+    if(NOT value MATCHES "^(\\{|\\[)")
+      set(value "\"${value}\"")
+    endif()
+    string(CONFIGURE [=[
+    "${key}" : ${value},
+    "trailing_place_holder" : "true"
+  ]=]
+                     replacement_string)
+
+    string(REPLACE [=[  "trailing_place_holder" : "true"]=] "${replacement_string}" json_blob
```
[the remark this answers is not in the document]
bdice: It seems like we're jumping through some very complicated hoops just to manipulate strings and generate JSON with CMake. Is there a way we can use some other language (Python?), or a self-contained executable, or maybe a third-party JSON library for CMake to express this? Maybe I'm not the right reviewer for this... 😕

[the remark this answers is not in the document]
robertmaynard: Looking at how we can re-organize the json fragments to be valid when they enter `rapids_cpm_pinning_create_and_set_member` if we can do that, we can use CMake's built in `string(JSON` to add new members.
But currently the fragments aren't valid json due to trailing commas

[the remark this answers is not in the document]
robertmaynard: Updated to remove the need for `trailing_place_holder`

rapids-cmake/cpm/detail/pinning_write_file.cmake
```diff
@@ -0,0 +1,257 @@
...
+  # Insert a json key and value
+  #
+  function(rapids_cpm_pinning_create_and_set_member json_blob_var key value)
+
+    # if the first char in value is { or [ we don't need quotes
+    if(NOT value MATCHES "^(\\{|\\[)")
+      set(value "\"${value}\"")
+    endif()
+    # We need to quote 'value' so that it is a valid json element. Only ones not allowed to be
+    # unquted are numbers by the json spec
```
bdice, line 117: Strings must be quoted. Numbers like float `1.23`, or integer `456`, booleans like `true`, null values `null`, or arrays `[1, 2, 3]` aren't quoted. I assume arrays aren't relevant here, but booleans in particular might be important?
```suggestion
    # unquoted are numbers, booleans, or null by the JSON spec
```

rapids-cmake/cpm/detail/pinning_write_file.cmake
```diff
@@ -0,0 +1,257 @@
...
+      string(APPEND _rapids_entry [=[,
+]=])
+    else()
+      string(APPEND _rapids_entry [=[
+]=])
+    endif()
+    string(APPEND _rapids_json "${_rapids_entry}")
+  endforeach()
+
+  # To make everything easier we add a fake package to the end so we don't have any trailing ','
```
bdice, line 243: It looks like this comment is no longer needed.
```suggestion
```

rapids-cmake/cpm/generate_pinned_versions.cmake
```diff
@@ -0,0 +1,107 @@
...
+rapids-cmake dependencies from branch names to pinned values. This can be used in subsequent
+builds, e.g:
+
+  1. Have CI run with `versions.json` which tracks dependency by branch name
+  2. Store the generated pinned `versions.json` from the CI builds
+  3. If build is good, create the release branch and commit the generated pinned `versions.json`
+     to have reproducible builds for that release
+
+``OUTPUT``
+Specify a file path to write the pinned versions information to. Can be called multiple
```
bdice, line 48:
```suggestion
Specify a file path where the pinned versions information will be written. Can be called multiple
```

rapids-cmake/cpm/generate_pinned_versions.cmake
```diff
@@ -0,0 +1,107 @@
...
+The generated json file will have the following entries for each package:
+
+.. code-block:: json
+
+  {
+    "version": "<CPM_PACKAGE_<package_name>_VERSION>",
+    "git_url": "<deduced>",
+    "git_tag": "<deduced>",
+    "git_shallow": "false",
+    "always_download": "true"
```
bdice, line 60: I mentioned above that booleans should not be quoted -- in this example, the values will be interpreted as strings with contents `"false"` or `"true"`.

robertmaynard: Good point, working on an update to `rapids_cpm_pinning_create_and_set_member` to detect json types correctly so we know when to quote.

robertmaynard: Okay updatred `rapids_cpm_pinning_create_and_set_member` and it now works around the issues reported upstream at https://gitlab.kitware.com/cmake/cmake/-/issues/25716

bdice (approved):
Yay! Thanks so much for your hard work and many iterations. This has improved a lot from the initial draft. 🙇‍♂️

rapids-cmake/cpm/detail/pinning_write_file.cmake
```diff
@@ -0,0 +1,259 @@
...
+  ${url_string}
+  ${sha_string}
+  "git_shallow": false,
+  "always_download": true
+  }]=]
+                   pinned_json_entry)
+
+  # Insert a json key and value
+  #
+  function(rapids_cpm_pinning_create_and_set_member json_blob_var key value)
```
vyasr, line 110: AFAICT this function isn't using anything from the parent function's scope, is it? If not, can we unnest it? Happy to see it named with a leading underscore or something to mark it as "private" (I vaguely recall cmake-lint being smart enough not to complain about missing docs for those), but nesting functions generally negatively impacts readability IMO.

rapids-cmake/cpm/detail/pinning_write_file.cmake
```diff
@@ -0,0 +1,259 @@
...
+
+  set(url_string)
+  set(sha_string)
+  if(${url_var})
+    string(CONFIGURE [=["git_url": "${${url_var}}",]=] url_string)
+  endif()
+  if(${sha_var})
+    string(CONFIGURE [=["git_tag": "${${sha_var}}",]=] sha_string)
+  endif()
+  # We start with a default template, and only add members that don't exist
```
vyasr, line 98: I assume this is mostly for setting things like always_download, patches, or proprietary binary. Could we also set the URL/SHA? If so, then maybe this relates to my previous comment regarding the url/sha failure modes. If those could be overwritten, then maybe we need to error-check the final output JSON rather than erroring if the variables weren't set up front to allow the delayed set.

rapids-cmake/cpm/detail/pinning_write_file.cmake
```diff
@@ -0,0 +1,259 @@
...
+#[=======================================================================[.rst:
+rapids_cpm_pinning_pretty_format_json
+-------------------------------------
+
+
+.. versionadded:: v24.04.00
+
+Formats provided JSON to be easily read by humans
+#]=======================================================================]
+function(rapids_cpm_pinning_pretty_format_json _rapids_json_var)
```
vyasr, line 154: Since we're constructing our JSON string why can't we just construct it to be pretty from the start so that we don't need to prettify it afterwards?

vyasr (changes requested):
Do we want to make `rapids-cpm-init` respect an environment variable/CMake option so that users can do something like `cmake ... -DRAPIDS_CMAKE_GENERATE_PINNED_VERSIONS` (name TBD)? Or do we expect that consumers of this feature will modify their calls to `rapids_cpm_init`?

I haven't reviewed tests yet, will do that in a second pass after we get through this first round of questions.

rapids-cmake/cpm/generate_pinned_versions.cmake
```diff
@@ -0,0 +1,107 @@
...
+
+The generated json file will have the following entries for each package:
+
+.. code-block:: json
+
+  {
+    "version": "<CPM_PACKAGE_<package_name>_VERSION>",
+    "git_url": "<deduced>",
+    "git_tag": "<deduced>",
+    "git_shallow": false,
```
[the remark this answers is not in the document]
robertmaynard: Exactly

rapids-cmake/cpm/detail/pinning_write_file.cmake
```diff
@@ -0,0 +1,259 @@
...
+                    OUTPUT_VARIABLE _RAPIDS_SHA
+                    OUTPUT_STRIP_TRAILING_WHITESPACE)
+  endif()
+  # Only set the provided variables if we extracted the information
+  if(_RAPIDS_URL)
+    set(${git_url_var} "${_RAPIDS_URL}" PARENT_SCOPE)
+  endif()
+  if(_RAPIDS_SHA)
+    set(${git_sha_var} "${_RAPIDS_SHA}" PARENT_SCOPE)
+  endif()
```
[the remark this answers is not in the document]
robertmaynard: Projects like nvcomp will be in `CPM_PACKAGES` but won't have any git information so we can't hard error here.

rapids-cmake/cpm/detail/pinning_write_file.cmake
robertmaynard: Due to projects like `nvcomp` we can't presume that the 'cloned repository` will have any git information.

So what we are doing here is constructing the set of json entries that are most correct. So by not adding `git_tag`, and `git_url` we can extract them from the override or `version.json` entry. This simplifies the following logic since we can just iterate the override and default entries adding any elements that don't exist in our new entry.

rapids-cmake/cpm/detail/pinning_write_file.cmake
robertmaynard: Happy to hoist it up

rapids-cmake/cpm/detail/pinning_write_file.cmake
```diff
@@ -0,0 +1,259 @@
...
+
+#[=======================================================================[.rst:
+rapids_cpm_pinning_extract_source_git_info
+------------------------------------------
+
+.. versionadded:: v24.04.00
+
+
+#]=======================================================================]
+function(rapids_cpm_pinning_extract_source_git_info package git_url_var git_sha_var)
```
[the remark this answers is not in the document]
robertmaynard: Yes I was just adding the bare bones documentation. I think that documenting what the goals and inputs/outputs of these functions make sense given the complexity

rapids-cmake/cpm/detail/pinning_write_file.cmake
```diff
@@ -0,0 +1,259 @@
...
+
+.. versionadded:: v24.04.00
+
+This function will write out the pinned version info to the provided files when we are in the root
+CMakeLists.txt
+#]=======================================================================]
+function(rapids_cpm_pinning_write_file)
+  if(NOT CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
+    return()
+  endif()
```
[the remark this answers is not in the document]
robertmaynard: You are correct this can be removed. It was around from an earlier version where I hooked every directory instead of just hooking the root directory.

rapids-cmake/cpm/detail/pinning_root_dir_hook.cmake
```diff
@@ -0,0 +1,31 @@
...
+
+# Make sure we always have CMake 3.23 policies when executing this file since we can be executing in
+# directories of users of rapids-cmake which have a lower minimum cmake version and therefore
+# different policies
+#
+cmake_policy(PUSH)
+cmake_policy(VERSION 3.23)
+
+# Include the needed functions that write out the the pinned versions file
+include("${rapids-cmake-dir}/cpm/detail/pinning_write_file.cmake")
```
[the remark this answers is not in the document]
robertmaynard: We are guaranteed as `rapids-cmake-dir` is a cache variable so all directories can de-reference it.

rapids-cmake/cpm/detail/pinning_write_file.cmake
```diff
@@ -0,0 +1,259 @@
...
+                    WORKING_DIRECTORY ${source_dir}
+                    ERROR_QUIET
+                    OUTPUT_VARIABLE _RAPIDS_URL
+                    OUTPUT_STRIP_TRAILING_WHITESPACE)
+    # Need to handle when we have applied N patch sets to the git repo and therefore the latest
+    # commit is just local
+    #
+    # Find all commits on our branch back to the common parent ( what we cloned )
+    #
+    execute_process(COMMAND ${GIT_EXECUTABLE} show-branch --current --sha1-name
```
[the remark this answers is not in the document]
robertmaynard: > To be safe do we need to be specifying the refs that we want here

That is the problem right, we don't know what ref we want. The pinning logic extracts all CPM packages which can include projects not encoded in rapids-cmake ( a pattern seen in raft, legion, and others ). So we don't know what git tag was used to clone the repository. The `--current` argument to `show-branch` is our closest approximation as it adds the SHA1 of the active branch.

>  Are we guaranteed that the included history won't go up to the parent branch? For example, in my local clone of rapids-cmake here's what I see with the latest branch-24.04 and branch-24.04:

When CPM / CMake clone a repo the result should be that we have a detached head. In those cases we get a reduced output from `git show-branch`.
```
~/W/t/rapids-cmake ((de37eafa)👍 )  $ git commit -am "test" --allow-empty
[detached HEAD 7afc018] test
~/W/t/rapids-cmake ((7afc018f)👍 )  $ git show-branch --current --sha1-name
! [branch-24.04] Add support for Python 3.11 (#547)
 * [HEAD] test
--
 * [7afc018] test
+* [de37eaf] Add support for Python 3.11 (#547)
```

rapids-cmake/cpm/detail/pinning_write_file.cmake
robertmaynard: Each package fragment constructed in `rapids_cpm_pinning_add_json_entry` is considered to be a root element when we are using it. This means that the internal JSON engine to CMake will format it to an indentation level that is incorrect once we wrap them together in the `package` object.

What I can look at doing is seeing if we can get cmake to reformat everything for us, using a failed `SET` call or something instead of having a custom formatting engine.

robertmaynard: The formatting that CMake would generate would be:
```
{
  "packages" : 
  {
    "CCCL" : 
    {
      "always_download" : true,
      "git_shallow" : false,
      "git_tag" : "36f379f29660761fe033a1306ca9dab6a88cb65c",
      "git_url" : "https://github.com/NVIDIA/cccl.git",
      "patches" : 
      [
....
```

Are we happy with that style of JSON compared to the more compact version that we have in `versions.json`. Personally I am happy with it since it would remove the need to have a custom ( and hacky ) pretty printer.

bdice: That’s fine. I don’t like the format but prefer not having the extra code for pretty print.

rapids-cmake/cpm/detail/pinning_write_file.cmake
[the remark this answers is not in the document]
vyasr: Right forgot we'd be in detached head mode. That makes sense.

rapids-cmake/cpm/detail/pinning_write_file.cmake
vyasr: That makes sense, so my main concern is really, is there a possibility that we could encoded a pinned version for a dependency that we later don't know how to acquire? I guess not, because if we were able to get it in the first place we must have done so somehow (be it via git, a download URL, or something else) and that same CMake logic is expected to be available at the time we use the pinning file, so there's no world where we get a pinning file with a dependency that we don't know how to acquire. Is that correct?

rapids-cmake/cpm/detail/pinning_write_file.cmake
vyasr: Yeah I'm in the same boat, I don't think the improved formatting is worth the extra code.

rapids-cmake/cpm/detail/pinning_write_file.cmake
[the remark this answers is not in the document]
vyasr: This comment makes it seem like you intended to add documentation, but there are still a few functions in this file (including this one) without any description.

rapids-cmake/cpm/detail/pinning_write_file.cmake
```diff
@@ -0,0 +1,206 @@
...
+      string(APPEND _rapids_entry [=[,
+]=])
+    else()
+      string(APPEND _rapids_entry [=[
+]=])
+    endif()
+    string(APPEND _rapids_json "${_rapids_entry}")
+  endforeach()
+
+  set(post_amble [=[
```
vyasr, line 193: Maybe `epilogue` so that we can use a real word :)

testing/utils/fill_cache/CMakeLists.txt
```diff
@@ -31,7 +31,7 @@ include(${rapids-cmake-dir}/cpm/spdlog.cmake)
 include(${rapids-cmake-dir}/cpm/fmt.cmake)
 include(${rapids-cmake-dir}/cpm/thrust.cmake)
 
-rapids_cpm_init()
+rapids_cpm_init(GENERATE_PINNED_VERSIONS)
```
vyasr, line 34: Will choosing to always generate the pin file affect any other tests? Also why do we need this when we should be setting this on a per-test basis for tests that call `rapids_cpm_init`

testing/other/CMakeLists.txt
```diff
@@ -13,8 +13,8 @@
...
 #=============================================================================
-add_cmake_config_test( rapids_cmake-multiple-cpm )
-add_cmake_config_test( rapids_cmake-multiple-simple )
+add_cmake_config_test( rapids_cmake-multiple-cpm NO_RAPIDS_CMAKE_HOOKS)
+add_cmake_config_test( rapids_cmake-multiple-simple NO_RAPIDS_CMAKE_HOOKS)
 
-add_cmake_config_test( FetchContent-legacy )
-add_cmake_config_test( FetchContent-hostile-legacy )
+add_cmake_config_test( FetchContent-legacy NO_RAPIDS_CMAKE_HOOKS)
+add_cmake_config_test( FetchContent-hostile-legacy NO_RAPIDS_CMAKE_HOOKS)
```
vyasr, line 20: Why are we now disabling these hooks? We used to do the fetch emulation for every test IIUC, why remove it for these now?

testing/cpm/cpm_generate_pins-nested/verify/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+
+set(projects rmm fmt spdlog CCCL cuco)
+# Get the baseline before the override is added
+foreach(proj IN LISTS ${projects})
+  rapids_cpm_package_details(${proj} ${proj}_version ${proj}_repository ${proj}_tag ${proj}_shallow ${proj}_exclude)
+endforeach()
+
+
+file(READ "${CMAKE_CURRENT_BINARY_DIR}/../rapids-cmake/pinned_versions.json" json_data)
+message(STATUS "${json_data}")
```
vyasr, line 33: Debug print?

testing/cpm/cpm_generate_pins-nested/verify/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+  rapids_cpm_package_details(${proj} ${proj}_version ${proj}_repository pin_${proj}_tag ${proj}_shallow ${proj}_exclude)
+  if(pin_${proj}_tag STREQUAL ${proj}_tag)
+    message(FATAL_ERROR "pinned ${proj} tag (${pin_${proj}_tag}) should differ compared to baseline ${${proj}_tag}")
+  endif()
+
+  # Everything should have shallow marked as false
+  # so that clones by SHA1 work
+  if(${proj}_shallow)
+    message(FATAL_ERROR "${proj}_shallow} is expected to be false, but got ${${proj}_shallow}")
+  endif()
```
vyasr, line 49: Out of scope for this PR, but given the amount of duplication required to write CMake tests it might be good for us to centralize helpers that get used in multiple tests.

testing/cpm/cpm_generate_pins-nested/verify/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+cmake_minimum_required(VERSION 3.23.1)
+project(rapids-test-project LANGUAGES CXX)
+
+include(${rapids-cmake-dir}/cpm/init.cmake)
+include(${rapids-cmake-dir}/cpm/package_override.cmake)
+include(${rapids-cmake-dir}/cpm/detail/package_details.cmake)
+
+rapids_cpm_init()
+
+set(projects rmm fmt spdlog CCCL cuco)
```
vyasr, line 25: Minor suggestion: to reduce hardcoding (and the risk of future error) you could write the set of packages to load in each subdirectory in the top-level lists file then pass that as a variable when building verify to guarantee they don't get out of sync.

testing/cpm/cpm_generate_pins-format-patches/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+rapids_cpm_rmm(DOWNLOAD_ONLY ON)
+
+include(${rapids-cmake-dir}/cpm/cuco.cmake)
+rapids_cpm_cuco(DOWNLOAD_ONLY ON)
+
+# Verify that the two files that we inserted into the RMM source tree exist
+# Which proves the patches in the override are properly applied
+if(NOT EXISTS "${rmm_SOURCE_DIR}/git_file_1.txt")
+  message(FATAL_ERROR "failed to apply rmm first patch")
+endif()
```
vyasr, line 35: Should we also check this file for cuco since you applied the same patch to it?

vyasr (changes requested):
[My question on the overall review](https://github.com/rapidsai/rapids-cmake/pull/530#pullrequestreview-1906892854) (not tied to any line) from the last review remains outstanding, but I don't think it's super important. The other remaining items are pretty small and we should be good to finish this up in one more round.

rapids-cmake/cpm/detail/pinning_write_file.cmake
robertmaynard: > so there's no world where we get a pinning file with a dependency that we don't know how to acquire. Is that correct?

Correct!

rapids-cmake/cpm/detail/pinning_write_file.cmake
robertmaynard: Just going to drop the variable entirely

testing/cpm/cpm_generate_pins-nested/verify/CMakeLists.txt
robertmaynard: Yuuup. Removed

testing/utils/fill_cache/CMakeLists.txt
robertmaynard: > Will choosing to always generate the pin file affect any other tests?

No,  this only generates a pin file when we check out the local source cache of the dependencies. No effect on the tests, was done to have as much coverage on the pinning logic as possible. Not required for anything, so can remove if we choose.

testing/cpm/cpm_generate_pins-format-patches/CMakeLists.txt
robertmaynard: Yes we should! This was missed since we changed the format-patch we applied to cuco and previously it didn't add new files to the repository.

rapids-cmake/cpm/detail/pinning_write_file.cmake
robertmaynard: Looking at this more, I believe we need to handle the use case when `CPM` finds a local version of a package ( e.g. fmt ) instead of building from source. In those cases `CPM_PACKAGES` will still list `fmt`, but it won't have a source dir.

I will need to extend `rapids_cpm_pinning_write_file` to handle checking that `CPM` actually added a CMake `FetchContent` record for the project before we try and construct a JSON entry for it.

rapids-cmake/cpm/detail/pinning_write_file.cmake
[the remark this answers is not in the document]
robertmaynard: Updated the functions to actually have documentation

testing/other/CMakeLists.txt
robertmaynard: The hooks this is disabling is the emulation that rapids-cmake has already been fetched ( added by testing/utils/emulate_fetching_rapids_cmake.cmake ).

Previous to this PR, directories would not have any rapids-cmake hook emulation automatically added, while simple `.cmake` files would.  

In fixing this oversight  needed to disable this hook emulation for these tests as they can verify that can actually fetch rapids-cmake in the legacy manner.

testing/cpm/cpm_generate_pins-nested/verify/CMakeLists.txt
robertmaynard: Updated the three new tests to do that. Also added in a check to ensure the caller sets the variable so we don't check zero projects.

rapids-cmake/cpm/detail/pinning_write_file.cmake
robertmaynard: Corrected this bug, added a test so we don't regress it.

vyasr: OK yes this kind of case was making me a bit uneasy. So the assumption in this case is that users of the pinned file will have the same version installed locally as whenever the pin file was created in order to fully reproduce the build. Should we include some information in the file to indicate that a dependency was found? How do we want to communicate this to users? For cases where complete reproducibility is required should we recommend setting `CPM_DOWNLOAD_ALL`?

robertmaynard: > So the assumption in this case is that users of the pinned file will have the same version installed locally as whenever the pin file was created in order to fully reproduce the build

Correct. The same assumption is applied for any existing `find_package` or `rapids_find_package` call as well. The end goal here is to pin packages we are building by source. 

In theory the creation of this file is done in a controlled env, like CI which has `CPM_DOWNLOAD_ALL` set so that we have a reproducible baseline env. From that point we just need to do the transformation from branches to exact SHA1. But that might not always be the case.

> Should we include some information in the file to indicate that a dependency was found? How do we want to communicate this to users? For cases where complete reproducibility is required should we recommend setting CPM_DOWNLOAD_ALL?

We currently don't have a way to represent an informative package in the json file. So I am more inclined to have the function output a message stating what packages it failed to record and advise people to also set `CPM_DOWNLOAD_ALL`.

I think that in a future PR we can extend the `version.json` to have the concept of `CPM_DOWNLOAD_ALL` which would allow for a declarative declaration

vyasr: OK cool I agree with that plan. I can approve once we have that messaging in place.

vyasr (approved):

rapids-cmake/cpm/detail/pinning_write_file.cmake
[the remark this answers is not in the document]
KyleFromNVIDIA: This command is causing problems when we clone a detached hash. Git is "guessing" which branch is the "base branch" and getting it incorrect, resulting in the wrong commit being picked here.

What's stopping us from doing the following?

```
git merge-base HEAD ${git_tag}
```

PR #544, opened 2024-02-26
Remove unneeded whitespace from json

robertmaynard:
## Description
While the JSON spec allows whitespace before or after any structural character, some json linters complain about it. So we remove the 'offending' whitespace.
## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,6 +15,8 @@ repos:
             ^rapids-cmake/cpm/patches/.*
           )
       - id: check-json
+      - id: pretty-format-json
+        args: ["--autofix", "--no-sort-keys"]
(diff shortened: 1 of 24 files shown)


KyleFromNVIDIA (approved):
What program did you use to find these? I think we'd do well to add it as a pre-commit hook.

robertmaynard:
> What program did you use to find these? I think we'd do well to add it as a pre-commit hook.

Brought up as a style issue during https://github.com/rapidsai/rapids-cmake/pull/530 and I manually found this with grep.

@bdice Is looking at adding the pre-commit hook currently

bdice:
@KyleFromNVIDIA I added a hook called `pretty-format-json`. It seems to do the job we want.

AyodeAwe (approved):

PR #558, opened 2024-03-12
rapids_cpm_package_details now validates required entries exist

robertmaynard:
## Description
rapids-cmake now provides proper error message reports when the required json entry fields don't exist in either the root `versions.json` or the package override.
## Checklist


bdice (approved):

rapids-cmake/cpm/detail/package_details.cmake
```diff
@@ -64,6 +64,17 @@ function(rapids_cpm_package_details package_name version_var url_var tag_var sha
   rapids_cpm_json_get_value(git_url)
   rapids_cpm_json_get_value(git_tag)
 
+  # Only do validation if we have an entry
+  if(json_data OR override_json_data)
+    # Validate that we have the required fields
+    foreach(var IN ITEMS version git_url git_tag)
+      if(NOT ${var})
```
bdice, line 71: Would an empty string `""` be allowed? (I realized this as I clicked approve -- the test file was using `"empty"`.)

robertmaynard: No that won't be allowed with the current implementation, since an empty string is a false value.

We want something in these entries to enforce 'proper' version.json entries, so I am okay with this middle ground.

bdice: Okay, that's fine! Thanks for clarifying.

PR #560, opened 2024-03-12
Deprecate rapids_cpm_libcudacxx and rapids_cpm_thrust.

bdice:
## Description
Some users of Thrust have been confused about whether to use `rapids_cpm_thrust` or `rapids_cpm_cccl`. Deprecating the old CPM functions for Thrust and libcudacxx would make the messaging clearer that we intend to support primarily CCCL going forward.

It's fine if we need to extend the deprecation period for an arbitrary length of time -- but adding this notice will help push users to adopt the new CCCL monorepo.

## Checklist

### DIFF
--- a/rapids-cmake/cpm/libcudacxx.cmake
+++ b/rapids-cmake/cpm/libcudacxx.cmake
@@ -27,6 +27,11 @@ tracking of these dependencies for correct export support.
 Uses the version of libcudacxx :ref:`specified in the version file <cpm_versions>` for consistency
 across all RAPIDS projects.
 
+.. deprecated:: v24.04.00
+  ``rapids_cpm_thrust`` uses Thrust 1.x. Users should migrate to
+  ``rapids_cpm_cccl`` which uses CCCL 2.x, including new versions of Thrust,
+  CUB, and libcudacxx.
+
 .. code-block:: cmake
 
   rapids_cpm_libcudacxx( [BUILD_EXPORT_SET <export-name>]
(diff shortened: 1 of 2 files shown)


bdice:
@robertmaynard I'd appreciate your thoughts on this. I don't have a strong timeline in mind (this could slip to 24.06) but it would help to start communicating to users that `rapids_cpm_cccl` is the direction we want them to go.

robertmaynard:
> @robertmaynard I'd appreciate your thoughts on this. I don't have a strong timeline in mind (this could slip to 24.06, or we could defer on removal for a longer time) but it would help to start communicating to users that `rapids_cpm_cccl` is the direction we want them to go.

I agree that we need a to deprecate this for a couple of releases so that all consumers are ready for the removal. I like starting in 24.04 and yanking in 24.10

rapids-cmake/cpm/libcudacxx.cmake
```diff
@@ -27,6 +27,11 @@ tracking of these dependencies for correct export support.
 Uses the version of libcudacxx :ref:`specified in the version file <cpm_versions>` for consistency
 across all RAPIDS projects.
 
+.. deprecated:: v24.04.00
+  ``rapids_cpm_thrust`` uses Thrust 1.x. Users should migrate to
```
bdice, line 31: Fix library name.
```suggestion
  ``rapids_cpm_libcudacxx`` uses libcudacxx 2.1.0. Users should migrate to
```

PR #570, opened 2024-03-25
nvcomp try proprietary binary when 'always_download' is on

robertmaynard:
## Description

The nvcomp package will ignore the downloaded proprietary binary, and build from source when `"always_download" : true` is in the `versions.json` file.

This was found when using a pinned override file, as it failed to use the correct nvcomp version. 

Merging this unblocks spark-rapids

## Checklist

### DIFF
--- a/rapids-cmake/cpm/nvcomp.cmake
+++ b/rapids-cmake/cpm/nvcomp.cmake
@@ -112,7 +112,7 @@ function(rapids_cpm_nvcomp)
     if(nvcomp_url)
       rapids_cpm_download_proprietary_binary(nvcomp ${nvcomp_url})
 
-      # unset CPM_DOWNLOAD_ALL if we have a proprietary binary enabled this ensure we actually use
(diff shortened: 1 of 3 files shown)


rapids-cmake/cpm/nvcomp.cmake
```diff
@@ -111,6 +111,10 @@ function(rapids_cpm_nvcomp)
     rapids_cpm_get_proprietary_binary_url(nvcomp ${version} nvcomp_url)
     if(nvcomp_url)
       rapids_cpm_download_proprietary_binary(nvcomp ${nvcomp_url})
+
+      # unset CPM_DOWNLOAD_ALL if we have a proprietary binary enabled this ensure we actually use
```
bdice, line 115: This is two separate sentences, right?
```suggestion
      # Unset CPM_DOWNLOAD_ALL if we have a proprietary binary enabled. This ensures we actually use
```

bdice (approved):

PR #579, opened 2024-04-09
`rapids_cython_create_modules()`: Generate Cython dependency file

Jacobfaib:
## Description

rapids_cython_create_modules() should instruct Cython to create dependency files for each `.pyx`.

## Checklist

### DIFF
--- a/rapids-cmake/cython-core/create_modules.cmake
+++ b/rapids-cmake/cython-core/create_modules.cmake
@@ -1,5 +1,5 @@
 # =============================================================================
-# Copyright (c) 2023, NVIDIA CORPORATION.
+# Copyright (c) 2023-2024, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
 # in compliance with the License. You may obtain a copy of the License at
--- a/rapids-cmake/cython-core/create_modules.cmake
+++ b/rapids-cmake/cython-core/create_modules.cmake
@@ -106,6 +106,7 @@ function(rapids_cython_create_modules)
     cmake_path(GET cython_filename FILENAME cython_module)
     cmake_path(REPLACE_EXTENSION cython_module "${_ext}" OUTPUT_VARIABLE cpp_filename)
     cmake_path(REMOVE_EXTENSION cython_module)
+    cmake_path(SET depfile NORMALIZE "${CMAKE_CURRENT_BINARY_DIR}/${cpp_filename}.dep")
 
     # Save the name of the module without the provided prefix so that we can control the output.
     set(cython_module_filename "${cython_module}")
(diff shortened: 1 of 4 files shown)


rapids-cmake/cython-core/create_modules.cmake
```diff
@@ -117,9 +118,10 @@ function(rapids_cython_create_modules)
     add_custom_command(OUTPUT ${cpp_filename}
                        DEPENDS ${cython_filename}
                        VERBATIM
-                       COMMAND "${CYTHON}" ARGS "${_language_flag}" -3 ${CYTHON_FLAGS_LIST}
```
Jacobfaib, line 120: Note this change is technically unrelated to the dependency file feature, however, without it I would get:
```sh
[1/3] Transpiling test.pyx to test.c
FAILED: test.c /Users/jfaibussowit/soft/nv/rapids-cmake/testing/cython-core/cython-core_create_modules/build/test.c
.. && /Users/jfaibussowit/soft/packages/base_venv/bin/cython "" -3 --directive ...
/Users/jfaibussowit/soft/packages/base_venv/bin/cython: No such file or directory: ''
```
Note the `""` argument where `${_language_flag}` should be since:
```cmake
set(_ext ".c")
set(_language_flag "")
if(_RAPIDS_CYTHON_CXX) # from function arguments
  set(_language_flag "--cplus")
  set(_ext ".cxx")
endif()
```
Not sure how CI did not catch this error previously @robertmaynard?

robertmaynard: presumably `_RAPIDS_CYTHON_CXX` is always set in all of our tests, so we never ran into the error.

Jacobfaib: `_RAPIDS_CYTHON_CXX` is due to the preceding `cmake_parse_arguments()` call. The test does not pass the `CXX` argument, so `_RAPIDS_CYTHON_CXX` is always false.

It might be that this test is never run?

vyasr: I think that you're both partially right. The tests are being run in CI, but [these tests are all added with `add_cmake_config_test`](https://github.com/rapidsai/rapids-cmake/blob/branch-24.06/testing/cython-core/CMakeLists.txt#L22), which means only the CMake configure happens, not the build. We can change those to `add_cmake_build_test`. I can reproduce the error you observe when I add that, and your update fixes that. In practice all of RAPIDS passes `CXX` so we don't actually ever use this API without the flag in practice.

I am not sure how you were observing this failure locally without making that change, but good find!

vyasr (commented):
I'm amazed that CMake makes it this easy to do this. Good to know. @jacobfaib it looks like you need to fix up style then we should be good. You can do that by (installing pre-commit if you haven't via e.g. `pip install pre-commit` and then) running `pre-commit run --all-files` inside your clone of rapids-cmake.

robertmaynard (approved):

rapids-cmake/cython-core/create_modules.cmake
Jacobfaib: > I am not sure how you were observing this failure locally without making that change, but good find!

I just did a `cmake -S . -B build` then `cmake --build build` inside the directory containing these files :)

vyasr (approved):

rapids-cmake/cython-core/create_modules.cmake
```diff
@@ -117,9 +118,10 @@ function(rapids_cython_create_modules)
     add_custom_command(OUTPUT ${cpp_filename}
```
henryiii, line 118: Shouldn't `${depfile}` be here too?

jcfr: Depfile support has been added through https://github.com/scikit-build/cython-cmake/commit/490f1998147af498793f88365b2a9293311a74ee

PR #588, opened 2024-04-19
Ensure support_conda_env uses `isystem`

robertmaynard:
## Description
Conda injects the path to the conda include directory as a system include when building via `CXXFLAGS` / `CFLAGS`  / etc. We need to match this behavior so that we have consistent search order, since system includes are used last compared to user includes.

## Checklist


trxcllnt (approved):

vyasr (approved):
If using the conda compilers this change will be redundant, right?

PR #590, opened 2024-04-22
clarify language around how FetchContent is used in RAPIDS.cmake

jameslamb:
## Description

Reading through, https://docs.rapids.ai/api/rapids-cmake/stable/basics/, I found this sentence a bit confusing:

> *The rapids-cmake module is designed to be acquired via CMake’s [Fetch Content](https://cmake.org/cmake/help/latest/module/FetchContent.html) into your project.*

Because the code block immediately below it does not actually call `FetchContent()` and friends.

image

This PR proposes a change to that language which I think might make it clearer that `RAPIDS.cmake` is itself using `FetchContent()`.

## Checklist

### DIFF
--- a/docs/basics.rst
+++ b/docs/basics.rst
@@ -5,8 +5,9 @@ RAPIDS-CMake Basics
 Installation
 ************
 
-The ``rapids-cmake`` module is designed to be acquired via CMake's `Fetch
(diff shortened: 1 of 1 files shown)


KyleFromNVIDIA (approved):

robertmaynard (approved):
Thanks for the update

Issue #587, opened 2024-04-18
[FEA] Allow providing a different default `versions.json`


trxcllnt:
**Is your feature request related to a problem? Please describe.**
Need the ability to specify the default `versions.json` file used by a build, rather than always using the one in rapids-cmake.

**Describe the solution you'd like**
* Support `rapids_cpm_init(DEFAULT_VERSION_FILE <file>)`
* Support `-DRAPIDS_CMAKE_CPM_DEFAULT_FILE=<file>`

**Describe alternatives you've considered**
Right now we have to fork rapids-cmake, modify `versions.json`, and configure each project with custom `-Drapids-cmake-repo=` and `-Drapids-cmake-branch=` arguments.

**Additional context**
This is necessary for doing CCCL upgrades, modifying versions of dependencies built in DLFW, etc.

robertmaynard:
We should do this feature and https://github.com/rapidsai/rapids-cmake/issues/552 at the same time so we have symmetry

vyasr:
Could you remind me why using an override json isn't sufficient here?

robertmaynard:
Closing as this feature is now complete with #596 being merged.

PR #595, opened 2024-04-26
Expand rapids_cpm_init to support custom default version files

robertmaynard:
## Description
First step to implelementing https://github.com/rapidsai/rapids-cmake/issues/587.

Exposes a new `CUSTOM_DEFAULT_VERSION_FILE` option to `rapids_cpm_init` that allows projects to specify a new default versions.json to be used instead of the one packaged with rapids-cmake

## Checklist

### DIFF
--- a/rapids-cmake/cpm/detail/load_preset_versions.cmake
+++ b/rapids-cmake/cpm/detail/load_preset_versions.cmake
@@ -1,5 +1,5 @@
 #=============================================================================
-# Copyright (c) 2021, NVIDIA CORPORATION.
+# Copyright (c) 2021-2024, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -25,7 +25,7 @@ Establish the `CPM` preset package information for the project.
 
 .. code-block:: cmake
 
-  rapids_cpm_load_preset_versions()
+  rapids_cpm_load_preset_versions([PRESET_FILE version_file])
 
 .. note::
   Will be called by the first invocation of :cmake:command:`rapids_cpm_init` or :cmake:command:`rapids_cpm_<pkg>`.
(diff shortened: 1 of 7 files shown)


rapids-cmake/cpm/init.cmake
```diff
@@ -34,11 +35,24 @@ The CPM module will be downloaded based on the state of :cmake:variable:`CPM_SOU
...
+
+  Using a built-in rapids-cmake package without an explicit entry in the custom default version file is undefined behavior.
+
+  If multiple calls to `rapids_cpm_init` occur with different default version files being used,
+  each version file will be loaded. If a project is listed in multiple default version files, the first
+  file values will be used, and all later calls for that packaged will be ignored.  This "first to record, wins"
+  approach is used to match FetchContent, and allows parent projects to override child
+  projects.
+
+  The provided json file must follow the `versions.json` format, which is :ref:`documented here<cpm_version_format>`.
```
KyleFromNVIDIA, line 55:
```suggestion
  This is an advanced option that allows projects to specify a custom default ``versions.json`` file that will
  be used instead of the one packaged inside rapids-cmake. Since this allows you to specify a new default
  ``versions.json`` it must contain information for all the built-in rapids-cmake packages ( cccl, fmt, rmm, etc )
  that are used by your project.

  Using a built-in rapids-cmake package without an explicit entry in the custom default version file is undefined behavior.

  If multiple calls to ``rapids_cpm_init`` occur with different default version files being used,
  each version file will be loaded. If a project is listed in multiple default version files, the first
  file values will be used, and all later calls for that packaged will be ignored.  This "first to record, wins"
  approach is used to match ``FetchContent``, and allows parent projects to override child
  projects.

  The provided json file must follow the ``versions.json`` format, which is :ref:`documented here<cpm_version_format>`.
```

rapids-cmake/cpm/init.cmake
```diff
@@ -47,6 +61,12 @@ in the build tree of the calling project
   and :cmake:module:`FetchContent() <cmake:module:FetchContent>` package. By providing a secondary
   file with extra`CPM` preset package information for the project.
 
+  If multiple calls to `rapids_cpm_init` occur with different OVERRIDE files being used,
+  each version file will be loaded. If a project is listed in multiple override files, the first
+  file values will be used, and all later calls for that packaged will be ignored.  This "first to record, wins"
+  approach is used to match FetchContent, and allows parent projects to override child
+  projects.
```
KyleFromNVIDIA, line 68:
```suggestion
  If multiple calls to ``rapids_cpm_init`` occur with different ``OVERRIDE`` files being used,
  each version file will be loaded. If a project is listed in multiple override files, the first
  file values will be used, and all later calls for that packaged will be ignored.  This "first to record, wins"
  approach is used to match ``FetchContent``, and allows parent projects to override child
  projects.
```

rapids-cmake/cpm/detail/load_preset_versions.cmake
```diff
@@ -34,15 +34,32 @@ Establish the `CPM` preset package information for the project.
...
+
   # Check if we have been loaded before, if so early terminate
-  get_property(already_loaded GLOBAL PROPERTY rapids_cpm_load_presets SET)
+  get_property(already_loaded GLOBAL PROPERTY rapids_cpm_load_presets_${_rapids_preset_version_file}
+               SET)
   if(already_loaded)
     return()
   endif()
-  set_property(GLOBAL PROPERTY rapids_cpm_load_presets "ON")
+  set_property(GLOBAL PROPERTY rapids_cpm_load_presets_${_rapids_preset_version_file} "ON")
```
KyleFromNVIDIA, line 59: One thing to consider here is that the same file might be accessed by different relative paths, and thus will set two different properties. If this is merely a cache for optimization then it might not matter.

KyleFromNVIDIA (changes requested):
Looks good for the most part, but I have a few suggestions regarding formatting and one question about global properties.

KyleFromNVIDIA (approved):

PR #602, opened 2024-05-07
Ensure nvcomps build and install layouts are consistent

robertmaynard:
## Description
Corrects a bug where the nvcomp packages from the build directory would reference the `lib64` layout in the cmake file since that would be the correct layout at install time but not build time.

We now transform the layout of nvcomp to be consistent between build and install

## Checklist

### DIFF
--- a/rapids-cmake/cpm/nvcomp.cmake
+++ b/rapids-cmake/cpm/nvcomp.cmake
@@ -111,19 +111,37 @@ function(rapids_cpm_nvcomp)
     rapids_cpm_get_proprietary_binary_url(nvcomp ${version} nvcomp_url)
     if(nvcomp_url)
       rapids_cpm_download_proprietary_binary(nvcomp ${nvcomp_url})
-
-      # Unset CPM_DOWNLOAD_ALL if we have a proprietary binary enabled. This ensures we actually use
-      # the proprietary binary
-      unset(CPM_DOWNLOAD_ALL)
     endif()
 
-    # Record the nvcomp_DIR so that if USE_PROPRIETARY_BINARY is disabled we can safely clear the
-    # nvcomp_DIR value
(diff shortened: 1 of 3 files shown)


ttnghia (approved):

rapids-cmake/cpm/nvcomp.cmake
```diff
@@ -111,19 +111,37 @@ function(rapids_cpm_nvcomp)
...
+      set(nvcomp_proprietary_binary_dir "${nvcomp_ROOT}/${lib_dir}/cmake/nvcomp")
       cmake_path(NORMAL_PATH nvcomp_proprietary_binary_dir)
       set(rapids_cpm_nvcomp_proprietary_binary_dir "${nvcomp_proprietary_binary_dir}"
           CACHE INTERNAL "nvcomp proprietary location")
+
+      # Enforce that we need to find the local download of nvcomp and nothing else when we have a
+      # proprietary binary enabled.
+      list(APPEND CMAKE_PREFIX_PATH "${nvcomp_ROOT}/${lib_dir}/cmake/nvcomp")
+      unset(CPM_DOWNLOAD_ALL)
+      set(CPM_USE_LOCAL_PACKAGES ON)
```
vyasr, line 144: This isn't a cache variable, right? i.e. setting this here will only affect the `rapids_cpm_find` call below, and not any calls in consumers from before that? I'm assuming that there are no concerns about transitive finds in nvcomp itself (or at least, we can live with that if we are).

vyasr (approved):
One small question, otherwise looks good.

rapids-cmake/cpm/nvcomp.cmake
robertmaynard: That is all correct. Local variable so it will only effect `nvcomp` and no concerns about transitive finds

PR #606, opened 2024-05-10
Add rapids_cpm_nvtx3.

bdice:
## Description
Adds `rapids_cpm_nvtx3`, adapted from the code that is currently implemented in libcudf. This aims to help with https://github.com/rapidsai/rmm/pull/1558.

## Checklist


bdice:
Hmm. I am not sure if `INSTALL_EXPORT_SET` works as desired because it doesn't look like NVTX defines any install rules? Maybe I am missing some piece of knowledge about CMake default behaviors.

docs/api.rst
```diff
@@ -59,6 +59,7 @@ package uses :ref:`can be found here. <cpm_versions>`
    /packages/rapids_cpm_libcudacxx
    /packages/rapids_cpm_nvbench
    /packages/rapids_cpm_nvcomp
+   /packages/rapids_cpm_nvtx
```
bdice, line 62:
```suggestion
   /packages/rapids_cpm_nvtx3
```

rapids-cmake/cpm/nvtx3.cmake
```diff
@@ -0,0 +1,107 @@
...
+#]=======================================================================]
+function(rapids_cpm_nvtx3)
+  list(APPEND CMAKE_MESSAGE_CONTEXT "rapids.cpm.nvtx3")
+
+  set(options)
+  set(one_value USE_PROPRIETARY_BINARY BUILD_EXPORT_SET INSTALL_EXPORT_SET)
+  set(multi_value)
+  cmake_parse_arguments(_RAPIDS "${options}" "${one_value}" "${multi_value}" ${ARGN})
+
+  # Fix up _RAPIDS_UNPARSED_ARGUMENTS to have EXPORT_SETS as this is need for rapids_cpm_find
```
KyleFromNVIDIA, line 59: Does this mean `_RAPIDS_UNPARSED_ARGUMENTS` is a magic variable that `rapids_cpm_find()` reads to figure out which arguments it should ignore?

rapids-cmake/cpm/nvtx3.cmake
```diff
@@ -0,0 +1,107 @@
...
+  endif()
+  if(_RAPIDS_BUILD_EXPORT_SET)
+    list(APPEND _RAPIDS_EXPORT_ARGUMENTS BUILD_EXPORT_SET ${_RAPIDS_BUILD_EXPORT_SET})
+  endif()
+  set(_RAPIDS_UNPARSED_ARGUMENTS ${_RAPIDS_EXPORT_ARGUMENTS})
+
+  set(to_install OFF)
+  if(INSTALL_EXPORT_SET IN_LIST ARGN)
+    set(to_install ON)
+  endif()
```
KyleFromNVIDIA, line 71: We can get rid of this block if we just condition the block below on `_RAPIDS_INSTALL_EXPORT_SET`.

rapids-cmake/cpm/nvtx3.cmake
```diff
@@ -0,0 +1,107 @@
...
+                  EXCLUDE_FROM_ALL ${exclude})
+
+  include("${rapids-cmake-dir}/cpm/detail/display_patch_status.cmake")
+  rapids_cpm_display_patch_status(NVTX3)
+
+  # Set up install rules
+  if(to_install)
+    include(GNUInstallDirs)
+    install(DIRECTORY "${NVTX3_SOURCE_DIR}/c/include/" DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}")
+    install(TARGETS nvtx3-c nvtx3-cpp EXPORT ${_RAPIDS_BUILD_EXPORT_SET})
```
KyleFromNVIDIA, line 95: What happens here if the user doesn't pass `_RAPIDS_BUILD_EXPORT_SET`?

KyleFromNVIDIA (commented):
A few questions, looking good so far.

rapids-cmake/cpm/nvtx3.cmake
bdice: I guess so. I copied this: https://github.com/search?q=repo%3Arapidsai%2Frapids-cmake%20%22Fix%20up%20_RAPIDS_UNPARSED_ARGUMENTS%22&type=code

robertmaynard: ` _RAPIDS_UNPARSED_ARGUMENTS` is a side-effect of `cmake_parse_arguments` which holds all unparsed arguments from `ARGN`

rapids-cmake/cpm/nvtx3.cmake
bdice: I don't really know the expected behavior here. I decided to try removing the "build export set" since we're not conditioning on that above. We were deciding whether to execute this block based on `if(INSTALL_EXPORT_SET IN_LIST ARGN)`. The build passed locally for me when I removed the "build export set" entirely. Take a look at 6ba6a376bb2184aec693097934dc8f71c37b8665 and see if it seems reasonable.

rapids-cmake/cpm/nvtx3.cmake
bdice: See other thread: https://github.com/rapidsai/rapids-cmake/pull/606/files#r1598804744

robertmaynard: this whole block needs to be moved to be after `rapids_cpm_package_details` and updated to also check the state of `exclude`. See cuco.cmake for a reference

rapids-cmake/cpm/nvtx3.cmake
```diff
@@ -0,0 +1,107 @@
...
+                  GIT_REPOSITORY ${repository}
+                  GIT_TAG ${tag}
+                  GIT_SHALLOW ${shallow} ${patch_command} SOURCE_SUBDIR c
+                  EXCLUDE_FROM_ALL ${exclude})
+
+  include("${rapids-cmake-dir}/cpm/detail/display_patch_status.cmake")
+  rapids_cpm_display_patch_status(NVTX3)
+
+  # Set up install rules
+  if(to_install)
```
robertmaynard: Guarding the checks around `_RAPIDS_BUILD_EXPORT_SET` inside `to_install` is incorrect. You need to be able to specify a BUILD_EXPORT without an `INSTALL_EXPORT_SET`

rapids-cmake/cpm/nvtx3.cmake
```diff
@@ -0,0 +1,107 @@
...
+    include(GNUInstallDirs)
+    install(DIRECTORY "${NVTX3_SOURCE_DIR}/c/include/" DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}")
+    install(TARGETS nvtx3-c nvtx3-cpp EXPORT ${_RAPIDS_BUILD_EXPORT_SET})
+    if(NOT _RAPIDS_BUILD_EXPORT_SET STREQUAL _RAPIDS_INSTALL_EXPORT_SET)
```
robertmaynard: If a user has requested a `BUILD_EXPORT_SET` we need to generate the `export()` rules and not the install rules.

robertmaynard (changes requested):

rapids-cmake/cpm/nvtx3.cmake
robertmaynard: The `BUILD_EXPORT_SET` should control the build directory exporting. If no INSTALL or BUILD export is requested that means `nvtx3` is a private dependency and doesn't need any exporting/installing.

bdice: @robertmaynard Is any action required here, or is that comment just informational to answer Kyle's question?

rapids-cmake/cpm/nvtx3.cmake
bdice: @robertmaynard I might need more hand-holding here. I don't know what this is asking for.

rapids-cmake/cpm/nvtx3.cmake
robertmaynard: Kyle is point out that the condition of the branch is if `_RAPIDS_INSTALL_EXPORT_SET` has been specified, but you are calling the install rule with `_RAPIDS_BUILD_EXPORT_SET`. You can't presume that `_RAPIDS_BUILD_EXPORT_SET` is set in these cases, nor that it has the same value as `_RAPIDS_INSTALL_EXPORT_SET`

bdice: I changed this in 6ba6a37 to remove usage of `_RAPIDS_BUILD_EXPORT_SET`. It appears to build without that. I don't know if it is needed.

rapids-cmake/cpm/nvtx3.cmake
KyleFromNVIDIA: Basically this:

```cmake
  # Set up install rules
  if(_RAPIDS_BUILD_EXPORT_SET)
    install(TARGETS nvtx3-c nvtx3-cpp EXPORT ${_RAPIDS_BUILD_EXPORT_SET})
  endif()
  if(_RAPIDS_INSTALL_EXPORT_SET)
    include(GNUInstallDirs)
    install(DIRECTORY "${NVTX3_SOURCE_DIR}/c/include/" DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}")
    install(TARGETS nvtx3-c nvtx3-cpp EXPORT ${_RAPIDS_INSTALL_EXPORT_SET})
  endif()
```

robertmaynard: Need to also disable install rules when `EXCLUDE_FROM_ALL` flag has been set. You would look to have something like:

```cmake
  # Set up install rules
  install(TARGETS nvtx3-c nvtx3-cpp EXPORT nvtx3-targets)
  if(_RAPIDS_BUILD_EXPORT_SET)
    rapids_export(BUILD nvtx3
      EXPORT_SET nvtx3-targets
      GLOBAL_TARGETS  nvtx3-c nvtx3-cpp
      NAMESPACE nvtx::
      )
  endif()
  if(_RAPIDS_INSTALL_EXPORT_SET AND NOT exclude)
    include(GNUInstallDirs)
    install(DIRECTORY "${NVTX3_SOURCE_DIR}/c/include/" DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}")
    rapids_export(INSTALL nvtx3
      EXPORT_SET nvtx3-targets
      GLOBAL_TARGETS  nvtx3-c nvtx3-cpp
      NAMESPACE nvtx::
      )
  endif()
```

KyleFromNVIDIA: Ah - so that's what the `exclude` was for. Thanks, we weren't sure.

robertmaynard (approved):

harrism:
Can we get this merged?

PR #607, opened 2024-05-14
Update CCCL to v2.5.0

trxcllnt:
## Description
Updates CCCL to v2.5.0 and a more recent cuCollections commit.

Using cuco at [6923b3](https://github.com/NVIDIA/cuCollections/commit/6923b3b1e526b3305de81f784fdca26736caf5ba) because it was before https://github.com/NVIDIA/cuCollections/pull/479, which is a breaking change for libcudf.

CCCL PR:
* https://github.com/NVIDIA/cccl/pull/1667

RAPIDS PRs:
* https://github.com/rapidsai/cudf/pull/15552
* https://github.com/rapidsai/cuml/pull/5886
* https://github.com/rapidsai/cugraph/pull/4404
* https://github.com/rapidsai/cuspatial/pull/1382

## Checklist


robertmaynard (approved):

rapids-cmake/cpm/versions.json
```diff
@@ -6,9 +6,10 @@
       "git_tag": "v${version}"
     },
     "CCCL": {
-      "version": "2.2.0",
-      "git_url": "https://github.com/NVIDIA/cccl.git",
-      "git_tag": "v${version}",
+      "version": "2.5.0",
+      "git_shallow": false,
+      "git_url": "https://github.com/trxcllnt/cccl.git",
+      "git_tag": "fea/build-rapids",
```
bdice, line 12: We need to switch back to the upstream CCCL repo and update the tag before merging. It seems like we could adopt the latest commit on `main`, since [CCCL PR 1667](https://github.com/NVIDIA/cccl/pull/1667) [is blocked by](https://github.com/NVIDIA/cccl/pull/1667#issuecomment-2150787054) [CCCL PR 1779](https://github.com/NVIDIA/cccl/pull/1779). Then this should be mergeable. We'll update to the next tag once it's available.
```suggestion
      "git_url": "https://github.com/NVIDIA/cccl.git",
      "git_tag": "fde1cf79bde6744b6739636f502b5edfefe302e1",
```

bdice (approved):

rapids-cmake/cpm/patches/cccl/revert_pr_211.diff
trxcllnt, line 1: @robertmaynard should we keep the old `revert_pr_211.diff` in the repo, and rename this one to something like `revert_pr_211_cccl_2.5.diff`, similar to what we did with the cuDF patches? Or is it fine to only include patches for the version of CCCL in `versions.json`?

bdice: I think we only need patches for the version specified in `versions.json`. I'm not familiar with why that was done for cuDF...

trxcllnt: > I'm not familiar with why that was done for cuDF
cuDF has two sets of patches because it currently builds with CCCL 2.2, and after this change will start building with 2.5. One set will fail to apply and the other set will succeed.

trxcllnt:
I think the tests for `GENERATE_PINNED_VERSIONS` is not expecting the `git_tag` to be a commit hash.

rapids-cmake/cpm/patches/cccl/revert_pr_211.diff
trxcllnt: I removed all the patches for CCCL <2.5 in https://github.com/rapidsai/rapids-cmake/pull/607/commits/846c29c5bf23da5fea5bb2e43b4f5f36ae2f0b35

bdice (approved):
The latest changes look fine to me. I'd like to hear from @robertmaynard in case the testing changes with generating pinned versions are indicative of a problem.

PR #614, opened 2024-05-22
Removes legacy rapids-cmake cython implementations as it is deprecated in 24.08

robertmaynard:
## Description
rapids-cmake cython module has been deprecated since 24.02 and now can be removed in 24.08

This removes the cython module and instead only offers cython-core.

## Checklist


robertmaynard:
@vyasr It would be good to get your opinion on this PR.

I was worried that dropping the `rapids-cython.cmake` would break people that included it but never used functions, so now it brings in the `cython-core` implementations.

vyasr:
I don't think that's a major concern. In theory I suppose it could happen from copy-pasting top-level CMakeLists.txt files, which is I imagine how many projects that aren't RAPIDS projects I originally set up ended up with this code, but I'm not too worried about it.

OTOH I could see an argument for doing what you've done so that in the long run we could go back to `rapids-cython` rather than `rapids-cython-core` as the canonical module name. The `-core` suffix was primarily useful to differentiate, but once there's only one we could switch back.

vyasr (approved):

PR #617, opened 2024-05-23
Always offer the install target names for nvtx3

robertmaynard:
## Description
This is required so that consumers of `nvtx3` can safely use `nvtx3::nvtx3-cpp` consistently no matter if they are using nvtx3 from source or an installed version.

## Checklist

### DIFF
--- a/rapids-cmake/cpm/nvtx3.cmake
+++ b/rapids-cmake/cpm/nvtx3.cmake
@@ -114,6 +114,13 @@ function(rapids_cpm_nvtx3)
     endif()
   endif()
 
+  if(NOT TARGET nvtx::nvtx3-c AND TARGET nvtx3-c)
+    add_library(nvtx::nvtx3-c ALIAS nvtx3-c)
+  endif()
+  if(NOT TARGET nvtx::nvtx3-cpp AND TARGET nvtx3-cpp)
+    add_library(nvtx::nvtx3-cpp ALIAS nvtx3-cpp)
+  endif()
+
   # Propagate up variables that CPMFindPackage provide
   set(nvtx3_SOURCE_DIR "${nvtx3_SOURCE_DIR}" PARENT_SCOPE)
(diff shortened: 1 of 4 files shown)


rapids-cmake/cpm/nvtx3.cmake
```diff
@@ -114,6 +114,13 @@ function(rapids_cpm_nvtx3)
     endif()
   endif()
 
+  if(NOT TARGET nvtx::nvtx3-c AND TARGET nvtx3-c)
```
bdice, line 117: The docs are using nvtx3:: but the logic says nvtx::. Which do we want?

KyleFromNVIDIA: The tests are also using `nvtx3::`.

trxcllnt:
@robertmaynard do we also want to add this to the `FINAL_CODE_BLOCK` of the `rapids_export()` calls? I recall we also needed to do this in [`get_arrow.cmake`](https://github.com/rapidsai/cudf/blob/branch-24.06/cpp/cmake/thirdparty/get_arrow.cmake#L275-L286).

rapids-cmake/cpm/nvtx3.cmake
robertmaynard: yeah we should use `nvtx3::`

robertmaynard:
> @robertmaynard do we also want to add this to the `FINAL_CODE_BLOCK` of the `rapids_export()` calls? I recall we also needed to do this in [`get_arrow.cmake`](https://github.com/rapidsai/cudf/blob/branch-24.06/cpp/cmake/thirdparty/get_arrow.cmake#L275-L286).

We won't need to do that with nvtx3 since rapids-cmake is generating the build and install `nvtx3-config.cmake` and `nvtx3-targets.cmake`.

vyasr (approved):

PR #621, opened 2024-05-23
Only output CUDA architectures we are building for once

robertmaynard:
## Description
When multiple projects inside the root level all use rapids-cmake we would previously output the CUDA architectures which would all have the same values.

## Checklist

### DIFF
--- a/rapids-cmake/cuda/set_architectures.cmake
+++ b/rapids-cmake/cuda/set_architectures.cmake
@@ -79,9 +79,11 @@ function(rapids_cuda_set_architectures mode)
     list(TRANSFORM CMAKE_CUDA_ARCHITECTURES APPEND "-real")
   endif()
 
-  string(REPLACE ";" "\n  " _cuda_architectures_pretty "${CMAKE_CUDA_ARCHITECTURES}")
-  message(STATUS "${CMAKE_PROJECT_NAME} CUDA architectures building for:\n  ${_cuda_architectures_pretty}"
-  )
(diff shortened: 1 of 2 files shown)


rapids-cmake/cuda/set_architectures.cmake
```diff
@@ -79,9 +79,11 @@ function(rapids_cuda_set_architectures mode)
     list(TRANSFORM CMAKE_CUDA_ARCHITECTURES APPEND "-real")
   endif()
 
-  string(REPLACE ";" "\n  " _cuda_architectures_pretty "${CMAKE_CUDA_ARCHITECTURES}")
-  message(STATUS "${CMAKE_PROJECT_NAME} CUDA architectures building for:\n  ${_cuda_architectures_pretty}"
-  )
+  if(PROJECT_IS_TOP_LEVEL)
```
KyleFromNVIDIA, line 82: What happens if the top-level project doesn't use this function, but has sub-projects that use it? Or is that not something we expect to happen?

robertmaynard: You are absolutely correct those are use-cases wouldn't work, and we do need to support them.

I think what we needs to happen is that we setup a global property. On the first pass we define the global property and store the contents of `CMAKE_CUDA_ARCHITECTURES`. On all subsequent executions of the function if the computed `CMAKE_CUDA_ARCHITECTURES` changes we ouptut the new values. That way we also capture edge cases where a sub-project changes the architectures

KyleFromNVIDIA (approved):

PR #635, opened 2024-06-26
rapids_cmake_support_conda_env adds `-O0` to debug compile lines

### LINKED ISSUE #634: [FEA] rapids_cmake_support_conda_env should add `-O0` to debug compile lines
When building in a conda env the default `CXXFLAGS`, `CFLAGS`, etc will inject `-O2` on the compile line making Debug bugs significantly less useful.

Instead of having each RAPIDS project manually handle this issue, rapids-cmake shoudl be extended to do it for users

`rapids_cmake_support_conda_env` should be updated to add:
```cmake
target_compile_options(${target} INTERFACE "$<$<CONFIG:Debug>:-O0>")
```
That will ensure that debug builds are built without optimizations.

**Additional context**

When you specify CMAKE_BUILD_TYPE=Debug CMake will provide the -g flag to produce a build with debug information. The optimization level in these cases -O2 is coming from your env ( aka conda ) which globally injects the -O2 along with a host of other optimization flags into CXXFLAGS, CFLAGS. and so forth. This combines together as you have seen.

CMake presumption is that builds occurring under env variables such as CXXFLAGS are explicit requests. Since -O0 is the default optimization level passing -g is sufficient for debug builds and what you wanted was a 'releasish' build with debug info ( aka what CMAKE_BUILD_TYPE=RelWithDebInfo would have done ).

***What is the right way to do this?***

 One argument is that the root issue is the inability to tell conda that you want to do 'debug' builds ( easily ). Therefore the activation scripts should flip the env variables, since conda does establish CXXFLAGS_DEBUG, etc. This solves the  problem since it extends outside of RAPIDS. I expect this isn't a viable solution given how long it would take ( if possible at all )

Another argument is that CMake should pick CXXFLAGS etc based on the CMAKE_BUILD_TYPE. Which like the above isn't really possible due to how CMake's compiler detection works. Things like CXXFLAGS are only parsed on the first invocation and are treated as invariant afterwards, while CMAKE_BUILD_TYPE can change for each execution.

Another argument would be that CMake debug flags -g should change to -g -O0 which is more reasonable and would line up with the ongoing discussion around optimization levels in general ( https://gitlab.kitware.com/cmake/cmake/-/issues/23374 ). This again is a long term solution and wouldn't unblock us this year, but I will remember to push on this design going forward.

robertmaynard:
## Description
When building in a conda env the default CXXFLAGS, CFLAGS, etc will inject -O2 on the compile line making Debug bugs significantly less useful. `rapids_cmake_support_conda_env` now injects a `-O0` to override the `-O2` set by conda to make debug builds usable.

Fixes #634

## Checklist

### DIFF
--- a/rapids-cmake/cmake/support_conda_env.cmake
+++ b/rapids-cmake/cmake/support_conda_env.cmake
@@ -21,20 +21,25 @@ rapids_cmake_support_conda_env
 
 .. versionadded:: v21.06.00
 
-Establish a target that holds the CONDA include and link directories.
+Establish a target that holds the necessary compile and link information
+to properly support building in CONDA envs.
 
   .. code-block:: cmake
 
     rapids_cmake_support_conda_env( <target_name> [MODIFY_PREFIX_PATH] )
 
 Creates a global interface target called `target_name` that holds
-the CONDA include and link directories, when executed.
+the CONDA compile options, include directories, and link directories when executed.
 
 .. versionadded:: v24.06.00
 
 The include directories that `target_name` holds will be `-isystem` to match
 the behavior of conda when it builds projects.
 
+.. versionadded:: v24.08.00
+
+The `target_name` target will add the required compile flags to ensure debug builds
+are generated with `-O0` instead of the conda env default of `-O2`.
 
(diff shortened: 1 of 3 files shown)


harrism (approved):
Thank you!

vyasr (approved):

PR #636, opened 2024-06-27
rapids_test_install_relocatable now handles SOVERSION target properties

robertmaynard:
## Description
Previously the `rapids_test_install_relocatable` would presume the `file(INSTALL` command would always be a single line.
But cmake generates a multi line entry when doing the SOVERSION symlinks, and rapids-cmake would crash when encountering them.

Now the  `determine_install_location_of_all_targets` internal function does proper parsing to extract the `file(INSTALL` commands so we can support multi-line entries.

## Checklist

### DIFF
--- a/rapids-cmake/test/detail/generate_installed_CTestTestfile.cmake
+++ b/rapids-cmake/test/detail/generate_installed_CTestTestfile.cmake
@@ -195,8 +195,8 @@ function(extract_install_info)
   set(multi_value FILES)
   cmake_parse_arguments(_RAPIDS_TEST "${options}" "${one_value}" "${multi_value}"
                         ${install_contents})
-  if(_RAPIDS_TEST_TYPE STREQUAL "EXECUTABLE " OR _RAPIDS_TEST_TYPE STREQUAL "SHARED_LIBRARY"
-     OR _RAPIDS_TEST_TYPE STREQUAL "STATIC_LIBRARY " OR _RAPIDS_TEST_TYPE STREQUAL "OBJECT_LIBRARY")
+  if(_RAPIDS_TEST_TYPE STREQUAL "EXECUTABLE" OR _RAPIDS_TEST_TYPE STREQUAL "SHARED_LIBRARY"
+     OR _RAPIDS_TEST_TYPE STREQUAL "STATIC_LIBRARY" OR _RAPIDS_TEST_TYPE STREQUAL "OBJECT_LIBRARY")
     foreach(build_loc IN LISTS _RAPIDS_TEST_FILES)
       cmake_path(GET build_loc FILENAME name)
       set_property(GLOBAL PROPERTY ${name}_install ${_RAPIDS_TEST_DESTINATION})
--- a/rapids-cmake/test/detail/generate_installed_CTestTestfile.cmake
+++ b/rapids-cmake/test/detail/generate_installed_CTestTestfile.cmake
@@ -184,13 +184,17 @@ endfunction()
 # =============================================================================
 # ============== Parse Install Location Functions         ====================
 # =============================================================================
-function(extract_install_info)
+function(extract_install_info line)
+  # remove the trailing `)` so that it doesn't get parsed as part of the file name
+  string(REGEX REPLACE "\\)$" "" line "${line}")
+  string(APPEND command_contents "${line}")
+
   # Leverate separate_arguments to parse a space-separated string into a list of items We use
   # `UNIX_COMMAND` as that means args are separated by unquoted whitespace ( single, and double
   # supported).
-  separate_arguments(install_contents UNIX_COMMAND ${ARGN})
+  separate_arguments(install_contents UNIX_COMMAND ${line})
 
-  set(options "file(INSTALL" ")")
+  set(options "file(INSTALL")
   set(one_value DESTINATION TYPE)
   set(multi_value FILES)
   cmake_parse_arguments(_RAPIDS_TEST "${options}" "${one_value}" "${multi_value}"
@@ -224,7 +228,7 @@ function(determine_install_location_of_all_targets)
         # Continue to add the lines of `file(INSTALL` till we hit the closing `)` That allows us to
         # support multiple line file commands
         string(APPEND command_contents "${line}")
-        if(line MATCHES "\\)")
+        if(line MATCHES "\\)$")
           # We have all the lines for this file command, now parse it
           extract_install_info(${command_contents})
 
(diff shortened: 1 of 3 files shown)


rapids-cmake/test/detail/generate_installed_CTestTestfile.cmake
```diff
@@ -184,24 +184,24 @@ endfunction()
...
+  # supported).
+  separate_arguments(install_contents UNIX_COMMAND ${ARGN})
+
+  set(options "file(INSTALL" ")")
+  set(one_value DESTINATION TYPE)
+  set(multi_value FILES)
+  cmake_parse_arguments(_RAPIDS_TEST "${options}" "${one_value}" "${multi_value}"
+                        ${install_contents})
+  if(_RAPIDS_TEST_TYPE STREQUAL "EXECUTABLE " OR _RAPIDS_TEST_TYPE STREQUAL "SHARED_LIBRARY"
+     OR _RAPIDS_TEST_TYPE STREQUAL "STATIC_LIBRARY " OR _RAPIDS_TEST_TYPE STREQUAL "OBJECT_LIBRARY")
```
KyleFromNVIDIA, line 199:
```suggestion
  if(_RAPIDS_TEST_TYPE STREQUAL "EXECUTABLE" OR _RAPIDS_TEST_TYPE STREQUAL "SHARED_LIBRARY"
     OR _RAPIDS_TEST_TYPE STREQUAL "STATIC_LIBRARY" OR _RAPIDS_TEST_TYPE STREQUAL "OBJECT_LIBRARY")
```

KyleFromNVIDIA (changes requested):
Regarding the overall design, would it be possible to use the File API for this use case? Parsing `cmake_install.cmake` in this manner is very brittle, and the File API now [allows](https://cmake.org/cmake/help/latest/command/cmake_file_api.html) the project itself to request API output (though I realize it might not be feasible to use this until we require CMake 3.27.)

As for the implementation, I see one small problem, but otherwise looks good.

robertmaynard:
> Regarding the overall design, would it be possible to use the File API for this use case? Parsing `cmake_install.cmake` in this manner is very brittle, and the File API now [allows](https://cmake.org/cmake/help/latest/command/cmake_file_api.html) the project itself to request API output (though I realize it might not be feasible to use this until we require CMake 3.27.)
> 

Looking at the file api I am seeing some holes that would be required to switch over to it ( a welcome change ).

baseline
```cmake 
cmake_file_api(
  QUERY
  API_VERSION 1
  CODEMODEL 2.3
  TOOLCHAINS 1
)
add_library(preload SHARED usage.cu)
set_target_properties(preload PROPERTIES
  VERSION 1.1.1
  SOVERSION 1
  LIBRARY_OUTPUT_DIRECTORY my_libs
  )
install(TARGETS preload DESTINATION lib/nested)
```

First the `install` component has a minor bug where double lists the relative part:
```json
"install" : {
  "destinations" : [
    {
      "backtrace" : 2,
      "path" : "lib/nested"
    },
    {
      "backtrace" : 2,
      "path" : "lib/nested"
    }
  ],
  "prefix" : {
    "path" : "/usr/local"
  }
 }
```

A serious part is that I can't find the build directory location. I believe the `nameOnDisk` and `paths` components would be how I compute that, but it isn't correct:
```json
"nameOnDisk" : "libpreload.so",
"paths" : 
{
  "build" : ".",
  "source" : "."
},
```

Lastly is that target properties like `SOVERSION` which cause side-effects in the binary and install directory aren't representated in the file API. We need to know about these to properly fix up things like `LD_PRELOAD` env settings that exist on tests

KyleFromNVIDIA:
> First the install component has a minor bug where double lists the relative part:

Is this something we could fix upstream?

> A serious part is that I can't find the build directory location. I believe the `nameOnDisk` and `paths` components would be how I compute that, but it isn't correct:

What part specifically is not correct? By "build directory location" do you mean the location of `CMakeCache.txt`? I believe the File API provides this elsewhere.

> Lastly is that target properties like `SOVERSION` which cause side-effects in the binary and install directory aren't representated in the file API. We need to know about these to properly fix up things like `LD_PRELOAD` env settings that exist on tests

Again, could we contribute this upstream?

robertmaynard:
> What part specifically is not correct? By "build directory location" do you mean the location of `CMakeCache.txt`? I believe the File API provides this elsewhere.

I would need the relative path of the shared library on disk from the root of `CMAKE_BINARY_DIR`. Currently It looks like `"build" : ".",` is the field but it shoudl be `"build" : "./my_libs/".

> Again, could we contribute this upstream?

Absolutley we can add fixing these things upstream to our timeline :)

KyleFromNVIDIA:
> I would need the relative path of the shared library on disk from the root of `CMAKE_BINARY_DIR`. Currently It looks like `"build" : ".",` is the field but it shoudl be `"build" : "./my_libs/".

`"build":` is merely telling you the `add_subdirectory()` that the target is part of. It doesn't included the output directory. However, `"artifacts":` does include the output directory.

robertmaynard:
Ahh perfect

KyleFromNVIDIA:
> > Lastly is that target properties like `SOVERSION` which cause side-effects in the binary and install directory aren't representated in the file API. We need to know about these to properly fix up things like `LD_PRELOAD` env settings that exist on tests
>
> Again, could we contribute this upstream?

I've opened https://gitlab.kitware.com/cmake/cmake/-/issues/26087 to discuss this.

rapids-cmake/test/detail/generate_installed_CTestTestfile.cmake
```diff
@@ -224,7 +228,7 @@ function(determine_install_location_of_all_targets)
         # Continue to add the lines of `file(INSTALL` till we hit the closing `)` That allows us to
         # support multiple line file commands
         string(APPEND command_contents "${line}")
-        if(line MATCHES "\\)")
+        if(line MATCHES "\\)$")
           # We have all the lines for this file command, now parse it
           extract_install_info(${command_contents})
```
KyleFromNVIDIA, line 233: This needs to be quoted now, otherwise it will be broken up by semicolons and `line` will only get the first part.

KyleFromNVIDIA (changes requested):

rapids-cmake/test/detail/generate_installed_CTestTestfile.cmake
robertmaynard: This works as expected:
```cmake
set(line "a;b;c;)")
message(STATUS "line: ${line}")
if(line MATCHES "\\)$")
  message(STATUS "line ends with )")
endif()
```
Generates output:
```
-- line: a;b;c;)
-- line ends with )
```

Your concern about semi-colons is valid, as the `extract_install_info` does need to be changed to use `ARGN` to handle the fact that any semicolons would make the function be passed 2+ arguments, and not the 1 we are presuming

KyleFromNVIDIA (approved):

PR #639, opened 2024-07-02
Try any failing tests up to 3 times to guard against network issues

robertmaynard:
## Description
rapids-cmake has seen network issues when downloading files from GitHub. We will try tests up to 3 times when they fail to validate the issue isn't a poor network. 

## Checklist

### DIFF
--- a/ci/test_cpp.sh
+++ b/ci/test_cpp.sh
@@ -35,7 +35,7 @@ EXITCODE=0
 trap "EXITCODE=1" ERR
 set +e
 
-ctest -j20 --schedule-random --repeat pass:3 --output-on-failure --no-tests=error
+ctest -j20 --schedule-random --repeat until-pass:3 --output-on-failure --no-tests=error
 
 rapids-logger "Test script exiting with value: $EXITCODE"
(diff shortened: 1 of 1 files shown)


ci/test_cpp.sh
```diff
@@ -35,7 +35,7 @@ EXITCODE=0
 trap "EXITCODE=1" ERR
 set +e
 
-ctest -j20 --schedule-random --output-on-failure --no-tests=error
+ctest -j20 --schedule-random --repeat pass:3 --output-on-failure --no-tests=error
```
bdice, line 38: Attempting to fix: https://cmake.org/cmake/help/latest/manual/ctest.1.html#cmdoption-ctest-repeat
```suggestion
ctest -j20 --schedule-random --repeat until-pass:3 --output-on-failure --no-tests=error
```

bdice:
I really want this fix since my PR #637 has been failing repeatedly due to network issues. I pushed a commit to fix a flag -- hope that's alright @robertmaynard.

bdice (approved):

PR #640, opened 2024-07-07
Remove CCCL patch for PR 211.

bdice:
## Description
While upgrading CCCL, we ran into a test failure in cuSpatial. We added a patch to revert some changes from CCCL but the root cause was a bug in cuSpatial. I have fixed that bug here: https://github.com/rapidsai/cuspatial/pull/1402

Once that PR is merged, we can remove this CCCL patch.

See also:
- https://github.com/rapidsai/rapids-cmake/pull/511
- https://github.com/NVIDIA/cccl/issues/1939

## Checklist


robertmaynard (approved):

rapids-cmake/cpm/versions.json
```diff
@@ -9,14 +9,7 @@
       "version": "2.5.0",
       "git_shallow": false,
       "git_url": "https://github.com/NVIDIA/cccl.git",
```
robertmaynard, line 11: Since you dropped the patch for cccl, you need to update the `cpm_find-patch-command` test to validate against a different project we patch

bdice: I looked at `versions.json` but I don't see any projects with long-term patches. We have patches for Thrust/libcudacxx but those will be going away. We also have patches for fmt and spdlog but those may be updated soon.

I can change this to libcudacxx but it's just kicking the can down the road.

bdice: @robertmaynard Please take a look now. I updated this in 711f4b0.

robertmaynard (approved):

PR #641, opened 2024-07-08
Set CUDA_RUNTIME_LIBRARY to documented case style

KyleFromNVIDIA:
CMake documentation lists "Shared" and "Static", not "SHARED" and "STATIC". Set the value with the documented casing.

## Checklist

### DIFF
--- a/rapids-cmake/cuda/set_runtime.cmake
+++ b/rapids-cmake/cuda/set_runtime.cmake
@@ -1,5 +1,5 @@
 #=============================================================================
-# Copyright (c) 2023, NVIDIA CORPORATION.
+# Copyright (c) 2023-2024, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


robertmaynard (commented):
This isn't a breaking change. CMake internally uppercases the value before evaluation so it can support `shared`, `SHARED` etc.

https://gitlab.kitware.com/cmake/cmake/-/blob/master/Source/cmGeneratorTarget.cxx?ref_type=heads#L3557

KyleFromNVIDIA:
Indeed - I had not noticed that when reading the CMake code. Thanks.

KyleFromNVIDIA:
Given that this PR doesn't really fix anything, I'm fine with just closing it and leaving things as they are.

robertmaynard:
> Given that this PR doesn't really fix anything, I'm fine with just closing it and leaving things as they are.

I think matching the documented public API for this function is a good call. We should towards the public API and not the internal implementation.

I would just like to revise the commit message and title to better reflect that. e.g. "Set CUDA_RUNTIME_LIBRARY to documented case style"

KyleFromNVIDIA:
OK, I've updated the PR title and description.

robertmaynard (approved):

PR #645, opened 2024-07-17
generate_resource_spec uses the enabled languages to determine compiler

robertmaynard:
## Description
Instead of presuming that `CMAKE_CUDA_COMPILER` is sufficient for using the CUDA language, we instead check
if CUDA is part of the enabled set of langauges

## Checklist


rapids-cmake/test/generate_resource_spec.cmake
```diff
@@ -60,7 +60,8 @@ function(rapids_test_generate_resource_spec DESTINATION filepath)
 
     set(compiler "${CMAKE_CXX_COMPILER}")
     set(lang CXX)
-    if(DEFINED CMAKE_CUDA_COMPILER)
+    get_property(rapids_languages GLOBAL PROPERTY ENABLED_LANGUAGES)
```
Jacobfaib, line 63: Github isn't letting me add comments on lines that haven't been changed, but should probably amend the CUDA/C++ existence check on lines 49-52 as well with this.

rapids-cmake/test/generate_resource_spec.cmake
```diff
@@ -60,7 +60,8 @@ function(rapids_test_generate_resource_spec DESTINATION filepath)
...
+    if("CUDA" IN_LIST rapids_languages)
```
Jacobfaib, line 64: Reading over the logs of the original bug report, I'm not sure this will fully fix it. In particular, the `detail/generate_resource_spec/CMakeLists.txt` has these lines:
```cmake
if(cuda_toolkit)
  find_package(CUDAToolkit REQUIRED QUIET) # <-- failed here
  target_link_libraries(generate_ctest_json PRIVATE CUDA::cudart_static)
  # ...
```
So the problem is that `cuda_toolkit` was truthy (as the first `find_package()` here in `generate_resource_spec.cmake` found... something) even though CUDA support was disabled.

So if `CUDA` is not in the enabled languages, consider also
```cmake
set(CUDAToolkit_FOUND OFF)
```

robertmaynard: CUDA language support is orthogonal to CUDAToolkit support. You can have a C++ compiler only project that uses the CUDA Toolkit but doesn't launch CUDA kernels directly.

So we aren't going to enforce requiring the CUDA language to be enabled to  use the CUDAToolkit or to use `generate_resource_spec`.  It shouldn't be a problem is CUDAToolkit is found with C++ as we have been doing here since the start

rapids-cmake/test/generate_resource_spec.cmake
robertmaynard: Fixed!

rapids-cmake/test/generate_resource_spec.cmake
```diff
@@ -46,7 +46,19 @@ files to be generated.
...
+  if("CXX" IN_LIST rapids_languages)
+    set(rapids_lang CXX)
+    set(rapids_have_cxx ON)
+  endif()
+  if("CUDA" IN_LIST rapids_languages)
+    set(rapids_lang CUDA)
+    set(rapids_have_cuda ON)
+  endif()
+
+  if(NOT (rapids_have_cxx OR rapids_have_cuda))
```
KyleFromNVIDIA, line 61: Seems to me like you could get rid of `rapids_have_cxx` and `rapids_have_cuda` altogether, replace them with `unset(rapids_lang)`, and check for `rapids_lang` here. WDYT?

KyleFromNVIDIA (approved):
Looks good overall, just one quick question. I don't need to review again.

rapids-cmake/test/generate_resource_spec.cmake
Jacobfaib: > So we aren't going to enforce requiring the CUDA language to be enabled to use the CUDAToolkit or to use generate_resource_spec. It shouldn't be a problem is CUDAToolkit is found with C++ as we have been doing here since the start

Yeah but that is precisely the bug. I have tracked the erroneous logic down as follows:

1. `CUDA` is explicitly disabled by the user, i.e. `get_propert(ENABLED_LANGUAGES)` will not contain "CUDA".
1. `generate_resource_spec.cmake` does `find_package(CUDAToolkit QUIET)`. This _succeeds_.
2. `generate_resource_spec.cmake` thus has `CUDAToolkit_FOUND` set to a truthy value.
3. `generate_resource_spec.cmake` passes this to `detail/generate_resource_spec/CMakeLists.txt` via `-Dcuda_toolkit=${CUDAToolkit_FOUND}`.
4. `detail/generate_resource_spec/CMakeLists.txt` then sees `cuda_toolkit=TRUE`, and does `find_package(CUDAToolkit REQUIRED). This _fails_.

In this case, CUDAToolkit was able to be found in the "main" cmake lists (probably due to some extra variables set by `conda`), but subsequently not found in the resource generation `CMakeLists.txt`.

So the issue is, we are using the wrong canary to determine whether CUDA support is both enabled _and_ requested.

robertmaynard: > So the issue is, we are using the wrong canary to determine whether CUDA support is both enabled and requested.

The issue is that we aren't passing sufficient information to the try_compile to support `find_package(CUDAToolkit)`. I know this is pendantic but we need to use the correct terms. The root project correctly found the CUDA Toolkit without the CUDA language enabled. 

When using previous versions of CMake what was the end result of the `generate_resource_spec`? Did it generate a json file with `HAVE_CUDA` being defined or not? I am thinking that the problematic part is:

```
generate_resource_spec.cmake does find_package(CUDAToolkit QUIET). This succeeds.
```
With previous CMake versions this would fail since the `CMAKE_CUDA_COMPILER` ( and ENV{CUDACXX} ) had not been used to help find the CUDAToolkit. Now we do find the CUDAToolkit since that information was provided to us, but the inner `try_compile` is failing as we don't propagate `CMAKE_CUDA_COMPILER` down.

Jacobfaib: Hmmm, so perhaps we should propagate `CMAKE_CUDA_COMPILER` down unconditionally then? I suppose it cannot hurt:

1. If `CUDAToolkit_FOUND` is falsey, nothing will inspect it
2. If an older CMake version is used, then it won't consult the variable in the search.
3. If the newest CMake is used, it will help the `find_package(CTK)` to succeed in both places.

robertmaynard: That is my thinking as well. The `CUDACXX` env variable should already be auto propagated since the `try_compile` is a sub-process of CMake itself...

rapids-cmake/test/generate_resource_spec.cmake
robertmaynard: That is great improvement. Updated

rapids-cmake/test/generate_resource_spec.cmake
```diff
@@ -46,7 +46,19 @@ files to be generated.
...
   list(APPEND CMAKE_MESSAGE_CONTEXT "rapids.test.generate_resource_spec")
 
-  if(NOT DEFINED CMAKE_CUDA_COMPILER AND NOT DEFINED CMAKE_CXX_COMPILER)
+  unset(rapids_lang)
+  get_property(rapids_languages GLOBAL PROPERTY ENABLED_LANGUAGES)
+  if("CXX" IN_LIST rapids_languages)
+    set(rapids_lang CXX)
+    # Even when the CUDA language is disabled we want to pass this since it is used by
+    # find_package(CUDAToolkit) to find the location
+    set(CMAKE_TRY_COMPILE_PLATFORM_VARIABLES CMAKE_CUDA_COMPILER)
```
KyleFromNVIDIA, line 55: I just checked the CMake source code, expecting that `CMAKE_CUDA_COMPILER` would already be propagated by `try_compile()` - it is not. That blows my mind. Surely there must be a reason?

robertmaynard: IIRC the `try_compile` leverages the `CMakeFiles/` directory of the parent process. So therefore only enabled languages would be propagated.

PR #633, opened 2024-06-24
Upgrade nvcomp to 4.0.1

vuule:
## Description
Upgrade to nvcomp 4.0. As part of that change we need to refactor the `nvcomp.cmake` to be aware of the new CMake files and targets that nvcomp generates.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/rapids-cmake/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #648, opened 2024-07-19
proprietary_binary_cuda_version_mapping allows for better CUDA version mapping

robertmaynard:
## Description
Allows us to support download urls that have unique mappings per CUDA major version

## Checklist


vyasr (approved):
Implementation seems fine. Can we merge this safely without forcing an immediate bump to 4.0, though, or did they also modify all previous published binaries (I assume not).

robertmaynard:
> Implementation seems fine. Can we merge this safely without forcing an immediate bump to 4.0, though, or did they also modify all previous published binaries (I assume not).

We can merge this safely without bumping to 4.0. 
The bump to 4.0 requires more work ( As seen in https://github.com/rapidsai/rapids-cmake/pull/633 ) so I wanted to pull this out so we could get it reviewed beforehand.

PR #651, opened 2024-07-22
Add rapids_cpm_bs_thread_pool()

KyleFromNVIDIA:
Add a function to download and export bshoshany/thread-pool.

## Checklist
- [ ] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/rapids-cmake/blob/HEAD/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.
- [ ] The `cmake-format.json` is up to date with these changes.
- [ ] I have added new files under rapids-cmake/
   - [ ] I have added include guards (`include_guard(GLOBAL)`)
   - [ ] I have added the associated docs/ rst file and update the api.rst

PR #652, opened 2024-07-22
Add rapids_cpm_nvtx3 to cmake-format-rapids-cmake.json

KyleFromNVIDIA:
## Description
Noticed while working on https://github.com/rapidsai/rapids-cmake/pull/651.

## Checklist

### DIFF
--- a/cmake-format-rapids-cmake.json
+++ b/cmake-format-rapids-cmake.json
@@ -183,6 +183,16 @@
           "USE_PROPRIETARY_BINARY": 1
         }
       },
+      "rapids_cpm_nvtx3": {
+        "pargs": {
+          "nargs": 0
+        },
+        "kwargs": {
+          "BUILD_EXPORT_SET": 1,
+          "INSTALL_EXPORT_SET": 1,
+          "USE_PROPRIETARY_BINARY": 1
+        }
+      },
       "rapids_cpm_rmm": {
         "pargs": {
           "nargs": 0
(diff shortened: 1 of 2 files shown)


cmake-format-rapids-cmake.json
```diff
@@ -183,6 +183,16 @@
...
         }
       },
+      "rapids_cpm_nvtx3": {
+        "pargs": {
+          "nargs": 0
+        },
+        "kwargs": {
+          "BUILD_EXPORT_SET": 1,
+          "INSTALL_EXPORT_SET": 1,
+          "USE_PROPRIETARY_BINARY": 1
```
vyasr, line 193: @robertmaynard this is consistent with how the APIs are written, but I don't think this is right. We don't actually respect `USE_PROPRIETARY_BINARY` for nvtx3 unless I've missed something, it's just in the docstring (possibly copy-pasted?). Should we update the CMake function in this PR as well? I think we'd just want to remove [this line](https://github.com/rapidsai/rapids-cmake/blob/branch-24.08/rapids-cmake/cpm/nvtx3.cmake#L55).

vyasr (changes requested):

cmake-format-rapids-cmake.json
KyleFromNVIDIA: I wondered about that too. If it's truly not being used then I am in favor of taking it out.

robertmaynard: Yeah drop it, it should only be documented for nvcomp.

vyasr (approved):

PR #671, opened 2024-08-13
`write_language` now properly walks up the `add_subdirectory` call stack

robertmaynard:
## Description
Instead of assuming the filesystem directory layout matches `add_subdirectory` calls, we now walk the CMake
`PARENT_DIRECTORY` property.

This allows us to handle more custom layouts of CMake projects

## Checklist

### DIFF
--- a/rapids-cmake/export/write_language.cmake
+++ b/rapids-cmake/export/write_language.cmake
@@ -1,5 +1,5 @@
 #=============================================================================
-# Copyright (c) 2021-2023, NVIDIA CORPORATION.
+# Copyright (c) 2021-2024, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


rapids-cmake/export/write_language.cmake
```diff
@@ -115,16 +115,14 @@ include("${CMAKE_BINARY_DIR}/cmake/PropagateCMake@lang@Compiler.cmake")
 # - Each directory but the root needs to include `PropagateCMake@lang@Compiler.cmake`
 # - Since the root directory doesn't have a parent it only needs to include
 #   `CMake@lang@Information`
+set(rapids_directories "${CMAKE_CURRENT_SOURCE_DIR}")
+get_directory_property(parent_dir DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}" PARENT_DIRECTORY)
```
vyasr, line 119: I'm not sure I understand what has changed. Does the `cmake_path(PARENT_PATH)` syntax not simply return the parent directory? Where is the `add_subdirectory`-related assumption being made? The change I see is from `cmake_path` to `get_directory_property`, but I don't understand why.

robertmaynard: `cmake_path(PARENT_PATH)` returns the parent filesystem directory but that isn't always the parent CMakeLists.txt directory.

Consider a file-system that looks like:
```
root/
 -> A/CMakeLists.txt
   -> A/nested/CMakeLists.txt
 -> B/CMakeLists.txt
 ```
 
 It is entirely possible that the CMake parent directory call stack for `A/nested/CMakeLists.txt` would look like:
 ```
 root
B/
A/nested/
```

This kind of issue occurs in the wild when something calls `add_subdirectory` on some project that is installed in an entirely different location ( CPM downloads outside the build directory ). Which results in cmake_language errors when we try to install hooks on a directory that hasn't actually been processed by `add_subdirectory`.

vyasr: Ah I think I was thinking about this backwards. I didn't realize what `get_directory_property` did. https://cmake.org/cmake/help/latest/prop_dir/PARENT_DIRECTORY.html was what I was missing.

vyasr (approved):

PR #673, opened 2024-08-13
Make `rapids_cpm_ccl()` disable CUB detail namespace

trxcllnt:
## Description
Define `CUB_DISABLE_NAMESPACE_MAGIC` and `CUB_IGNORE_NAMESPACE_MAGIC_ERROR` so CUB doesn't include the `__CUDA_ARCH_LIST__` in its ABI.

## Checklist

### DIFF
--- a/rapids-cmake/cpm/cccl.cmake
+++ b/rapids-cmake/cpm/cccl.cmake
@@ -146,10 +146,14 @@ function(rapids_cpm_cccl)
   if(TARGET CCCL::CCCL)
     # Can be removed once we move to CCCL 2.3
     #
+    target_compile_definitions(CCCL::CCCL INTERFACE CUB_DISABLE_NAMESPACE_MAGIC)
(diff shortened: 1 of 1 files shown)


vyasr:
@robertmaynard can you take a look here? I've lost track of what decisions we've made where about keeping or removing this particular ABI mangling. I know it's caused headaches with differences between host and device compilation of the same headers.

robertmaynard:
We know that for sure that projects the build with `-fvisibility=hidden` will have no impact since the CUB mangling is a fix for clashes in the global symbol table lookup.

My major concern is that by fully disabling the CUB detail magic logic we also drop the `CUB_VERSION` from the namespace name. That I think is important as it means that we could clash with DSO built with older ( 1.X ) versions of CUB which didn't have any name mangling. This issue currently exists with the `Thrust` types but we sorta ignore it for now.

In a perfect world we could tell CCCL to drop the CUDA_ARCH mangling but leave the `VERSION` mangling. That would give us reproducible builds and also ensure we don't clash at runtime

PR #675, opened 2024-08-16
Allow nested lib location for nvcomp

KyleFromNVIDIA:
## Description
Modify `rapids_cpm_nvcomp()` to allow `CMAKE_INSTALL_LIBDIR` and `CMAKE_INSTALL_INCLUDEDIR` to be nested directories. This is needed to support building wheels that set these variables in order to put their libraries in their own package directory.

## Checklist
- [ ] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/rapids-cmake/blob/HEAD/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.
- [ ] The `cmake-format.json` is up to date with these changes.
- [ ] I have added new files under rapids-cmake/
   - [ ] I have added include guards (`include_guard(GLOBAL)`)
   - [ ] I have added the associated docs/ rst file and update the api.rst

PR #676, opened 2024-08-16
rapids_cpm_nvcomp has consistent behavior on every cmake execution

robertmaynard:
## Description
Previously nvcomp wouldn't restore `nvcomp_proprietary_binary` and the install rules for nvcomp on the second execution of CMake. Now we properly restore all needed nvcomp related state so that for each execution of cmake we have consistent behavior.

## Checklist

### DIFF
--- a/rapids-cmake/cpm/nvcomp.cmake
+++ b/rapids-cmake/cpm/nvcomp.cmake
@@ -102,6 +102,11 @@ function(rapids_cpm_nvcomp)
     endif()
   endif()
 
+  # Set up the version of nvcomp we have downloaded to match the OS layout. that means ensuring we
+  # have a `lib64` directory on Fedora based machines
+  include("${rapids-cmake-dir}/cmake/install_lib_dir.cmake")
+  rapids_cmake_install_lib_dir(lib_dir)
+
   # second see if we have a proprietary pre-built binary listed in versions.json and download it if
   # requested.
   set(nvcomp_proprietary_binary OFF) # will be set to true by rapids_cpm_get_proprietary_binary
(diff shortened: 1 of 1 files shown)


rapids-cmake/cpm/nvcomp.cmake
```diff
@@ -143,10 +143,10 @@ function(rapids_cpm_nvcomp)
...
       # Record the nvcomp_DIR so that if USE_PROPRIETARY_BINARY is disabled we can safely clear the
       # nvcomp_DIR value
-      set(nvcomp_proprietary_binary_dir "${nvcomp_ROOT}/${lib_dir}/cmake/nvcomp")
-      cmake_path(NORMAL_PATH nvcomp_proprietary_binary_dir)
-      set(rapids_cpm_nvcomp_proprietary_binary_dir "${nvcomp_proprietary_binary_dir}"
-          CACHE INTERNAL "nvcomp proprietary location")
+      set(nvcomp_proprietary_root "${nvcomp_ROOT}")
+      cmake_path(NORMAL_PATH nvcomp_proprietary_root)
+      set(rapids_cpm_nvcomp_proprietary_root "${nvcomp_proprietary_root}"
+          CACHE INTERNAL "nvcomp proprietary root dir location")
```
KyleFromNVIDIA, line 149: How will this affect the `cpm_nvcomp-proprietary_binary-lib-location-nested` test that was added in #675? It seems to me that it will no longer find the file if it's in `nested/lib64/cmake/nvcomp`, since that's not one of the places `find_package()` looks. Requesting clarification here.

robertmaynard: The `nvcomp_proprietary_binary_dir` was never used for searches so it shouldn't have any imact.
The `nvcomp_DIR` cache variable will hold the found location from the first execution and be used for all following executions.

KyleFromNVIDIA (approved):

PR #677, opened 2024-08-16
Bump cuco version to fetch several bug fixes

PointKernel:
## Description
This PR bumps the cuco version to the latest to fetch bug fixes for https://github.com/rapidsai/cudf/issues/16405 and https://github.com/rapidsai/cudf/issues/16526

## Checklist
- [ ] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/rapids-cmake/blob/HEAD/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.
- [ ] The `cmake-format.json` is up to date with these changes.
- [ ] I have added new files under rapids-cmake/
   - [ ] I have added include guards (`include_guard(GLOBAL)`)
   - [ ] I have added the associated docs/ rst file and update the api.rst

PR #685, opened 2024-08-23
Add workaround for gcc bug in cuco

miscco:
We discovered an unfortunate bug in the gcc constexpr evaluator that broke the alignment calculations in cuco.

There is a simple workaround so employ that to unbreak rapids with cccl ToT


bdice (commented):
Is there a reason to use a patch and not to update cuCollections and then update the SHA in versions.json?

bdice (approved):
Style check requires a trailing newline in the file. Otherwise LGTM.

bdice:
@miscco asked me offline about the testing procedure we need for this PR. I pointed to the downstream CI testing we typically do here: https://github.com/rapidsai/rapids-cmake/pull/677#issuecomment-2294110382

Here is the comparison between the previous and new commits of cuCollections: https://github.com/NVIDIA/cuCollections/compare/abc5095b011feccd7d3ecb36b369741b96ae6f0e..d3477661d771e0d6fd22259bf6dd6f8c64a7401c

The diff is pretty small and confined, so I would be okay merging this PR without downstream CI testing if it saves time. @miscco verified that cuDF builds with this change. But I'd want to get an okay from @PointKernel and/or @robertmaynard before merging.

PointKernel:
Thanks for bringing this up.

I reviewed the changes yesterday, and as @bdice mentioned, they are minimal. I don't see any need to retest the RAPIDS suite for such a minor update.

PointKernel (approved):

robertmaynard:
Will merge early tomorrow :+1:

PR #692, opened 2024-09-19
Add INSTALL_TARGET argument to rapids_add_test()

KyleFromNVIDIA:
## Description
If another command, such as `mpiexec`, is used to execute a target, `rapids_add_test()` currently has no way of installing the target dependency. Add an `INSTALL_TARGET` argument to support this use case.

## Checklist


KyleFromNVIDIA:
Currently testing this PR over at https://github.com/rapidsai/cugraph/pull/4656

bdice (approved):
I _think_ I understand this. But it's right at the edge of my knowledge of targets/install commands. Feel free to seek a second reviewer if you think it'd be worthwhile.

rapids-cmake/test/add.cmake
```diff
@@ -89,6 +89,15 @@ function(rapids_test_add)
...
     set(command "$<TARGET_FILE:${command}>")
   endif()
+  if(DEFINED _RAPIDS_TEST_INSTALL_TARGET)
+    if(NOT TARGET ${_RAPIDS_TEST_INSTALL_TARGET})
+      message(FATAL_ERROR "rapids_add_test given INSTALL_TARGET \"${_RAPIDS_TEST_INSTALL_TARGET}\", which does not exist"
+      )
+    endif()
+    set(target_to_install ${_RAPIDS_TEST_INSTALL_TARGET})
+  else()
+    set(target_to_install ${command_or_target})
```
vyasr, line 99: Can we add a check that `target_to_install` is a valid target here, just to be safe and provide users more useful diagnostics?
```suggestion
    set(target_to_install ${command_or_target})
    if (NOT TARGET ${target_to_install})
      message(FATAL_ERROR "rapids_add_test could not infer a target to install")
    endif()
```

vyasr (approved):
Works for me. I don't think there's a general solution for inferring the target so making it explicit is good.

rapids-cmake/test/add.cmake
KyleFromNVIDIA: This would regress the use case of running a command that does not involve a target at all. Is that something we need?

KyleFromNVIDIA: We could make it a warning instead.

vyasr: Good point. Since that is a valid use case, perhaps `VERBOSE` is the appropriate mode.

KyleFromNVIDIA: OK, I've made it a `VERBOSE` message.

PR #706, opened 2024-10-17
Allow CCCL 2.7 to be used in a rapids-cmake override

robertmaynard:
## Description
Support the latest RC of CCCL in rapids-cmake. This requires a reworking of our install logic as upstream has refactored things.

## Checklist


robertmaynard:
> Looks good - did you test this with a RAPIDS build? 

I haven't tested CCCL 2.7 with a RAPIDS build.

>Do we need to update the patches, too, or wait until later when we upgrade?

rapids-cmake doesn't have any CCCL patches so there wasn't anything to test there. CUDF patches most likely won't apply and will need to be updated.

PR #708, opened 2024-10-24
Warn when override of a default package uses the incorrect case

robertmaynard:
## Description
It is a common mistake to presume that all entries in a package override json should use lower case names, since the vast majority of default packages are all lower case. This causes silent 'failures' when the override doesn't work.

We now normalize all package data internally so that  overrides for `ABC`, `abc`, and `Abc` all map to the same package. In addition we have added information tracking the proper name of bundled packages. This allows us to warn when an override uses the non correct package name, and treat it like the proper name.

## Checklist


vyasr:
Do we need to put in some sort of check to avoid accidentally introducing name collisions if there really are two packages that only differ by capitalization? I don't think it's critical but IIUC this PR is going to make such cases collide unconditionally.

robertmaynard:
> Do we need to put in some sort of check to avoid accidentally introducing name collisions if there really are two packages that only differ by capitalization? I don't think it's critical but IIUC this PR is going to make such cases collide unconditionally.

Do you mean when the package is listed twice in the defaults.json? The PR does warn if an override has a different case. 
Multiple overrides of the same package can't clash as only the first occurence of each package is read.

vyasr:
Yes, I mean in defaults.json if there is a real CCCL and cccl package for example.

vyasr (approved):

PR #715, opened 2024-11-05
Search for the normalized package name when patching

vyasr:
## Description

#708 broke patches specified via overrides due to the missing case normalization.

## Checklist


vyasr:
I would suggest that we add tests in a follow-up since getting this merged is needed to unblock current builds.

robertmaynard (approved):

PR #724, opened 2024-12-04
Fix failures in cpm_generate_pins-nested tests

vyasr:
## Description

[Recent nightly failures of the `cpm_generate_pins-nested` test](https://github.com/rapidsai/rapids-cmake/actions/runs/12133373349/attempts/1) are unexpected because neither rapids-cmake, CPM, nor CMake have changed recently. [The error](https://github.com/rapidsai/rapids-cmake/actions/runs/12133373349/job/33828828013#step:8:5066) shows that [the list of projects to verify](https://github.com/rapidsai/rapids-cmake/blob/branch-25.02/testing/cpm/verify_generated_pins/CMakeLists.txt#L28) is being completely filtered out because none of them have a `<project>_SOURCE_DIR` variable set. This led me down a rabbit hole of finding a number of different issues in both rapids-cmake itself and in tests. I'll detail them one by one below.

### Problem 1: Our logic for when to download packages for which rapids-cmake provides specialized `rapids_cpm*` commands is wrong

The first piece of evidence here is the difference between recent builds of our CI images. [The most recent builds of miniforge-cuda have spdlog in the base environment](https://github.com/rapidsai/ci-imgs/actions/runs/12145567995/job/33867533318#step:8:274), while [previous builds do not](https://github.com/rapidsai/ci-imgs/actions/runs/12036072167/job/33556500014#step:8:307). That should not affect our tests [because we set `CPM_DOWNLOAD_ALL` when populating the CMake cache](https://github.com/rapidsai/rapids-cmake/blob/branch-25.02/testing/utils/fill_cache/CMakeLists.txt#L36), but it turns out that #348 introduced a bug where [the `always_download` variable gets defaulted to OFF](https://github.com/rapidsai/rapids-cmake/blob/branch-25.02/rapids-cmake/cpm/detail/package_details.cmake#L95) and because of [how it propagates to `CPM_DOWNLOAD_ALL` in parent scope](https://github.com/rapidsai/rapids-cmake/blob/branch-25.02/rapids-cmake/cpm/detail/package_details.cmake#L118) we wind up with `CPM_DOWNLOAD_ALL` always being false. As a result, we have actually been ignoring `CPM_DOWNLOAD_ALL` for any packages for which `rapids_cpm_package_details` is called (the more specific `CPM_DOWNLOAD_<project>` should still have been working since that takes precedence).

Solution: We unset the `always_download` variable initially instead of setting it to off so that the check is handled correctly.

### Problem 2: Even with the above logic fixed, the actual test will use the spdlog from the environment

The above fix ensures that when we populate the CPM source cache in our tests we download spdlog instead of finding it in the environment. However, when tests are subsequently run they will still prefer the package unless `CPM_DOWNLOAD_ALL` (or `CPM_DOWNLOAD_<package>`) is set because of how CPM prioritizes.

Solution: To resolve this, for all rapids-cpm tests we are now setting `CPM_DOWNLOAD_<pkg>` for all of the packages that have been added to our CPM cache. This fix is safe since it effectively just avoids finding local copies of the built package; the source will always come from the cache by design.

This change is sufficient for tests to pass, but on closer inspection we see that only spdlog is verified by the verification script and not rmm or fmt. That's because of the next problems.

### Problem 3: All of our tests of pinning are incorrectly checking for rmm with the wrong case

[The `project` call in rmm uses the uppercase name "RMM"](https://github.com/rapidsai/rmm/blob/branch-25.02/CMakeLists.txt#L25), but [the entry in rapids-cmake's versions.json](https://github.com/rapidsai/rapids-cmake/blob/branch-25.02/rapids-cmake/cpm/versions.json#L66) uses the lowercase name. As a result, our current logic for verifying generate pins can never test rmm because the project [is initially filtered based on `<project>_SOURCE_DIR` being defined](https://github.com/rapidsai/rapids-cmake/blob/branch-25.02/testing/cpm/verify_generated_pins.cmake#L33), which will use the uppercase name from the `project` call, but when actually verifying we use `rapids_cpm_package_details` to get the versions.json data _using the same string_, which will now have the wrong case if we passed the uppercase version originally.

### Problem 4: All of our tests of pinning are incorrectly checking for fmt with the wrong case

This is the same problem as the above. [fmt uses `project(FMT)`](https://github.com/fmtlib/fmt/blob/master/CMakeLists.txt#L151).

Solution for 3 and 4: We now use the `CPM_PACKAGE_<project>_SOURCE_DIR` variables instead of the `<project>_SOURCE_DIR` variables. The former are guaranteed to be using the export names of the package and will therefore always be case-consistent with the values in versions.json.

### Other changes in this PR

- [This gtest test](https://github.com/rapidsai/rapids-cmake/blob/branch-25.02/testing/cpm/cpm_find-gtest-no-gmock/CMakeLists.txt) is explicitly based on finding a mocked up version of the package locally, so with the changes in this PR we now have to turn of `CPM_DOWNLOAD_GTest` explicitly inside the test.
- I removed scikit-build from the testing CPM cache since as of #614 we no longer support its usage via the legacy rapids-cython module.

The change to `verify_generated_pins.cmake` ensures that all of the above changes are actually having the desired effect, since now rather than silently passing when only a subset of projects requesting verification are downloaded we will error. All projects must be pulled correctly from the CPM cache during the test (which implicitly checks all four problems above) in order for tests to pass.

## Checklist

### DIFF
--- a/rapids-cmake/cpm/detail/package_details.cmake
+++ b/rapids-cmake/cpm/detail/package_details.cmake
@@ -92,6 +92,9 @@ function(rapids_cpm_package_details package_name version_var url_var tag_var sha
   set(exclude_from_all OFF)
   rapids_cpm_json_get_value(exclude_from_all)
 
+  # Ensure that always_download is not set by default so that the if(DEFINED always_download) check
+  # below works as expected in the default case.
+  unset(always_download)
   if(override_json_data AND json_data AND git_details_overridden)
     # `always_download` default value requires the package to exist in both the default and override
     # and that the git url / git tag have been modified.
--- a/rapids-cmake/cpm/detail/package_details.cmake
+++ b/rapids-cmake/cpm/detail/package_details.cmake
@@ -92,7 +92,6 @@ function(rapids_cpm_package_details package_name version_var url_var tag_var sha
   set(exclude_from_all OFF)
   rapids_cpm_json_get_value(exclude_from_all)
 
-  set(always_download OFF)
   if(override_json_data AND json_data AND git_details_overridden)
     # `always_download` default value requires the package to exist in both the default and override
     # and that the git url / git tag have been modified.
--- a/testing/cpm/cpm_find-gtest-no-gmock/CMakeLists.txt
+++ b/testing/cpm/cpm_find-gtest-no-gmock/CMakeLists.txt
@@ -1,5 +1,5 @@
 #=============================================================================
-# Copyright (c) 2022, NVIDIA CORPORATION.
+# Copyright (c) 2022-2024, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -20,6 +20,9 @@ include(${rapids-cmake-dir}/cpm/init.cmake)
 include(${rapids-cmake-dir}/cpm/gtest.cmake)
 
 set(CMAKE_PREFIX_PATH "${CMAKE_CURRENT_SOURCE_DIR}/mock_installed_gtest")
+# Downloading is turned on by default for all packages in the CPM cache when
+# testing, but in this case we specifically want it off.
+set(CPM_DOWNLOAD_GTest OFF)
 
 rapids_cpm_init()
 rapids_cpm_gtest()
--- a/testing/cpm/verify_generated_pins.cmake
+++ b/testing/cpm/verify_generated_pins.cmake
@@ -30,7 +30,7 @@ function(verify_generated_pins target_name)
 
   # only check projects that were downloaded by CPM (ignore those already in the build environment)
   foreach(proj IN LISTS _RAPIDS_PROJECTS)
-    if(${proj}_SOURCE_DIR)
+    if(CPM_PACKAGE_${proj}_SOURCE_DIR)
       list(APPEND projects-to-verify ${proj})
     endif()
   endforeach()
--- a/testing/cpm/verify_generated_pins.cmake
+++ b/testing/cpm/verify_generated_pins.cmake
@@ -28,10 +28,9 @@ function(verify_generated_pins target_name)
     set(_RAPIDS_PIN_FILE "${CMAKE_CURRENT_BINARY_DIR}/rapids-cmake/pinned_versions.json")
   endif()
 
-  # only check projects that were downloaded by CPM (ignore those already in the build environment)
   foreach(proj IN LISTS _RAPIDS_PROJECTS)
-    if(CPM_PACKAGE_${proj}_SOURCE_DIR)
-      list(APPEND projects-to-verify ${proj})
+    if(NOT CPM_PACKAGE_${proj}_SOURCE_DIR)
+      message(FATAL_ERROR "Attempting to verify a project that was not cloned as part of this build")
     endif()
   endforeach()
 
@@ -39,7 +38,7 @@ function(verify_generated_pins target_name)
     COMMAND ${CMAKE_COMMAND} "-S${CMAKE_CURRENT_FUNCTION_LIST_DIR}/verify_generated_pins/" "-B${CMAKE_BINARY_DIR}/${target_name}_verify_build"
     "-Drapids-cmake-dir=${rapids-cmake-dir}"
     "-Dpinned_versions_file=${_RAPIDS_PIN_FILE}"
-    "-Dprojects-to-verify=${projects-to-verify}"
+    "-Dprojects-to-verify=${_RAPIDS_PROJECTS}"
     "-Dprojects-not-in-list=${_RAPIDS_PROJECTS_NOT_EXIST}"
     VERBATIM
   )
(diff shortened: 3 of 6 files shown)


rapids-cmake/cpm/detail/package_details.cmake
```diff
@@ -92,7 +92,6 @@ function(rapids_cpm_package_details package_name version_var url_var tag_var sha
   set(exclude_from_all OFF)
   rapids_cpm_json_get_value(exclude_from_all)
 
-  set(always_download OFF)
```
robertmaynard, line 95: We should unset the variable here instead. That way any user variable called `always_download` won't trigger this incorrect behavior as well.

robertmaynard (changes requested):
An amazing write-up on this

robertmaynard:
As for Issue 2. I think it is safe to map the `NO_CPM_CACHE` test option to also set `CPM_DOWNLOAD_<pkg>` for every package that we have in our CPM testing cache

robertmaynard:
Problem 3: All of our tests of pinning are incorrectly checking for rmm with the wrong case:

1. The `project` call will establish varibles based on the the casing in the `project()` it self. So projects like `rmm` and `fmt` for those variables will make them all upper case. ( as you pointed out for rmm )
2. rapids-cmake uses `rmm` lowercase as that is the casing in the `rapids_export` call ( https://github.com/rapidsai/rmm/blob/branch-25.02/CMakeLists.txt#L170 ) .

I think the solution is that the test checks should use `CPM_PACKAGE_${package}_SOURCE_DIR`  with the casing found in the versions.json file.  ( https://github.com/rapidsai/rapids-cmake/blob/branch-25.02/rapids-cmake/cpm/detail/pinning_write_file.cmake#L46 ) This is what we use internally to rapids-cmake to build up the list of projects we can safely check for having source code

vyasr:
> I think the solution is that the test checks should use CPM_PACKAGE_${package}_SOURCE_DIR with the casing found in the versions.json file. ( https://github.com/rapidsai/rapids-cmake/blob/branch-25.02/rapids-cmake/cpm/detail/pinning_write_file.cmake#L46 ) This is what we use internally to rapids-cmake to build up the list of projects we can safely check for having source code

Yes I was considering this as well. CPM sets the `CPM_PACKAGE_*` variables as internal so I wasn't sure if that was the wisest choice, but given our tight integration with CPM perhaps that is OK?

vyasr:
> As for Issue 2. I think it is safe to map the `NO_CPM_CACHE` test option to also set `CPM_DOWNLOAD_<pkg>` for every package that we have in our CPM testing cache

I think that this is orthogonal to `NO_CPM_CACHE`. The CPM cache only kicks in if `CPMAddPackage` is called so that we look in the source cache instead of accessing the network for downloads. What we need is to avoid this whole process being preempted by `CPMFindPackage` finding the built package somewhere on the system already. I think we may want to set `CPM_DOWNLOAD_<pkg>` for every test so that when `NO_CPM_CACHE` is on we guarantee downloading rather than using something on the system and when `NO_CPM_CACHE` is off we always use the source from the cache rather than a pre-built library somewhere on the system.

robertmaynard:
> > I think the solution is that the test checks should use CPM_PACKAGE_${package}_SOURCE_DIR with the casing found in the versions.json file. ( https://github.com/rapidsai/rapids-cmake/blob/branch-25.02/rapids-cmake/cpm/detail/pinning_write_file.cmake#L46 ) This is what we use internally to rapids-cmake to build up the list of projects we can safely check for having source code
> 
> Yes I was considering this as well. CPM sets the `CPM_PACKAGE_*` variables as internal so I wasn't sure if that was the wisest choice, but given our tight integration with CPM perhaps that is OK? I agree that if possible we want to use the name of the export and not the project name if the two are inconsistent.

Lets use it. We already rely on it inside the pinning logic

robertmaynard:
> > As for Issue 2. I think it is safe to map the `NO_CPM_CACHE` test option to also set `CPM_DOWNLOAD_<pkg>` for every package that we have in our CPM testing cache
> 
> I think that this is orthogonal to `NO_CPM_CACHE`. The CPM cache only kicks in if `CPMAddPackage` is called so that we look in the source cache instead of accessing the network for downloads. What we need is to avoid this whole process being preempted by `CPMFindPackage` finding the built package somewhere on the system already. I think we may want to set `CPM_DOWNLOAD_<pkg>` for every test so that when `NO_CPM_CACHE` is on we guarantee downloading rather than using something on the system and when `NO_CPM_CACHE` is off we always use the source from the cache rather than a pre-built library somewhere on the system.

That makes sense :+1:

vyasr:
I uncovered one more issue in https://github.com/rapidsai/rapids-cmake/issues/725, but I'm not going to try and fix that here.

rapids-cmake/cpm/detail/package_details.cmake
vyasr: Done.

robertmaynard (approved):

PR #733, opened 2024-12-16
Correct nvbench + conda test to explicitly pull spdlog

robertmaynard:
## Description
As of https://github.com/rapidsai/rmm/pull/1722, rmm no longer depends on spdlog. However, some tests were depending on spdlog transitively through rmm. This PR adds spdlog as an explicit dependency to correct those failures.

## Checklist


bdice (approved):
Looks good -- I'll add the description @vyasr requested so we can merge this.

vyasr (approved):

PR #736, opened 2024-12-23
remove setup.cfg, use pyproject.toml

jameslamb:
## Description

This repo doesn't have a Python package or use `setuptools`, but it has a `setup.cfg`... to hold configuration for tools like `codespell`.

This proposes removing that and using a `pyproject.toml` to store that configuration, matching the conventions used across most of the rest of RAPIDS.

It also updates pre-commit configurations to use the latest version of `rapids-dependency-file-generator` (v1.17.0).

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,7 +2,7 @@
 
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.4.0
+    rev: v5.0.0
     hooks:
       - id: trailing-whitespace
         exclude: |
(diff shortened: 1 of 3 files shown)


bdice (commented):
We might need some pre-commit fixes for this repo. I realized in https://github.com/rapidsai/rmm/pull/1769 that we didn't have codespell set up correctly for RMM.

I'll see if I can push a change and then approve this.

bdice:
I applied the fix and it caught an issue with the config. I think the skip files needed to be wrapped in quotes or something. Also we don't need to ignore `.github` -- we aren't doing that in rmm, at least, and there are no typos, so I think it's fine to keep it included.

jameslamb:
Ah great, thank you for fixing that! I didn't realize `codespell` didn't look for `pyproject.toml` by default.

PR #737, opened 2025-01-02
Add rapids-logger

vyasr:
## Description

This PR adds support for fetching [`rapids-logger`](https://github.com/rapidsai/rapids-logger/) using rapids-cmake to ensure a consistent version is used by all RAPIDS libraries.

Contributes to rapidsai/build-planning#104.

## Checklist


rapids-cmake/cpm/logger.cmake
```diff
@@ -0,0 +1,67 @@
...
+
+.. |PKG_NAME| replace:: logger
+.. include:: common_package_args.txt
+
+Result Functions
+^^^^^^^^^^^^^^^^
+  :cmake:command:`rapids_make_logger` is made available
+
+#]=======================================================================]
+function(rapids_cpm_logger)
```
bdice, line 43: This should be `rapids_cpm_rapids_logger`, since the name of the package is "rapids-logger". This leads to confusing patterns elsewhere, because we just get the word "logger" (like in `logger.cmake`) without a clear indication that this is referring to a RAPIDS-maintained rapids-logger package.

rapids-cmake/cpm/versions.json
```diff
@@ -23,6 +23,13 @@
       "git_url": "https://github.com/NVIDIA/cuCollections.git",
       "git_tag": "096346b739da3fb1d9c3b402190c0a3a7e554440"
     },
+    "rapids_logger": {
+      "version": "0.1.0",
+      "git_shallow": false,
+      "always_download": true,
+      "git_url": "https://github.com/rapidsai/rapids-logger",
+      "git_tag": "98cb4d232dc4612f80d462a0a7bc5490dbc22399"
```
bdice, line 31: Will this be versioned like RAPIDS (25.02) or SemVer or something else? It would be nice if this used proper tags instead of commit hashes.

vyasr: I would like for it to use SemVer, but I'm not married to that. In general I would expect it to change pretty infrequently once we're happy with its state, so I don't think matching RAPIDS CalVer would be that useful since we'd be making a lot of identical releases.

However, I definitely don't want to address rapids-logger versioning quite yet. The package is changing too frequently and we'll wind up bumping versions very often. Let's not hold up this PR on the versioning decision.

rapids-cmake/cpm/logger.cmake
vyasr: Done.

rapids-cmake/cpm/versions.json
bdice: That's fine, I also like SemVer for this. Before 25.02 goes out, I'd like this to be a tag instead of a hash.

bdice (approved):
Please fix the docs build (some additional renames are needed), then this LGTM.

vyasr:
Whoops I missed a file path. Done.

rapids-cmake/cpm/versions.json
vyasr: Yes, it's always been my intent to reach a 0.1 release or equivalent for rapids-logger for RAPIDS 25.02.

PR #743, opened 2025-01-14
Use CCCL v2.7.0 final tag.

bdice:
## Description
CCCL is currently pinned to `v2.7.0-rc2`. Now that a final tag for `v2.7.0` is available, we should use that instead.

See diff: https://github.com/NVIDIA/cccl/compare/v2.7.0-rc2...v2.7.0

## Checklist

### DIFF
--- a/rapids-cmake/cpm/versions.json
+++ b/rapids-cmake/cpm/versions.json
@@ -14,7 +14,7 @@
       "version": "2.7.0",
       "git_shallow": false,
       "git_url": "https://github.com/NVIDIA/cccl.git",
-      "git_tag": "83028fcdcfa8de51ab359e61ce07198a6003a65e"
+      "git_tag": "v${version}"
     },
     "cuco": {
       "version": "0.0.1",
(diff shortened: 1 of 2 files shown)


bdice:
From reviewing the diff, I do not anticipate any risk from merging this as-is, without needing to do test PRs for all of RAPIDS. @robertmaynard What do you think?

bdice:
I’ll merge this around the same time as the GCC 13 upgrades (#741) so that we only bust the RAPIDS sccache one time.

PointKernel:
CCCL just merged a PR that enables progress on relaxed constexpr work in cuDF: https://github.com/NVIDIA/cccl/pull/3382

Could we include that change in this PR as well? I was planning to open a PR to update the CCCL version in `rapids-cmake`, but I noticed this PR and wanted to check here first.

bdice:
@PointKernel We should move to a stable tag first. Can we defer on a future upgrade until 25.04? CCCL RAPIDS builds are currently failing and may be broken for a bit. We can’t adopt new CCCL versions with confidence until the CCCL RAPIDS CI is fixed.

PointKernel:
> Can we defer on a future upgrade until 25.04? 

Sure, no problem.

PR #745, opened 2025-01-22
Use sysroot 2.17 on CUDA < 11.8.

bdice:
## Description
Nightly CI was failing on CUDA 11.4 due to #741. This fixes it.
Old versions of `nvcc` for CUDA 11 do not support sysroot 2.28 until CUDA 11.8.

## Checklist

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -58,6 +58,16 @@ dependencies:
         matrices:
           - matrix:
               arch: x86_64
+              cuda: "11.[2456]"
+            packages:
+              - sysroot_linux-64==2.17
+          - matrix:
+              arch: aarch64
+              cuda: "11.[2456]"
+            packages:
+              - sysroot_linux-aarch64==2.17
+          - matrix:
+              arch: x86_64
             packages:
               - sysroot_linux-64==2.28
           - matrix:


vyasr (approved):

dependencies.yaml
```diff
@@ -56,6 +56,16 @@ dependencies:
...
+          - matrix:
+              arch: x86_64
+              cuda: "11.[2456]"
+            packages:
+              - sysroot_linux-64==2.17
+          - matrix:
+              arch: aarch64
+              cuda: "11.[2456]"
+            packages:
+              - sysroot_linux-aarch64==2.17
```
jakirkham, line 68: ARM is not compatible with GLIBC 2.17

```
$ strings /opt/conda/envs/cuda11/lib/libcurand.so | grep GLIBC
GLIBC_2.17
GLIBC_2.27
GLIBC_2.17
```

So think we should stick to GLIBC 2.28 here

```suggestion
          - matrix:
              arch: aarch64
              cuda: "11.[2456]"
            packages:
              - sysroot_linux-aarch64==2.28
```

jakirkham (changes requested):
Thanks Bradley! 🙏

Think CUDA 11 ARM still needs 2.28. Have added a suggestion below

As the CI failure was only on x86_64, think this would be ok

dependencies.yaml
bdice: Good call. We can equivalently just erase this and let it use the matrix entry below.
```suggestion
```

jakirkham: Agree that looks cleaner. Let's do it 👍

bdice: Oops: this is not correct. https://anaconda.org/conda-forge/nvcc_linux-aarch64/files?version=11.4

The aarch64 packages for nvcc 11.4 (up to 11.7) require sysroot 2.17. The current state of the PR (using sysroot 2.17) is correct for all nvcc versions < 11.8, and is the same on x86 and ARM.

bdice: I agree that CUDA libraries require glibc versions newer than 2.17 -- but the `nvcc` compiler metapackage for CUDA 11 is only installable with `sysroot 2.17` until we get to recent builds of `nvcc 11.8`.

bdice: @jakirkham I am going to merge this as-is so that we can unblock nightly CI.

I think there's nothing else we should do here, but if you want to lift the restrictions for old CUDA compilers, a follow-up PR would be welcome.

jakirkham: So I had looked into this recently. Here is what I observed:

* The overly strict `sysroot` constraint was flagged as a bug
  * https://github.com/conda-forge/nvcc-feedstock/issues/107
* A fix was made to relax it
  * https://github.com/conda-forge/nvcc-feedstock/pull/106
* The old versions don't have this fix, but could benefit from the same treatment

Typically that is enough to justify a repodata patch and it makes sense

Additionally noticed that `nvcc_{{ target_platform }}`'s pinnings could be handled more systematically, which is now fixed

* https://github.com/conda-forge/nvcc-feedstock/pull/115

Following this submitted a repodata patch

* https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/950

The repodata patch landed before last night's nightly run, which [passed in CUDA 11.4 (the affected job)]( https://github.com/rapidsai/rapids-cmake/actions/runs/12943660249/job/36103356990 )

Note there is [a CUDA 12.5 failure in that nightly]( https://github.com/rapidsai/rapids-cmake/actions/runs/12943660249/job/36103358376#step:8:5479 ), but that is due to actual test failures

```
The following tests FAILED:
	570 - install_relocatable-with-gtest_discover_tests-ninja_configure (Failed)
	571 - install_relocatable-with-gtest_discover_tests-ninja (Not Run) ninja test
Errors while running CTest
```

So AFAICT this issue is already fixed

jakirkham: I'm sorry as well. I meant to update this thread with those details and just hadn't had time to do so. The CUDA 12.8 rollout started very soon after 

If you are feeling stuck, please feel free to DM me. Am here to help

PR #765, opened 2025-01-31
Switch to using rapids-logger as a library

vyasr:
## Description

This PR changes `rapids_cpm_rapids_logger` to use the new shared library version of `rapids_logger` instead of the old C++ file generation approach. In addition to changing the versions.json file, there are a couple of associated changes to the function and also some small cleanup.

Contributes to https://github.com/rapidsai/build-planning/issues/104.

## Checklist


vyasr:
I'm expecting tests to fail for the moment. I'll update those once I have RAPIDS builds working with this change.

rapids-cmake/cpm/versions.json
```diff
@@ -79,8 +78,8 @@
     },
     "rmm": {
       "version": "${rapids-cmake-version}",
-      "git_url": "https://github.com/rapidsai/rmm.git",
-      "git_tag": "branch-${version}"
+      "git_url": "https://github.com/vyasr/rmm.git",
+      "git_tag": "feat/rapids_logger_library"
```
vyasr, line 82: This will be removed before merging. I made this change for the various repo tests.

vyasr: All the generate-pins test failures are caused by this right now. It's trying to check out a branch that I don't have on my fork of rmm.

rapids-cmake/cpm/versions.json
```diff
@@ -33,9 +33,8 @@
     "rapids_logger": {
       "version": "0.1.0",
       "git_shallow": false,
-      "always_download": true,
       "git_url": "https://github.com/rapidsai/rapids-logger.git",
-      "git_tag": "8968ab3337f31c845d4e3bf6c55ae89242ded22b"
+      "git_tag": "main"
```
vyasr, line 37: This suggestion should be accepted before merging. If we need any more commits in the rapids-logger repo we can bump the commit then.
```suggestion
      "git_tag": "d8a7ba8a105163eb2dde42a8e71a1cb3680100e2"
```

bdice (approved):

vyasr:
The remaining failures here are because there are tests that pull rmm and there is a catch-22 with this PR and https://github.com/rapidsai/rmm/pull/1808. I'll request an admin merge here so that I can rerun and get CI passing on the rmm PR to merge that one.

PR #775, opened 2025-02-10
Fix tests broken by rapids-logger changes

vyasr:
## Description

#765 required an admin merge on this repository to break a circular dependency between this repository and rmm because rapids-cmake provides a centralized `rapids_cpm_rmm` function. Unfortunately, because of how certain tests in this repo were set up, they could not be tested against a fork (like #765 did) without manually copying over branches to the fork, so that was not done. Now that the rmm PR is merged, we can see that there are a few issues to be ironed out with the tests here. In particular, since rmm no longer clones fmt, the pin testing needs to be updated to not look for it.

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 #=============================================================================
-# Copyright (c) 2021-2025, NVIDIA CORPORATION.
+# Copyright (c) 2021-2024, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -27,6 +27,7 @@
 # Enforce the minimum required CMake version for all users
 cmake_minimum_required(VERSION 3.26.4 FATAL_ERROR)
 
+set(rapids-cmake-dir "test")
```
vyasr, line 30:
```suggestion
```

bdice (approved):

PR #776, opened 2025-02-11
Set `-Wno-deprecated-gpu-targets` when compiling for RAPIDS architectures

### LINKED ISSUE #761: Temporarily add "-Wno-deprecated-gpu-targets" to `set_runtime`
As of CUDA 12.8, when compiling for sm75 nvcc prints a warning 
```
nvcc warning : Support for offline compilation for architectures prior to '<compute/sm/lto>_75' will be removed in a future release (Use -Wno-deprecated-gpu-targets to suppress warning).
```
We could add the suppression flag to all of RAPIDS, but given that we have a centralized way of linking to the CUDA runtime anyway perhaps we should simply set the compile flag [in `rapids_cuda_set_runtime` where we link to the CUDA runtime](https://github.com/rapidsai/rapids-cmake/blob/branch-25.02/rapids-cmake/cuda/set_runtime.cmake#L66).

This change would have the nice side effect of immediately telling us if any RAPIDS project is not using `rapids_cuda_set_runtime`, which could lead to unrelated subtle bugs around linking the CUDA targets unless they do things just right.

bdice:
## Description
Closes #761.

## Checklist

### DIFF
--- a/rapids-cmake/cuda/set_architectures.cmake
+++ b/rapids-cmake/cuda/set_architectures.cmake
@@ -83,13 +83,13 @@ function(rapids_cuda_set_architectures mode)
     rapids_cuda_detect_architectures(supported_archs CMAKE_CUDA_ARCHITECTURES)
 
     list(TRANSFORM CMAKE_CUDA_ARCHITECTURES APPEND "-real")
+  endif()
 
-    # CUDA 12.8.0 and later warns when compiling for arch 70. We ignore this warning when compiling
-    # for RAPIDS architectures.
-    if(CMAKE_CUDA_COMPILER_ID STREQUAL "NVIDIA" AND CMAKE_CUDA_COMPILER_VERSION
-                                                    VERSION_GREATER_EQUAL 12.8.0)
-      string(APPEND CMAKE_CUDA_FLAGS " -Wno-deprecated-gpu-targets")
-    endif()
+  # CUDA 12.8.0 and later warns when compiling for arch 70. We ignore this warning when compiling
+  # for RAPIDS architectures.
+  if(CMAKE_CUDA_COMPILER_ID STREQUAL "NVIDIA" AND CMAKE_CUDA_COMPILER_VERSION VERSION_GREATER_EQUAL
+                                                  12.8.0)
+    string(APPEND CMAKE_CUDA_FLAGS " -Wno-deprecated-gpu-targets")
   endif()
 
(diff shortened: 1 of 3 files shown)


rapids-cmake/cuda/set_architectures.cmake
```diff
@@ -83,6 +83,13 @@ function(rapids_cuda_set_architectures mode)
     rapids_cuda_detect_architectures(supported_archs CMAKE_CUDA_ARCHITECTURES)
 
     list(TRANSFORM CMAKE_CUDA_ARCHITECTURES APPEND "-real")
+
+    # CUDA 12.8.0 and later warns when compiling for arch 70. We ignore this warning when compiling
+    # for RAPIDS architectures.
+    if(CMAKE_CUDA_COMPILER_ID STREQUAL "NVIDIA" AND CMAKE_CUDA_COMPILER_VERSION
+                                                    VERSION_GREATER_EQUAL 12.8.0)
+      set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -Wno-deprecated-gpu-targets")
```
bdice, line 91: This doesn't seem to work for me. I am guessing the issue is with variable scope but I don't know how to resolve it.

bdice: Oh, I bet I wasn't using `RAPIDS` as my arches. 🤦 I will try again.

robertmaynard: You need to use `PARENT_SCOPE` for this to propagate up a level

robertmaynard: Style wise it should look like:
```
string(APPEND CMAKE_CUDA_FLAGS " -Wno-deprecated-gpu-targets")

.....
# Down on line 113
set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS}" PARENT_SCOPE)

```

robertmaynard: Pushed the needed changes to your PR

vyasr:
@robertmaynard do we need to do any further testing of this or are you good with merging it as-is?

vyasr (approved):

bdice (commented):
Thanks @robertmaynard -- I approve your changes, so I will merge this. We can let nightlies rebuild the sccache over the weekend.

PR #783, opened 2025-02-24
Add support for components in Cython `rapids_cython_create_modules`

pentschev:
## Description
Add support for components in Cython `rapids_cython_create_modules`, introducing a new `COMPONENT` option that installs that component if specified.

This was discussed offline with @vyasr in the context of https://github.com/rapidsai/ucxx/pull/376 where we needed to build Cython tests and make them available but without distributing in the primary package, and thus we create a new package `ucxx-tests` to be consumed primarily by CI. Since most of the project is already architected using CMake components it seemed the correct approach to make the Cython tests be also built using a CMake component, and the changes here are confirmed working by that PR, where the new option is used [here](https://github.com/rapidsai/ucxx/pull/376/files#diff-7dcf8d7be7a5de0b642dcba3cbd9a54a837b1ca99ef9b071a74c0e292c48a61bR14).

Given my very limited knowledge of CMake it is likely that the solution proposed in this PR is not the best nor covers more advanced pitfalls, therefore I would appreciate any guidance to improve or fix anything necessary. I would also appreciate suggestions on what would be the appropriate way to write tests for this, I believe I might need to adapt parts of [export-verify-install-components](https://github.com/rapidsai/rapids-cmake/blob/branch-25.04/testing/export/export-verify-install-components/CMakeLists.txt) and add it to [cython-core_create_modules](https://github.com/rapidsai/rapids-cmake/blob/branch-25.04/testing/cython-core/cython-core_create_modules/CMakeLists.txt), but if there are better solutions, please let me know. Finally it's not clear to me whether I should add the new `COMPONENT` option to [`cmake-format.json`](https://github.com/rapidsai/rapids-cmake/blob/b01af4594684c5639e18172428402d1df0affe77/cmake-format-rapids-cmake.json#L361-L370) since it doesn't seem to include all the ones that currently exist, as [`MODULE_PREFIX` and `ASSOCIATED_TARGETS`](https://github.com/rapidsai/rapids-cmake/blob/b01af4594684c5639e18172428402d1df0affe77/rapids-cmake/cython-core/create_modules.cmake#L31-L32) are missing from it.

## Checklist

### DIFF
--- a/testing/cython-core/cython-core_create_modules_component/check_installed_components.cmake
+++ b/testing/cython-core/cython-core_create_modules_component/check_installed_components.cmake
@@ -14,14 +14,18 @@
 # limitations under the License.
 #=============================================================================
 
-add_cmake_config_test(rapids-cython-core.cmake)
+file(GLOB _install_component1_files ${CMAKE_CURRENT_BINARY_DIR}/install-component1/*.so)
+file(GLOB _install_component2_files ${CMAKE_CURRENT_BINARY_DIR}/install-component2/*.so)
 
-add_cmake_config_test(cython-core_init.cmake)
-add_cmake_config_test(cython-core_create_modules_errors.cmake SHOULD_FAIL "You must call rapids_cython_init before calling this function")
-
-add_cmake_build_test(cython-core_create_modules)
-add_cmake_config_test(cython-core_create_modules_with_library)
-add_cmake_config_test(cython-core_create_modules_with_prefix)
-add_cmake_build_test(cython-core_create_modules_component)
-
-add_cmake_config_test(cython-core_add_rpath_entries)
+if(NOT _install_component1_files MATCHES "test1")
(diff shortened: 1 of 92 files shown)


bdice (changes requested):
Oops -- just realized this didn't have tests. Can we add tests here?

testing/cython-core/cython-core_create_modules_component/check_installed_components.cmake
```diff
@@ -0,0 +1,31 @@
+#=============================================================================
+# Copyright (c) 2023-2025, NVIDIA CORPORATION.
```
bdice, line 2:
```suggestion
# Copyright (c) 2025, NVIDIA CORPORATION.
```

bdice (approved):
Just one small copyright fix. Thanks @KyleFromNVIDIA for the tests, this looks good to me.

PR #790, opened 2025-03-05
Remove now superfluous fmt patch

vyasr:
## Description

With the migration to rapids-logger in 25.04, we no longer compile fmt by default, and in cases where we wish to do so explicitly (i.e. when we turn off the symbol controls) we still only compile it in host TUs. Therefore, we can drop this patch, which currently precludes the usage of a system fmt library if one has already been built. Non-RAPIDS consumers who need to build fmt in nvcc without warnings can apply this patch themselves with an override.

## Checklist


vyasr:
I'm expecting that we have some test somewhere that is relying on this and will need to be updated. I'm running the test suite locally to verify.

vyasr:
Actually never mind, everything passes for me locally.

bdice (approved):
Should we also deprecate rapids_cpm_fmt?

robertmaynard (approved):

robertmaynard:
> Should we also deprecate rapids_cpm_fmt/spdlog?

Don't tempt me with a good time :) 

But seriously yes we can add deprecation in 25.04/25.06 and plan on removing them around 25.10. I do expect some pushback from non RAPIDS projects so we will want a longer deprecation window

PR #792, opened 2025-03-06
Also download fmt when building nvbench statically

vyasr:
## Description

This change ensures that we do not wind up with benchmark binaries that statically link nvbench only to dynamically link to an fmt binary in the build environment.

## Checklist


vyasr:
I have this up as a PoC with https://github.com/rapidsai/cudf/pull/18177. The problem now is that we're back to the state without the patches removed in https://github.com/rapidsai/rapids-cmake/pull/790, which is that we get warnings when compiling with nvcc. That is annoying for cudf benchmarks but does not currently block them because we don't compile with warnings as errors. It might be more of an issue for other consumers of nvbench itself, which [is also using rapids-cmake to fetch fmt](https://github.com/NVIDIA/nvbench/blob/main/cmake/NVBenchDependencies.cmake#L10). 

We're going to have to make a call on the tradeoffs here.
- If we revert #790, then we basically preclude the usage of preexisting fmt libraries in all cases (the use case described in rapidsai/cudf#16951) because [the existence of a patch command is sufficient for rapids_cpm_find to trigger downloading](https://github.com/rapidsai/rapids-cmake/blob/branch-25.04/rapids-cmake/cpm/find.cmake#L84).
- If we go with the current route, we may be causing more pain than I originally anticipated with #790 since I forgot that we have a concrete consumer (nvbench) that does rely on using fmt in device TUs compiled by nvcc.

I'm open to either solution. In the long run if we want to devendor fmt (and spdlog) from rapids-cmake we should get the fix upstreamed to fmt itself, and probably maintain the patch in nvbench in the interim.

vyasr:
OK for now I'm going with the solution of forcing `rapids_cpm_nvbench` to download fmt when we're building nvbench statically, and also suppressing the nvcc warnings if we have cloned fmt. Once we get the patch upstreamed to fmt and can update nvbench to use that version, we can remove the warnings suppression. We will want to keep the force downloading for the foreseeable future since we want to ensure that a static build of nvbench does not introduce a dynamic library dependency on fmt. An alternative would be to modify nvbench upstream to use the header-only version of fmt, which would also solve this problem.

robertmaynard (approved):

PR #795, opened 2025-03-13
Update gtest version

vyasr:
## Description

The version of gtest that we are using (1.13) is no longer compatible with the version of gtest that is default in conda-forge (1.15.2). The issue is in "private APIs" changing: gtest is intended to be used via its macros, so directly calling some of the functions whose APIs have changed is not supported, but the problem is that the macros produce code using those functions. The second problem is that gtest and gmock mark their includes as isystem when they are built from source, which means that when building inside any environment where gtest is installed to system locations, if that version of gtest has incompatible versions of these APIs we wind up using the headers from the system installation (e.g. conda's) but then we link to our local build, resulting in linker errors. That is the issue that is currently arising. Due to the number of warnings that would surface if we tried to change the `-isystem` to `-I` for gtest, our best option is to just update gtest rather than muck with any of that. Since all of RAPIDS builds a static copy of gtest and avoids bundling it now, it should be safe to upgrade the version now.

## Checklist

### DIFF
--- a/rapids-cmake/cpm/versions.json
+++ b/rapids-cmake/cpm/versions.json
@@ -42,7 +42,7 @@
       "git_tag": "${version}"
     },
     "GTest": {
-      "version": "1.13.0",
+      "version": "1.15.2",
       "git_url": "https://github.com/google/googletest.git",
       "git_tag": "v${version}"
     },
--- a/rapids-cmake/cpm/versions.json
+++ b/rapids-cmake/cpm/versions.json
@@ -42,7 +42,7 @@
       "git_tag": "${version}"
     },
     "GTest": {
-      "version": "1.15.2",
+      "version": "1.16.0",
       "git_url": "https://github.com/google/googletest.git",
       "git_tag": "v${version}"
     },


rapids-cmake/cpm/versions.json
```diff
@@ -42,7 +42,7 @@
       "git_tag": "${version}"
     },
     "GTest": {
-      "version": "1.13.0",
+      "version": "1.15.2",
```
bdice, line 45: Why not `1.16.0`?

bdice (approved):
Why not 1.16.0? What testing is needed for this? Otherwise this is fine.

rapids-cmake/cpm/versions.json
vyasr: I'm happy to switch over to 1.16. I'll test that now.

vyasr:
> Why not 1.16.0? What testing is needed for this? Otherwise this is fine.

As far as testing, I have rebuilt rmm, cudf, kvikio tests with the latest gtest and they seem to be fine. I think that is sufficient testing given that gtest is generally stable w.r.t. its API and we're only running into this today because of incompatibilities in the internals across versions. If there is additional testing that you believe is necessary, please let me know.

I should clarify that this upgrade is currently necessary to unblock local builds in devcontainers if you rebuild your environment because [gtest 1.15 is being hard-pinned by msgpack-c](https://anaconda.org/conda-forge/msgpack-c/files) and causing this problem.

rapids-cmake/cpm/versions.json
vyasr: 1.16 also worked fine for the build. I'll bump.

Issue #799, opened 2025-03-20
[FEA] Add "build patch" functionality


KyleFromNVIDIA:
**Is your feature request related to a problem? Please describe.**
As it stands right now, any time a project has a patch, that forces the project to be built and not found. However, some patches are only related to the build process and not the runtime functionality. It doesn't make sense for a "build patch" to force the project to be built.

**Describe the solution you'd like**
We should add "build patch" functionality. When a patch only affects the build process, it should not force the project to be built. The patch should only be applied if the package can't be found and the project must be built.

**Describe alternatives you've considered**
Working around project build issues in other ways, like reordering them in such a way that makes them build properly.

**Additional context**
https://github.com/rapidsai/cuspatial/pull/1550

PR #801, opened 2025-03-21
Add build patch functionality

KyleFromNVIDIA:
## Description
This enables patches that only affect the build process of a project and not its exported headers or runtime functionality. If all of a project's patches are build patches, it is not required to be downloaded and can be found with `find_package()`.

Resolves https://github.com/rapidsai/rapids-cmake/issues/799

TODO:

- [x] Add tests for `rapids_cpm_find()`

## Checklist


robertmaynard (changes requested):
Docs need updating for the `build` entry in the json file

rapids-cmake/cpm/detail/generate_patch_command.cmake
```diff
@@ -78,6 +78,7 @@ function(rapids_cpm_generate_patch_command package_name version patch_command)
   set(patch_files_to_run)
   set(patch_issues_to_ref)
   set(patch_required_to_apply)
+  set(${build_patch_only} BUILD_PATCH_ONLY PARENT_SCOPE)
```
robertmaynard, line 81: I think the default needs to be that this is unset. Otherwise existing json files that exist ( legion,  spark-rapids ) will start to behave differently compared to previous rapids-cmake releases

KyleFromNVIDIA: See farther down. If _any_ patch doesn't have `"build": true` (which is all of the ones that currently exist) then this variable gets unset.

robertmaynard (approved):

miscco:
I find that this breaks us building raft in the CCCL CI, I have provided a patch that adds a false argument to the function call https://github.com/rapidsai/raft/pull/2619

PR #804, opened 2025-03-24
Fix export of nvtx3

KyleFromNVIDIA:
## Description
Instead of exporting to `nvtx3-targets`, export to the exports specified by the arguments.

## Checklist


AyodeAwe (approved):

rapids-cmake/cpm/versions.json
```diff
@@ -67,8 +67,8 @@
     },
     "nvtx3": {
       "version": "3.1.0",
-      "git_url": "https://github.com/NVIDIA/NVTX.git",
-      "git_tag": "v${version}"
+      "git_url": "https://github.com/bdice/NVTX.git",
+      "git_tag": "cmake-relocatable"
```
KyleFromNVIDIA, line 71: This will need to be reverted before merging.

bdice: Will we need to wait until NVTX releases a new version with the changes from https://github.com/NVIDIA/NVTX/pull/114? Or would we be comfortable with having RAPIDS pin to an as-yet unreleased SHA of NVTX?

KyleFromNVIDIA: I believe we need the changes from https://github.com/NVIDIA/NVTX/pull/114, and I believe those changes are also incompatible with rapids-cmake prior to merging this PR. In other words, it has to be done all at once (unless we backport https://github.com/NVIDIA/NVTX/pull/114 as a patch to this repo.)

KyleFromNVIDIA: Oh, I see what you mean.

I'd be fine with pinning to a specific SHA until they release the new version.

bdice: Okay cool! Let's do that, once https://github.com/NVIDIA/NVTX/pull/114 is merged. If we are blocked and need to fix something in RAPIDS without waiting for that to merge, I would accept pinning to a fork of NVTX (mine, I guess) for the short term.

KyleFromNVIDIA: I don't think we're blocked at the moment. I was able to come up with a workaround that left NVTX as-is, but we definitely still want to fix this.

bdice (approved):
Approving based on https://github.com/rapidsai/rapids-cmake/pull/804/files#r2021893912.

rapids-cmake/cpm/versions.json
robertmaynard: If we are going to pin to a fork, let's put it in the RAPIDS org instead of a user fork ( and pin to a SHA1 as well please ).

KyleFromNVIDIA: We're not pinning to a fork. We're going to get https://github.com/NVIDIA/NVTX/pull/114 merged and then pin to the SHA of that merge.

KyleFromNVIDIA: This PR has been updated with the merged SHA. I think this is ready to merge.

ttnghia:
This PR seems to cause issue with our submodule sync in spark-rapids-jni. The problem is that, NVTX#114 is merged into the `dev` branch, not the default branch (`release-v3`). When doing submodule sync, rapids-cmake script `rapids_cpm_pinning_write_file` tries to get the latest commit from the default branch, which breaks the build.

PR #807, opened 2025-03-25
Define `-DCUB_DISABLE_NAMESPACE_MAGIC` and `-DCUB_IGNORE_NAMESPACE_MAGIC_ERROR` in `rapids_cpm_cccl`

trxcllnt:
## Description
Adds additional definitions to also disable CUB's arch-dependent namespaces

## Checklist

### DIFF
--- a/rapids-cmake/cpm/cccl.cmake
+++ b/rapids-cmake/cpm/cccl.cmake
@@ -159,10 +159,14 @@ function(rapids_cpm_cccl)
   if(TARGET CCCL::CCCL)
     # Can be removed once we move to CCCL 2.3
     #
+    target_compile_definitions(CCCL::CCCL INTERFACE CUB_DISABLE_NAMESPACE_MAGIC)
+    target_compile_definitions(CCCL::CCCL INTERFACE CUB_IGNORE_NAMESPACE_MAGIC_ERROR)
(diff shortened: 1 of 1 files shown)


robertmaynard (commented):
I believe that this is fine given the fact that we already pass  `THRUST_DISABLE_ABI_NAMESPACE` and have no impact

robertmaynard:
@miscco For heads up as he is looking at these namespaces defines from the CCCL side.

IIRC @trxcllnt wants this so that we get more cache hits on our CCCL/RAPIDS builds

rapids-cmake/cpm/cccl.cmake
```diff
@@ -159,10 +159,14 @@ function(rapids_cpm_cccl)
   if(TARGET CCCL::CCCL)
     # Can be removed once we move to CCCL 2.3
```
bdice, line 160: This comment seems outdated. @trxcllnt Should it be removed?

trxcllnt: I don't know

robertmaynard: The comment is old and should be removed, the code not so much

trxcllnt:
@bdice @robertmaynard can we merge this?

robertmaynard (approved):

PR #809, opened 2025-03-31
Allow retrying downloads on network failures

pentschev:
## Description
Retrying to download files is useful when there are network instabilities that may succeed after a few retries.

## Checklist


rapids-cmake/cmake/download_with_retry.cmake
```diff
@@ -0,0 +1,90 @@
...
+
+  while(NOT download_success AND current_retry LESS ${_RAPIDS_MAX_RETRIES})
+    if(current_retry GREATER 0)
+      message(STATUS "Retrying download (attempt ${current_retry} of ${_RAPIDS_MAX_RETRIES}) after ${_RAPIDS_RETRY_DELAY} seconds..."
+      )
+      execute_process(COMMAND ${CMAKE_COMMAND} -E sleep ${_RAPIDS_RETRY_DELAY})
+    endif()
+
+    # Remove any existing file to ensure clean download
+    if(EXISTS "${output_file}")
```
robertmaynard, line 63: We should require a MD5 hash be provided to this function. If a file already exists and the MD5 hash doesn't match we download again.

Otherwise code that calls this function and doesn't first do a proper existence check will re-download the file every time CMake is run.

rapids-cmake/cmake/download_with_retry.cmake
```diff
@@ -0,0 +1,90 @@
...
+      file(REMOVE "${output_file}")
+    endif()
+
+    file(DOWNLOAD "${url}" "${output_file}" LOG download_log)
+
+    # Check if file exists and has content
+    if(EXISTS "${output_file}")
+      file(SIZE "${output_file}" file_size)
```
robertmaynard, line 71: Use the MD5 hash instead of file_size as we have had partially downloaded files before

robertmaynard (changes requested):
We need tests added as well.

Some tests:

- Tests can first create a bad file at location and call `rapids-cmake/cmake/download_with_retry.cmake` to validate the proper file is downloaded.
- As the `output_file` is user provided we should also try downloading two different urls to the same output location. We can validate the second file is properly downloaded over the existing file.
- Validate that if the file already exists ( via md5 match ) we don't download anything at all. Done via a non-existant URL, and a known file + hash

rapids-cmake/cmake/download_with_retry.cmake
pentschev: Should we provide an _optional_ MD5? The problem with MD5 although definitely more reliable is its use in other projects, we must then ensure all projects keep an up-to-date MD5 for `RAPIDS.cmake`, which is what I'm also currently using this file for projects depending on rapids-cmake. Alternatively, would it be possible to ensure `RAPIDS.cmake` is rarely/never changed?

pentschev (commented):
Thanks @robertmaynard , I'll see if I can also succeed in writing some tests, also have responded to your MD5 suggestion, please let me know what you think.

rapids-cmake/cmake/download_with_retry.cmake
pentschev: Currently the problems observed were in downloading CPM (addressed in this PR) and `RAPIDS.cmake` in other RAPIDS projects. In those projects we have:

```cmake
  file(
    DOWNLOAD
    "https://raw.githubusercontent.com/rapidsai/rapids-cmake/branch-${RAPIDS_VERSION_MAJOR_MINOR}/RAPIDS.cmake"
    "${CMAKE_CURRENT_BINARY_DIR}/RAPIDSMP_RAPIDS-${RAPIDS_VERSION_MAJOR_MINOR}.cmake"
  )
```

Which can now `include("../cmake/download_with_retry.cmake")` (a copy of the file needs to be part of the project itself) and then replace the above with:

```
  rapids_download_with_retry(
    "https://raw.githubusercontent.com/rapidsai/rapids-cmake/branch-${RAPIDS_VERSION_MAJOR_MINOR}/RAPIDS.cmake"
    "${CMAKE_CURRENT_BINARY_DIR}/RAPIDSMP_RAPIDS-${RAPIDS_VERSION_MAJOR_MINOR}.cmake"
  )
```

So far the changes above seemed sufficient in the cases I've been personally observing and testing, but there could indeed be more to prove your chicken and egg problem statement. I do agree this is not overall the optimal solution (that would likely be not to have network instabilities in the first place), but it's been very disruptive for several people to deal with those so this solution seems better than the status quo at least.

One hack I can think of to deal with the above is to always download the `RAPIDS.cmake` file, and since the projects need to download the proper version and know what version that is (because they know `RAPIDS_VERSION_MAJOR_MINOR`) and then replace the version `25.06` (for example) in the file with `XX.YY`, then use that MD5 always. However, that approach seems redundant since we would always need to know the MD5, which implies in this case we know its content as well and could just copy `RAPIDS.cmake` itself into the project and then just keep the branch up-to-date with the value from `RAPIDS_VERSION_MAJOR_MINOR`, but that would mean `RAPIDS.cmake` is ensured not to be modified (or modified everywhere simultaneously). Not sure if this is a commitment rapids-cmake can make.

robertmaynard: At this point it could be better to copy `RAPIDS.cmake` into the projects instead of adding another file that helps with download retries. 

We absolutely cann't be downloading the file everytime a user runs cmake, so some form of hash is required to properly stop that from occuring.

pentschev: I've made the changes to add MD5 validation. The tests will now fail because it relies on downloading files that don't exist yet because the PR hasn't been merged yet (please let me know if you have a better idea of what to test such that we don't need to rely on a dual-PR change or something of that sort).

As for including a copy of `RAPIDS.cmake`, I think that would be a reasonable change to do for all of RAPIDS projects then. How about the following workflow:

1. Copy `RAPIDS.cmake` into all projects;
2. When cutting new development branches, `ci/release/update-version.sh` updates the branch of the `RAPIDS.cmake` file included in the directory and copyright year;
3. To provide protection, check on every PR that the contents of `RAPIDS.cmake` haven't changed, if they did someone would need to manually update them.

The workflow above will cause minimal breakage provided nothing more than the branch/year in `RAPIDS.cmake` is changed on every release, anything more than that would require a more coordinated/centralized effort to ensure all repos have been properly updated.

WDYT?

robertmaynard: As for the tests have them for now use URL to your fork of rapids-cmake. After the tests pass / we sign off on the PR we can revert those changes and merge the PR

robertmaynard: @KyleFromNVIDIA @jameslamb as they have more experience on file verification in CI .

pentschev: Adjusted the URL in https://github.com/rapidsai/rapids-cmake/pull/809/commits/879d6f168674c5591c590c7cb341f101bd0e7633, tests should pass now.

KyleFromNVIDIA: This approach sounds fine to me. I'm not sure yet of the best way to make sure `RAPIDS.cmake` is up to date. Maybe a new pre-commit hook, or a new shared workflow?

rapids-cmake/cmake/download_with_retry.cmake
```diff
@@ -0,0 +1,94 @@
...
+    # Remove any existing file to ensure clean download
+    if(EXISTS "${output_file}")
+      file(REMOVE "${output_file}")
+    endif()
+
+    file(DOWNLOAD "${url}" "${output_file}" LOG download_log)
+
+    # Check if file exists and validate MD5
+    if(EXISTS "${output_file}")
+      file(MD5 "${output_file}" downloaded_md5)
```
KyleFromNVIDIA, line 74: While MD5 is considered "good enough" for simple file integrity checks that don't require protection against malicious actors, I consider any use of it to be bad practice in general and am loathe to sign off on it. Other ciphers, like SHA-256, are widely available and cost nothing extra to use. Please switch to SHA-256.

KyleFromNVIDIA (commented):
I think the overall approach is fine. Just a few minor comments.

rapids-cmake/cmake/download_with_retry.cmake
pentschev: I was thinking simply of an error when building and `RAPIDS.cmake` is outdated. I didn't think of a pre-commit hook but that definitely sounds like a better solution, is there a simple way to write a new pre-commit hook for that purpose? I've never done that so I'm clueless what kind of effort that would take, similar to a shared workflow, no idea there either.

rapids-cmake/cmake/download_with_retry.cmake
pentschev: Agreed, switched to sha256 in https://github.com/rapidsai/rapids-cmake/pull/809/commits/3843dd5d861bcf88521ac5fa5ae71ef9d423e59c .

pentschev:
@KyleFromNVIDIA could you help me figure out how to resolve the issues? This is what is failing:

<details>

```
 372/1110 Test #1029: cpm_nvtx-build-config-subdir-ninja_multi-config .............................................***Failed    0.37 sec
[1/1] cd /__w/rapids-cmake/rapids-cmake/build/cpm/cpm_nvtx-build-config-subdir-ninja_multi-config-build && /opt/conda/envs/test/bin/cmake -E rm -rf /__w/rapids-cmake/rapids-cmake/build/cpm/cpm_nvtx-build-config-subdir-ninja_multi-config-build/check_uses_nvtx_dir/build && /opt/conda/envs/test/bin/cmake -S="/__w/rapids-cmake/rapids-cmake/build/cpm/cpm_nvtx-build-config-subdir-ninja_multi-config-build/check_uses_nvtx_dir" -B="/__w/rapids-cmake/rapids-cmake/build/cpm/cpm_nvtx-build-config-subdir-ninja_multi-config-build/check_uses_nvtx_dir/build"
FAILED: CMakeFiles/verify_build_config-Debug /__w/rapids-cmake/rapids-cmake/build/cpm/cpm_nvtx-build-config-subdir-ninja_multi-config-build/CMakeFiles/verify_build_config-Debug 
cd /__w/rapids-cmake/rapids-cmake/build/cpm/cpm_nvtx-build-config-subdir-ninja_multi-config-build && /opt/conda/envs/test/bin/cmake -E rm -rf /__w/rapids-cmake/rapids-cmake/build/cpm/cpm_nvtx-build-config-subdir-ninja_multi-config-build/check_uses_nvtx_dir/build && /opt/conda/envs/test/bin/cmake -S="/__w/rapids-cmake/rapids-cmake/build/cpm/cpm_nvtx-build-config-subdir-ninja_multi-config-build/check_uses_nvtx_dir" -B="/__w/rapids-cmake/rapids-cmake/build/cpm/cpm_nvtx-build-config-subdir-ninja_multi-config-build/check_uses_nvtx_dir/build"
-- The CXX compiler identification is GNU 10.4.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /opt/conda/envs/test/bin/x86_64-conda-linux-gnu-c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Error at /__w/rapids-cmake/rapids-cmake/build/cpm/cpm_nvtx-build-config-subdir-ninja_multi-config-build/uses-nvtx3-dependencies.cmake:104 (include):
  include could not find requested file:
    /cmake/download_with_retry.cmake
Call Stack (most recent call first):
  /__w/rapids-cmake/rapids-cmake/build/cpm/cpm_nvtx-build-config-subdir-ninja_multi-config-build/uses-nvtx3-dependencies.cmake:119 (rapids_cpm_download)
  /__w/rapids-cmake/rapids-cmake/build/cpm/cpm_nvtx-build-config-subdir-ninja_multi-config-build/uses-nvtx3-config.cmake:67 (include)
  CMakeLists.txt:6 (find_package)
CMake Error at /__w/rapids-cmake/rapids-cmake/build/cpm/cpm_nvtx-build-config-subdir-ninja_multi-config-build/uses-nvtx3-dependencies.cmake:105 (rapids_download_with_retry):
  Unknown CMake command "rapids_download_with_retry".
Call Stack (most recent call first):
  /__w/rapids-cmake/rapids-cmake/build/cpm/cpm_nvtx-build-config-subdir-ninja_multi-config-build/uses-nvtx3-dependencies.cmake:119 (rapids_cpm_download)
  /__w/rapids-cmake/rapids-cmake/build/cpm/cpm_nvtx-build-config-subdir-ninja_multi-config-build/uses-nvtx3-config.cmake:67 (include)
  CMakeLists.txt:6 (find_package)
-- Configuring incomplete, errors occurred!
ninja: build stopped: subcommand failed.
```

</details>

I think those have been added in https://github.com/rapidsai/rapids-cmake/pull/804 earlier today, that's why it didn't fail before. It seems that `rapids_download_with_retry` is not being included properly, what is the correct way of including it? They do not directly use `rapids_download_with_retry`, should the include come indirectly from another include or does it need to be explicitly included?

KyleFromNVIDIA:
The problem is in `rapids-cmake/export/write_dependencies.cmake`. It's reading the contents of `rapids-cmake/cpm/detail/download.cmake` and writing them out to the dependencies file. Unfortunately, that file now expects to include `download_with_retry.cmake` with this PR, but that file hasn't been installed.

pentschev:
Thanks @KyleFromNVIDIA , do you think https://github.com/rapidsai/rapids-cmake/pull/809/commits/f72cca27decdad53eeb684688a6492e36d346c1a is the proper way to solve this?

KyleFromNVIDIA:
The problem with f72cca2 is that even though the contents of `download_with_retry.cmake` have been written into the dependencies file, it's still trying to include `download_with_retry.cmake`, which doesn't exist in the install context. You'll need to do something clever, perhaps either guarding the `include()` behind a variable or making it `OPTIONAL`. I'd also suggest putting it before the contents of `download.cmake`.

rapids-cmake/cpm/detail/download.cmake
```diff
@@ -83,15 +83,12 @@ function(rapids_cpm_download)
...
     message(VERBOSE "Downloading CPM.cmake to ${CPM_DOWNLOAD_LOCATION}")
-    file(DOWNLOAD
-         https://github.com/cpm-cmake/CPM.cmake/releases/download/v${CPM_DOWNLOAD_VERSION}/CPM.cmake
-         ${CPM_DOWNLOAD_LOCATION} LOG download_log)
-
-    file(MD5 ${CPM_DOWNLOAD_LOCATION} cpm_hash)
-    if(NOT cpm_hash STREQUAL CPM_DOWNLOAD_MD5_HASH)
-      message(FATAL_ERROR "CPM.cmake hash mismatch [got=${cpm_hash} expected=${CPM_DOWNLOAD_MD5_HASH}] to download details below\n ${download_log}"
-      )
+    if(NOT DEFINED rapids_download_with_retry_included)
```
KyleFromNVIDIA, line 86: This is more or less the approach that I had in mind, and I'm fine with leaving it as it is. However, I just realized we can use the `if(COMMAND)` predicate to make it even simpler:

```suggestion
    if(NOT COMMAND rapids_download_with_retry)
```

rapids-cmake/export/write_dependencies.cmake
```diff
@@ -74,6 +74,14 @@ function(rapids_export_write_dependencies type export_set file_path)
   # Do we need a Template header?
   set(_RAPIDS_EXPORT_CONTENTS)
   if(uses_cpm)
+    # Include download_with_retry.cmake in the exported dependencies first
+    file(READ "${CMAKE_CURRENT_FUNCTION_LIST_DIR}/../cmake/download_with_retry.cmake"
+         download_with_retry_logic)
+    string(APPEND _RAPIDS_EXPORT_CONTENTS ${download_with_retry_logic})
+    string(APPEND _RAPIDS_EXPORT_CONTENTS
+           "\n# Guard to prevent including download_with_retry.cmake again\n")
+    string(APPEND _RAPIDS_EXPORT_CONTENTS "set(rapids_download_with_retry_included TRUE)\n\n")
```
KyleFromNVIDIA, line 83: If you implement the approach I described above, this can be removed:

```suggestion
```

KyleFromNVIDIA (approved):
One small suggestion for bonus points. Approving so you're not blocked.

rapids-cmake/export/write_dependencies.cmake
KyleFromNVIDIA, line 83: If you implement the approach I described above, you can take this out too:

```suggestion
```

pentschev:
Thanks @KyleFromNVIDIA for the suggestions, tests are now passing again! 🎉

pentschev:
Based on empirical evidence I increased the number of retries to 10. Downloading CPM requires more than 3 iterations 11% of the time, and more than 5 iterations 2% of the time.

rapids-cmake/cmake/download_with_retry.cmake
```diff
@@ -0,0 +1,94 @@
...
+rapids_download_with_retry
+--------------------------
+
+.. versionadded:: v25.06.00
+
+Downloads a file from a URL with retry logic for handling network issues.
+
+  .. code-block:: cmake
+
+    rapids_download_with_retry(url output_file sha256 [max_retries] [retry_delay])
```
bdice, line 16: Match case with the documented parameter names below. Either capitalize these or lowercase the docstring params.
```suggestion
    rapids_download_with_retry(url output_file sha256 [MAX_RETRIES] [RETRY_DELAY])
```

rapids-cmake/cmake/download_with_retry.cmake
```diff
@@ -0,0 +1,94 @@
...
+  The URL to download from.
+
+``output_file``
+  The path where the downloaded file should be saved.
+
+``sha256``
+  The expected SHA256 checksum of the file.
+
+``MAX_RETRIES``
+  Maximum number of retry attempts. Defaults to 3.
```
bdice, line 32:
```suggestion
  Maximum number of retry attempts. Defaults to 10.
```

rapids-cmake/cmake/download_with_retry.cmake
```diff
@@ -0,0 +1,94 @@
...
+
+``RETRY_DELAY``
+  Delay between retries in seconds. Defaults to 5.
```
bdice, line 35: I did a bunch of reading on retries while reviewing https://github.com/rapidsai/kvikio/pull/603 and I think we should do two things:
- Don't make this delay configurable, just use sensible defaults.
- Default behavior should be exponential backoff with a maximum, like [1, 2, 4, 4, 4, 4, ...].

Here is a ChatGPT code snippet proposal, take what you want from it: https://chatgpt.com/share/67f53797-4944-8009-9f03-2f2bdca599b4

rapids-cmake/cmake/download_with_retry.cmake
```diff
@@ -0,0 +1,94 @@
+#=============================================================================
+# Copyright (c) 2024-2025, NVIDIA CORPORATION & AFFILIATES.
+# SPDX-License-Identifier: Apache-2.0
+#=============================================================================
```
bdice, line 4:
```suggestion
#=============================================================================
include_guard(GLOBAL)
```

rapids-cmake/cpm/detail/download.cmake
```diff
@@ -83,15 +83,12 @@ function(rapids_cpm_download)
...
     message(VERBOSE "Downloading CPM.cmake to ${CPM_DOWNLOAD_LOCATION}")
-    file(DOWNLOAD
-         https://github.com/cpm-cmake/CPM.cmake/releases/download/v${CPM_DOWNLOAD_VERSION}/CPM.cmake
-         ${CPM_DOWNLOAD_LOCATION} LOG download_log)
-
-    file(MD5 ${CPM_DOWNLOAD_LOCATION} cpm_hash)
-    if(NOT cpm_hash STREQUAL CPM_DOWNLOAD_MD5_HASH)
-      message(FATAL_ERROR "CPM.cmake hash mismatch [got=${cpm_hash} expected=${CPM_DOWNLOAD_MD5_HASH}] to download details below\n ${download_log}"
-      )
+    if(NOT COMMAND rapids_download_with_retry)
```
bdice, line 86: This "if" shouldn't be needed in my understanding -- but we were missing `include_guard(GLOBAL)` in that file.

rapids-cmake/export/write_dependencies.cmake
```diff
@@ -74,6 +74,11 @@ function(rapids_export_write_dependencies type export_set file_path)
   # Do we need a Template header?
   set(_RAPIDS_EXPORT_CONTENTS)
   if(uses_cpm)
+    # Include download_with_retry.cmake in the exported dependencies first
+    file(READ "${CMAKE_CURRENT_FUNCTION_LIST_DIR}/../cmake/download_with_retry.cmake"
+         download_with_retry_logic)
+    string(APPEND _RAPIDS_EXPORT_CONTENTS ${download_with_retry_logic})
+
     file(READ "${CMAKE_CURRENT_FUNCTION_LIST_DIR}/../cpm/detail/download.cmake" cpm_logic)
```
bdice, line 82: I'm confused, I don't see `cpm/detail/download.cmake` in here: https://github.com/rapidsai/rapids-cmake/tree/branch-25.06/rapids-cmake/cmake/detail

Is this a bug?

rapids-cmake/export/write_dependencies.cmake
```diff
@@ -74,6 +74,11 @@ function(rapids_export_write_dependencies type export_set file_path)
   # Do we need a Template header?
   set(_RAPIDS_EXPORT_CONTENTS)
   if(uses_cpm)
+    # Include download_with_retry.cmake in the exported dependencies first
+    file(READ "${CMAKE_CURRENT_FUNCTION_LIST_DIR}/../cmake/download_with_retry.cmake"
```
bdice, line 78: Is adding `include_guard(GLOBAL)` going to mess with this? I don't know. Either way, we need an include guard on that file. Perhaps there is a workaround like moving the implementation to a detail file that is included by the main file with an include guard.

testing/cmake/download_with_retry-basic.cmake
```diff
@@ -0,0 +1,50 @@
...
+# Create a test directory
+file(MAKE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/download_test")
+
+# Test URL - using a static test file with known contents
+set(test_url "https://raw.githubusercontent.com/rapidsai/rapids-cmake/c0d8c09c5590ecf38a9f9897c93e686e3da1858b/testing/cmake/test_files/test1.txt")
+set(output_file "${CMAKE_CURRENT_BINARY_DIR}/download_test/test_file.txt")
+# Expected SHA256 of the test file (content: "This is a test file for rapids-cmake download tests.")
+set(expected_sha256 "7a90fe28cdb40b030ad3323af3fd9292d849db0be1d79a7000f9ff150c00161f")
+
+# Test 1: Basic download
```
bdice, line 27: Do these tests need to clean up after themselves (remove the downloaded files)?

rapids-cmake/cmake/download_with_retry.cmake
```diff
@@ -0,0 +1,94 @@
...
+
+#[=======================================================================[.rst:
+rapids_download_with_retry
```
bdice, line 7: This needs docs. Please create a new file `docs/command/rapids_cmake_download_with_retry.rst` and add it to https://github.com/rapidsai/rapids-cmake/blob/branch-25.06/docs/api.rst.

rapids-cmake/cpm/detail/download.cmake
KyleFromNVIDIA: It's needed so that we don't try to include the file when the contents of this file get exported. Without this `if()`, trying to read `download_with_retry.cmake` fails in an export context.

rapids-cmake/export/write_dependencies.cmake
KyleFromNVIDIA: `download.cmake` already has `include_guard(GLOBAL)`, so adding it to `download_with_retry.cmake` won't make a difference at this point.

testing/cmake/download_with_retry-basic.cmake
KyleFromNVIDIA: I would. I'll leave a code suggestion to do so.

testing/cmake/download_with_retry-basic.cmake
```diff
@@ -0,0 +1,50 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#=============================================================================
+include(${rapids-cmake-dir}/cmake/download_with_retry.cmake)
+
+# Create a test directory
+file(MAKE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/download_test")
```
KyleFromNVIDIA, line 19:
```suggestion
# Create a test directory
file(REMOVE_RECURSE "${CMAKE_CURRENT_BINARY_DIR}/download_test")
file(MAKE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/download_test")
```

testing/cmake/download_with_retry-overwrite.cmake
```diff
@@ -0,0 +1,50 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#=============================================================================
+include(${rapids-cmake-dir}/cmake/download_with_retry.cmake)
+
+# Create a test directory
+file(MAKE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/download_test")
```
KyleFromNVIDIA, line 19:
```suggestion
# Create a test directory
file(REMOVE_RECURSE "${CMAKE_CURRENT_BINARY_DIR}/download_test")
file(MAKE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/download_test")
```

rapids-cmake/cmake/download_with_retry.cmake
pentschev: That would invalidate the [empirical evidence](https://github.com/rapidsai/rapids-cmake/pull/809#issuecomment-2786673041), i.e., it might make `10` retries insufficient in some cases, without really speeding up things in a significant manner and increasing implementation complexity at the same time. I'd prefer to keep the 5 seconds instead if you're don't feel strongly about it.

rapids-cmake/export/write_dependencies.cmake
pentschev: So should https://github.com/rapidsai/rapids-cmake/pull/809/files#r2033382923 be included or is it better if we skip it?

KyleFromNVIDIA: I say go ahead and add it.

rapids-cmake/cmake/download_with_retry.cmake
pentschev: Also added missing api.rst changes in https://github.com/rapidsai/rapids-cmake/pull/809/commits/ee941f31d8050556ca90823e90415914886e7ed9

rapids-cmake/cmake/download_with_retry.cmake
KyleFromNVIDIA: Actually, I see now that these are keyword arguments. This should be formatted more like:

```
    rapids_download_with_retry(url output_file sha256 [MAX_RETRIES <max_retries>] [RETRY_DELAY <retry_delay>])
```

rapids-cmake/cmake/download_with_retry.cmake
```diff
@@ -0,0 +1,95 @@
...
+  The expected SHA256 checksum of the file.
+
+``MAX_RETRIES``
+  Maximum number of retry attempts. Defaults to 10.
+
+``RETRY_DELAY``
+  Delay between retries in seconds. Defaults to 5.
+
+#]=======================================================================]
+function(rapids_download_with_retry url output_file sha256)
```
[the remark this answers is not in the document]
pentschev: Done in https://github.com/rapidsai/rapids-cmake/pull/809/commits/b076953231b8db81d01724166a1ce3084bc25acd

rapids-cmake/cmake/download_with_retry.cmake
bdice: I'm not sure if I understand. The empirical evidence does not indicate whether time between attempts affects success rate.

bdice (approved):

rapids-cmake/export/write_dependencies.cmake
bdice: Thank you. I read this about 5 times to check myself and couldn't see the difference. 🙃

rapids-cmake/cmake/download_with_retry.cmake
pentschev: The empirical evidence is using 5 seconds as variable, if we change any variables from the evidence the results _may_ be different. I agree there's no guarantee the results will be different, but it's guaranteed that the evidence is invalidated.

Issue #798, opened 2025-03-18
[FEA] Set -ffile_prefix_map when in a conda-build


vyasr:
**Is your feature request related to a problem? Please describe.**
The issue that we would like to resolve is documented in rapidsai/build-planning#159.

**Describe the solution you'd like**
rapids-cmake's `rapids_cmake_support_conda_env` contains various bits of logic specialized to conda environments, and further, to building inside conda-build. Based on the discussions in rapidsai/build-planning#159, we should consider augmenting this feature to also remap the file prefix. This ensures that we do not embed non-relocatable paths into our binaries via the usage of the `__FILE__` macro. We could inject this logic into the conda-build-specific sections of `rapids_cmake_support_conda_env`.

As long as we are fine with using the root of the C++ CMakeLists.txt as the new prefix, this should be straightforward. Things will get a bit hairier if we want to remap the base prefix to be something like the root of the repository (as demonstrated in https://github.com/rapidsai/rmm/pull/1861) since that is a directory above the build, but it should still be possible

jakirkham:
As context `-fdebug-prefix-map` is set by [the Conda compilers]( https://github.com/conda-forge/ctng-compiler-activation-feedstock/blob/13e7be8cc4e4472957b5a6aad0047424d860862f/recipe/activate-g%2B%2B.sh#L91-L94 ) when run in conda-build. Including the relevant bit below

```bash
"... -fdebug-prefix-map=${SRC_DIR}=/usr/local/src/conda/${PKG_NAME}-${PKG_VERSION} -fdebug-prefix-map=${PREFIX}=/usr/local/src/conda-prefix"
```

It is worth seeing whether we want to set `-ffile-prefix-path` to affect this or just remove it and use something like `-fmacro-prefix-path` for a more focused replacement

jakirkham:
Am guessing we will want something like this in CMake

```cmake
target_compile_options(${target} INTERFACE "$<$<COMPILE_LANGUAGE:CXX>:-ffile-prefix-path=${CMAKE_CURRENT_SOURCE_DIR}/=''>")
target_compile_options(${target} INTERFACE "$<$<COMPILE_LANGUAGE:CXX>:-ffile-prefix-path=${CMAKE_CURRENT_SOURCE_DIR}/=''>")
[...]

PR #817, opened 2025-04-08
Use relative paths for `$PREFIX` in conda-build

jakirkham:
## Description

Fixes https://github.com/rapidsai/rapids-cmake/issues/798

Add compiler flags in conda-build runs to remap all absolute paths with `$PREFIX` to relative paths based on being in the `$PREFIX`. This keeps the meaningful portion of the path while dropping the build time specific `$PREFIX`. Plus bypasses the need to do prefix replacement with these paths.

<hr>

## Checklist


KyleFromNVIDIA (approved):

jakirkham:
Thanks Kyle! 🙏

@vyasr would you like to take a look as well?

bdice:
Is it possible to test this?

jakirkham:
Is there a way to checkout this PR in a build of say RMM or cuDF? If so, we could use that to test this change in downstream CI

bdice:
Yes. Here are test PRs:
- https://github.com/rapidsai/rmm/pull/1893
- https://github.com/rapidsai/cudf/pull/18511

rapids-cmake/cmake/support_conda_env.cmake
```diff
@@ -164,6 +164,17 @@ function(rapids_cmake_support_conda_env target)
...
 
+      # Make all paths like `$PREFIX/<path>` relative to `$PREFIX` so `<path>`.
+      target_compile_options(${target}
+                             INTERFACE "$<$<COMPILE_LANGUAGE:C>:-ffile-prefix-map=$ENV{PREFIX}/=''>"
+      )
+      target_compile_options(${target}
+                             INTERFACE "$<$<COMPILE_LANGUAGE:CXX>:-ffile-prefix-map=$ENV{PREFIX}/=''>"
+      )
+      target_compile_options(${target}
+                             INTERFACE "$<$<COMPILE_LANGUAGE:CUDA>:-Xcompiler=-ffile-prefix-map=$ENV{PREFIX}/=''>"
```
vyasr, line 175: Minor nit, this should work as a single command.
```suggestion
                             INTERFACE "$<$<COMPILE_LANGUAGE:C>:-ffile-prefix-map=$ENV{PREFIX}/=''>"
                             "$<$<COMPILE_LANGUAGE:CXX>:-ffile-prefix-map=$ENV{PREFIX}/=''>"
                             "$<$<COMPILE_LANGUAGE:CUDA>:-Xcompiler=-ffile-prefix-map=$ENV{PREFIX}/=''>"
```

vyasr (approved):
LGTM. One very minor suggestion but nothing blocking. I just merged latest so that I could rerun CI on the test PRs since this PR is now out of date w.r.t. the rmm layout changes, so [the cudf test job](https://github.com/rapidsai/cudf/pull/18511) is failing. Once it passes we should be good.

robertmaynard (changes requested):
The docs for `rapids_cmake_support_conda_env` need to be expanded to document this new behavior ( and what version it is added in ). See the paragraph about `-O0` for an example

rapids-cmake/cmake/support_conda_env.cmake
jakirkham: Sounds reasonable. Thanks Vyas! 🙏

rapids-cmake/cmake/support_conda_env.cmake
```diff
@@ -41,6 +41,13 @@ the behavior of conda when it builds projects.
 The `target_name` target will add the required compile flags to ensure debug builds
 are generated with `-O0` instead of the conda env default of `-O2`.
 
+.. versionadded:: v25.06.00
+
+The `-ffile-prefix-map` is now passes to remap absolute paths starting
+with `$ENV{PREFIX}` to paths relative to it. This ensures paths baked
+into binaries are relative to the environment prefix a usually
+reasonable default. It also bypasses the need to rewrite these paths.
```
jakirkham, line 49: Rob, how does this look?

jakirkham (commented):
Thanks all for the feedback!

Also thanks for testing Bradley! 🙏

Included Vyas' suggestion. Thank for mentioning that simplification

Also included a doc/release note based on Rob's feedback

Please let me know if there is anything else 🙂

rapids-cmake/cmake/support_conda_env.cmake
bdice: Suggesting some minor tweaks.
```suggestion
.. versionadded:: v25.06.00

The flag `-ffile-prefix-map` is now passed to remap absolute paths
starting with `$ENV{PREFIX}` to paths relative to it. This ensures
paths baked into binaries are relative to the environment prefix.
This prevents conda from rewriting these paths when the package is
installed.
```

robertmaynard: Can we re-phrase the first sentence to:
```
The flag `-ffile-prefix-map` is now passed to remap absolute paths starting with `$ENV{PREFIX}` to relative paths starting from `$ENV{PREFIX}` when contained in pre-processor macros like `__FILE__`. 
```

I want to make it clear that this flag doesn't impact how other compiler flags  ( e.g `-I` or `-isystem` ) are parsed, but just impacts code generation.

jakirkham: Rob, took a slightly different approach, but tried to capture the spirit of this suggestion

rapids-cmake/cmake/support_conda_env.cmake
```diff
@@ -41,6 +41,14 @@ the behavior of conda when it builds projects.
 The `target_name` target will add the required compile flags to ensure debug builds
 are generated with `-O0` instead of the conda env default of `-O2`.
 
+.. versionadded:: v25.06.00
+
+The flag `-ffile-prefix-map` is now passed to remap absolute paths starting
+with `$ENV{PREFIX}` to paths relative to it in binaries generated via
+compilation. This ensures paths baked into binaries are relative to the
+environment prefix. This prevents Conda from rewriting these paths when the
+package is installed.
```
jakirkham, line 50: How about this revision?

bdice (approved):

robertmaynard (approved):

jakirkham:
Thanks all! 🙏

PR #818, opened 2025-04-08
Add `include_guard(GLOBAL)` to download_with_retry.cmake

KyleFromNVIDIA:
## Description
And strip the include guard from download_with_retry.cmake and download.cmake when writing them to the export file.

Follow-up to https://github.com/rapidsai/rapids-cmake/pull/809

## Checklist


KyleFromNVIDIA:
That won't work. The reason that guard is there is to avoid including `download_with_retry.cmake` in an export context, because it won't know where to find the file.

PR #822, opened 2025-04-10
Make `RAPIDS.cmake` inlineable without requiring updates

KyleFromNVIDIA:
## Description
Remove the default version from `RAPIDS.cmake`, so that it no longer needs to be updated. From now on, projects that inline `RAPIDS.cmake` must provide the `rapids-cmake-version` variable.

## Checklist
- [ ] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/rapids-cmake/blob/HEAD/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.
- [ ] The `cmake-format.json` is up to date with these changes.
- [ ] I have added new files under rapids-cmake/
   - [ ] I have added include guards (`include_guard(GLOBAL)`)
   - [ ] I have added the associated docs/ rst file and update the api.rst

PR #825, opened 2025-04-10
Make `RAPIDS.cmake` inlineable without requiring updates

KyleFromNVIDIA:
## Description
Remove the default version from `RAPIDS.cmake`, so that it no longer
needs to be updated. From now on, projects that inline `RAPIDS.cmake`
must provide the `rapids-cmake-version` variable.

xref: https://github.com/rapidsai/build-infra/issues/206

Identical to https://github.com/rapidsai/rapids-cmake/pull/822, which had to be reverted due to being merged prematurely and breaking downstream projects.

## Checklist

### DIFF
--- a/RAPIDS.cmake
+++ b/RAPIDS.cmake
@@ -20,9 +20,9 @@
 cmake_minimum_required(VERSION 3.30.4 FATAL_ERROR)
 
 # Allow users to control which version is used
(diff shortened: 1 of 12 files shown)


bdice:
@KyleFromNVIDIA I think we can merge this now? I believe all the downstream changes have landed.

PR #827, opened 2025-04-15
Use cpp subdirectory for RMM.

bdice:
## Description
Updates for https://github.com/rapidsai/rmm/pull/1883.

## Checklist

### DIFF
--- a/rapids-cmake/cpm/rmm.cmake
+++ b/rapids-cmake/cpm/rmm.cmake
@@ -79,7 +79,7 @@ function(rapids_cpm_rmm)
                   CPM_ARGS
                   GIT_REPOSITORY ${repository}
                   GIT_TAG ${tag}
-                  GIT_SHALLOW ${shallow} ${patch_command}
+                  GIT_SHALLOW ${shallow} ${patch_command} SOURCE_SUBDIR cpp
                   EXCLUDE_FROM_ALL ${to_exclude}
                   OPTIONS "BUILD_TESTS OFF" "BUILD_BENCHMARKS OFF")
 


robertmaynard:
We need to make this conditional based on the `version` returned from the json information.

We can have something like SPARK using latest rapids-cmake but pinned to a rmm/cudf before the `cpp` directory existed.

bdice:
@robertmaynard Do you mean we need to check this across releases, or within the 25.06 release? Do users expect it is allowed to mix rapids-cmake 25.06 with RMM 25.04?

rapids-cmake/cpm/rmm.cmake
```diff
@@ -73,13 +73,23 @@ function(rapids_cpm_rmm)
   include("${rapids-cmake-dir}/cpm/detail/generate_patch_command.cmake")
   rapids_cpm_generate_patch_command(rmm ${version} patch_command build_patch_only)
 
+  # Use `SOURCE_SUBDIR cpp` if RMM version is 25.06 or higher
+  string(REGEX MATCH "([0-9]+)\\.([0-9]+).*" version_parts ${version})
```
robertmaynard, line 77: please use `rapids_cmake_parse_version`

robertmaynard (changes requested):

robertmaynard:
> Do users expect it is allowed to mix rapids-cmake 25.06 with RMM 25.04?

Yes. If we want to drop support for pre `src` directory versions of rmm we can add a deprecatation notice in 25.06 and remove in 25.08 or 25.10

KyleFromNVIDIA:
The `cpm_generate_pins-override` failures are due to @bdice's fork of RMM not having the `v22.02.00` tag. Those errors will go away once we go back to using mainline RMM (or if @bdice adds that tag to his fork). Still looking at the other failures.

KyleFromNVIDIA:
Looks like my commit fixed all the remaining test failures. CI is effectively clean now.

testing/cpm/cpm_generate_pins-format-patches/override.json
```diff
@@ -15,6 +15,7 @@
     },
     "rmm": {
       "version": "22.02",
+      "git_url": "https://github.com/rapidsai/rmm.git",
```
KyleFromNVIDIA, line 18: This doesn't strictly need to be here once we go back to not using @bdice's fork, but I don't think leaving it in will hurt anything. I'm ambivalent about reverting it.

testing/cpm/cpm_generate_pins-override/CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/override.json
   "packages" : {
     "rmm" : {
       "version" : "22.02",
+      "git_url" : "https://github.com/rapidsai/rmm.git",
```
KyleFromNVIDIA, line 29: This doesn't strictly need to be here once we go back to not using @bdice's fork, but I don't think leaving it in will hurt anything. I'm ambivalent about reverting it.

rapids-cmake/cpm/versions.json
```diff
@@ -78,8 +78,8 @@
     },
     "rmm": {
       "version": "${rapids-cmake-version}",
-      "git_url": "https://github.com/rapidsai/rmm.git",
-      "git_tag": "branch-${version}"
+      "git_url": "https://github.com/bdice/rmm.git",
+      "git_tag": "5685e7d180b00a139f70c88389eda786ac33d9a4"
```
KyleFromNVIDIA, line 82: This needs to be reverted before merging.

```suggestion
      "git_url": "https://github.com/rapidsai/rmm.git",
      "git_tag": "branch-${version}"
```

rapids-cmake/cpm/rmm.cmake
```diff
@@ -73,13 +73,28 @@ function(rapids_cpm_rmm)
   include("${rapids-cmake-dir}/cpm/detail/generate_patch_command.cmake")
   rapids_cpm_generate_patch_command(rmm ${version} patch_command build_patch_only)
 
+  # Use `SOURCE_SUBDIR cpp` if RMM version is 25.06 or higher
+  include("${rapids-cmake-dir}/cmake/parse_version.cmake")
+  rapids_cmake_parse_version(MAJOR ${version} major_version)
+  rapids_cmake_parse_version(MINOR ${version} minor_version)
+  if(major_version GREATER 25 OR (major_version EQUAL 25 AND minor_version GREATER_EQUAL 6))
```
KyleFromNVIDIA, line 80: Any reason not to do this?

```suggestion
  if(version VERSION_GREATER_EQUAL 25.06)
```

KyleFromNVIDIA (approved):

testing/cpm/cpm_generate_pins-format-patches/override.json
bdice: Let's leave it. This test is supposed to be pinned to an arbitrarily old version anyway.

rapids-cmake/cpm/rmm.cmake
bdice: Good idea. I just need to write like it's CMake and not like it's C preprocessor.

PR #853, opened 2025-05-22
Deprecate fmt and spdlog

### LINKED ISSUE #813: [FEA] Deprecate and remove fmt and spdlog from rapids-cpm
With the advent of https://github.com/rapidsai/rapids-logger, RAPIDS libraries no longer use fmt at all and only use spdlog transitively via rapids-logger. rapids-cmake has long has to maintain patches for fmt for various reasons, and tightly pinning fmt versions causes numerous issues as outlined in https://github.com/rapidsai/build-planning/issues/104 and various associated issues.

We should deprecate support for fmt and spdlog in 25.06 and remove them in a future version. Since spdlog support is still being used in rapids-logger, we will have to move that CMake logic into rapids-logger's CMake.

None

**Additional context**
I suspect that we will need a somewhat long deprecation period (I suggest 6 months, so removal in 26.02) because I wouldn't be surprised if there are projects that are relying on rapids_cpm_fmt at minimum. I think that is something we should accommodate a reasonable timeline for, but it is not reason enough for us keep this functionality around. While rapids-cmake is now used beyond RAPIDS for various things, I do not think the centralized store in versions.json is suitable for broader use. Users can always use `rapids_cpm_find` directly to find fmt, they will simply have to specify versions themselves (which is a good thing; they shouldn't be constrained by our specified version) and apply any patches themselves (also a good thing; our patches are not generally applicable).

vyasr:
## Description

Resolves #813 

## Checklist

### DIFF
--- a/rapids-cmake/cpm/fmt.cmake
+++ b/rapids-cmake/cpm/fmt.cmake
@@ -50,6 +50,10 @@ Result Variables
 #]=======================================================================]
 function(rapids_cpm_fmt)
   list(APPEND CMAKE_MESSAGE_CONTEXT "rapids.cpm.fmt")
+  rapids_cmake_policy(DEPRECATED_IN 25.08
+                      REMOVED_IN 26.04
+                      MESSAGE [=[`rapids_cpm_fmt` is deprecated. If you need to fetch fmt, please use rapids_cpm_find directly.]=]
+  )
 
   set(to_install OFF)
   if(INSTALL_EXPORT_SET IN_LIST ARGN)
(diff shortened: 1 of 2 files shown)


rapids-cmake/cpm/fmt.cmake
```diff
@@ -50,6 +50,11 @@ Result Variables
 #]=======================================================================]
 function(rapids_cpm_fmt)
   list(APPEND CMAKE_MESSAGE_CONTEXT "rapids.cpm.fmt")
+  include("${rapids-cmake-dir}/cmake/detail/policy.cmake")
+  rapids_cmake_policy(DEPRECATED_IN 25.08
+                      REMOVED_IN 26.04
```
bdice, line 55: 26.04 is a very long time away. We've always done removal within a 6 month window. Can we make this 25.10 or 25.12?

vyasr: I was pushing this out because the last time we deprecated something major we were asked to bring it back. I'm fine shortening this one though, I know that at least nvbench has stopped using `rapids_cpm_fmt` so maybe other consumers can be quickly convinced to do so as well. I can move to 25.12.

bdice (approved):

PR #855, opened 2025-05-29
Remove CUDA 11 support

KyleFromNVIDIA:
## Description
Issue: https://github.com/rapidsai/build-planning/issues/184

## Checklist

### DIFF
--- a/docs/packages/cuda_version_mapping.json
+++ b/docs/packages/cuda_version_mapping.json
@@ -1,5 +1,6 @@
 {
   "proprietary_binary_cuda_version_mapping": {
+    "11": "11.8",
     "12": "12.3"
   },
   "proprietary_binary": {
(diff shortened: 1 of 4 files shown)


docs/packages/cuda_version_mapping.json
```diff
@@ -1,6 +1,5 @@
 {
   "proprietary_binary_cuda_version_mapping": {
-    "11": "11.8",
```
KyleFromNVIDIA, line 3: Let's wait until most other projects have fully removed CUDA 11, otherwise they might get tripped up by this.

bdice: @KyleFromNVIDIA Are we good to go now? Or do you think we need to wait longer? (What criteria?)

robertmaynard: Legion and Matx would be things we would need to consider.

PR #864, opened 2025-06-18
rapids-cmake: Add support for a version suffix to mean using main

### LINKED ISSUE #863: [FEA] Support new RAPIDS branching model
RAPIDS is moving away from developing on release-associated branches like branch-25.08 (for example) to a single main branch on which all development occurs. This model breaks how RAPIDS libraries have fetched other RAPIDS libraries because we can no longer simply assume that code on branch-25.08 in one repository want to use the corresponding branch in another repository. This also breaks `rapids_cpm_rmm`, which effectively does the same thing using the RAPIDS version as encoded in rapids-cmake itself. To address this need, we will need a new strategy for deciding what commit hash to pull from another repository. The new approach still has to work for previously released versions of our libraries and during the burn down and code freeze phases when development is occurring simultaneously on a release branch and the main branch (which has effectively moved to the next release).

rapids-cmake should provide some standard functionality for selecting the appropriate version of a dependency to fetch. The actual `rapids_cpm_find` calls can continue to live in each repository, but rapids-cmake should provide a function `rapids_get_compatible_commit` (or whatever name) that helps select the commit-ish to pass to the find calls. That function's logic must account for the following requirements:
- When building on main during a release cycle, dependencies fetched should also be on main.
- When building old versions of a given library, dependencies fetched must be from the tag corresponding to the same release.

We also need to support building from source trees that are not git clones, which means that we cannot rely entirely on git operations to figure out distances. 

None

**Additional context**
None

robertmaynard:
## Description
To support the new RAPIDS wide branching model rapids-cmake has to support switching between the `main` branch for `rapids_cpm_rmm` and a release branch like `branch-25.06`. 

This is implemented by reading the rapids-cmake `VERSION` file and when we detect a value that ends in `a` we know we are being used from the `main` branch (a development version) and not from a release branch.

Fixes #863 

## Checklist


robertmaynard:
This is only the first step on switching more projects over to the new RAPIDS branch model. 
We still have to deploy updates to all the `VERSION` and `rapids_config.cmake` files in each project plus updates to all the release scripts to unsuffix the `VERSION` file when branching.

bdice (approved):

Issue #851, opened 2025-05-22
[FEA] Upgrade to CMP.cmake v0.42.0 to avoid FetchContent_Populate deprecation warnigns


oleksandr-pavlyk:
**Is your feature request related to a problem? Please describe.**
Building NVBench with cmake 4.0 there cmake deprecation warnings on use of FetchContent_Populate which is coming from CPM 0.40 fetched by rapids-cmake

**Describe the solution you'd like**
rapids-cmake could upgrade to CPM.cmake/v0.42.0. Locally doing so for NVBench resolves warnings. Verified using `-DCPM_DOWNLOAD_LOCATION=CPM.cmake` downloaded using ``wget https://github.com/cpm-cmake/CPM.cmake/releases/download/v0.42.0/CPM.cmake`` 

**Describe alternatives you've considered**
The alternative is to set policy in the NVBench cmake scripts to suppress the warnings, but it is just as easy to install newer version of CMP.cmake locally.

**Additional context**
The warning message printed, with call stack implicating `rapids-cmake`:

```
-- CPM: Adding package fmt@11.2.0 (11.2.0)
CMake Warning (dev) at /home/opavlyk/miniforge/envs/nvbench/share/cmake-4.0/Modules/FetchContent.cmake:1953 (message):
  Calling FetchContent_Populate(fmt) is deprecated, call
  FetchContent_MakeAvailable(fmt) instead.  Policy CMP0169 can be set to OLD
  to allow FetchContent_Populate(fmt) to be called directly for now, but the
  ability to call it with declared details will be removed completely in a
  future version.
Call Stack (most recent call first):
  build/cmake/CPM_0.40.0.cmake:1074 (FetchContent_Populate)
  build/cmake/CPM_0.40.0.cmake:868 (cpm_fetch_package)
  build/cmake/CPM_0.40.0.cmake:306 (CPMAddPackage)
  build/_deps/rapids-cmake-src/rapids-cmake/cpm/find.cmake:195 (CPMFindPackage)
  cmake/NVBenchDependencies.cmake:11 (rapids_cpm_find)
  CMakeLists.txt:56 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.

```

vyasr:
[...]

Issue #788, opened 2025-02-28
[BUG] rapids_cpm_override in 25.04 fails when a project requires `SOURCE_SUBDIR`


robertmaynard:
**Describe the bug**
After rapids-cmake 25.04 updated to CPM 0.40.2 projects that had `nvtx3` entries started to fail to build due to `nvt3` targets no longer existing.

Before CPM 0.40.2 when a rapids-cmake override exists the following calls occur:
```
  FetchContent_Populate(
    nvtx3
    GIT_REPOSITORY ${override_repository}
    GIT_TAG ${override_tag}

  )
  FetchContent_Declare(
    nvtx3
    GIT_REPOSITORY ${repository}
    GIT_TAG ${tag}
    SOURCE_SUBDIR c
  )
  
  FetchContent_GetProperties(nvtx3)
  if(NOT nvtx3_POPULATED)
    add_subdirectory(${nvtx3_SOURCE_DIR}/${nvtx3_SOURCE_SUBDIR} ${nvtx3_BINARY_DIR})
  endif()
```
 The key parts are that CPM would itself track `nvtx3_SOURCE_SUBDIR` and call add_subdirectory based on the last value seen by it.

When we switch to CPM 0.40.2 all of that logic has been removed ( due to complying with policy 168 ) and instead FetchContent always handles the call to  `add_subdirectory`. When FetchContent calls `add_subdirectory` it only considers information provided to it in the first `FetchContent_Declare` and therefore ignores the `SOURCE_SUBDIR` value.

**Steps/Code to reproduce bug**
Have nvtx3 in a rapids-cmake override file

robertmaynard:
Given the fact that this issue is currently breaking downstream rapids-cmake users, we need to rollback to using CPM `0.40.0` as we develop a proper long term solution

robertmaynard:
## Ideas on how to solve ##

1. rapids-cmake override doesn't call FetchContent at all. But instead we defer everything to `rapids_cpm_find`. 
   The downside is that this means we won't override direct calls to CPMFindPackage/CPMAddPackage. 
2. The `SOURCE_SUBDIR` becomes a required component of `versions.json`. 
3. https://cmake.org/cmake/help/latest/command/cmake_language.html#dependency-providers and we see if we can provide our custom urls / info. More exploration is required

[...]

PR #882, opened 2025-07-25
Expand versions.json to have src_dir representation

robertmaynard:
## Description
This expands the `versions.json` to have a `src_dir` entry which allows us to update our CPM version ( included here ). To make this easier for each `rapids-cmake/cpm` package we also introduce `detail/package_info` as a replacement for `detail/package_details` which is now deprecated.

Fixes: #851 #788

## Checklist


ttnghia:
Turned out that this is not a bug. We have to manually update our `versions.json` file to get this work 👍

PR #897, opened 2025-08-13
Use family arch flags whenever possible

robertmaynard:
## Description
Use chip and family conditional flags whenever possible to make sure we have best possible perf flags

## Checklist

### DIFF
--- a/rapids-cmake/cuda/detail/detect_architectures.cmake
+++ b/rapids-cmake/cuda/detail/detect_architectures.cmake
@@ -49,7 +49,12 @@ int main(int argc, char** argv)
       char buff[32];
       cudaDeviceProp prop;
       if (cudaGetDeviceProperties(&prop, dev) != cudaSuccess) { continue; }
-      sprintf(buff, \"%d%d\", prop.major, prop.minor);
+      if(prop.major >= 9) {
+        // Enable chip specific optimizations for sm90+
+        sprintf(buff, \"%d%da\", prop.major, prop.minor);
+      } else {
+        sprintf(buff, \"%d%d\", prop.major, prop.minor);
+      }
       archs.insert(buff);
     }
   }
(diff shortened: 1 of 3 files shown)


rapids-cmake/cuda/set_architectures.cmake
```diff
@@ -56,26 +56,32 @@ Result Variables
...
-    list(REMOVE_ITEM supported_archs "100")
-  endif()
-
-  if(CMAKE_CUDA_COMPILER_ID STREQUAL "NVIDIA" AND CMAKE_CUDA_COMPILER_VERSION VERSION_GREATER_EQUAL
-                                                  13.0.0)
-    list(REMOVE_ITEM supported_archs "70")
+  if(CMAKE_CUDA_COMPILER_ID STREQUAL "NVIDIA")
+
+    if(CMAKE_CUDA_COMPILER_VERSION VERSION_GREATER_EQUAL 13.0.0)
+      set(supported_archs "75" "80" "86" "90a" "100f" "120a" "121f")
```
bdice, line 62: Does this give us 121 PTX that would be forward-compatible beyond the 120 family? I think we still need that for CUDA >=13.0.0 and ==12.9.0.

rapids-cmake/cuda/detail/detect_architectures.cmake
```diff
@@ -49,7 +49,12 @@ int main(int argc, char** argv)
       char buff[32];
       cudaDeviceProp prop;
       if (cudaGetDeviceProperties(&prop, dev) != cudaSuccess) { continue; }
-      sprintf(buff, \"%d%d\", prop.major, prop.minor);
+      if(prop.major >= 9) {
```
bdice, line 52: I did some formatting work on this recently. ;)
```suggestion
      if (prop.major >= 9) {
```

rapids-cmake/cuda/set_architectures.cmake
robertmaynard: I see a couple of other issues as well with < 12.8. Working on another round of changes

bdice (approved):
LGTM!

PR #899, opened 2025-08-14
set_architectures + NATIVE works even without a GPU

robertmaynard:
## Description
It would previously append -real to the pre-built list of architectures and we would end up with `75-real-real` and so forth.

## Checklist


vyasr (approved):
Got it, so we add `-real` only in the cases that we need in`detect_architectures.cmake` so that we don't have to append it unconditionally (and incorrectly) in `set_architectures.cmake`.

PR #919, opened 2025-10-02
Adding `ENABLE_UNSTABLE` option for `rapids_cpm_cccl`

nirandaperera:
## Description

This PR adds an  `ENABLE_UNSTABLE` option to `rapids_cpm_cccl`, which will add `cudax` headers to the cmake build. Its set to OFF by default. 

## Checklist


bdice (commented):
We’ll want to add some tests to this PR. Verify that the cudax target exists if the option is enabled, and doesn’t exist if it is disabled or not set. Maybe ask Cursor for help, there can be a fair amount of boilerplate in CMake…

nirandaperera:
@bdice could you please add labels to this PR. I dont seem to have permission to update labels.

bdice:
@robertmaynard or @KyleFromNVIDIA, would you be able to give this a review? It seems fine to me but I feel like there might be something else we need. Do we need any changes in `cmake-format-rapids-cmake.json`? https://github.com/rapidsai/rapids-cmake/blob/d81225b948f2f79d291bfc286f23c7491beeb1de/cmake-format-rapids-cmake.json#L128-L136

KyleFromNVIDIA (approved):

PR #921, opened 2025-10-07
Can't use Genex in `GLOBAL_TARGETS`

robertmaynard:
## Description
You can't use generator expression in the arguments to `GLOBAL_TARGET`, as it will get converted to something like:

```cmake

if(TARGET $<IF:$<BOOL:${_RAPIDS_ENABLE_UNSTABLE}>,CCCL::cudax,>)
   .... 
endif()

```

Which is invalid as generator expression can't be used in an `if` statement. Instead just always pass optional targets in `GLOBAL_TARGETS` as the function already does an existince check.

## Checklist


bdice (approved):
Thanks!

nirandaperera:
Thanks @robertmaynard . TIL!

PR #1002, opened 2026-04-10
Avoid rapids-logger from making BUILD_SHARED_LIBS a cache variable

vyasr:
## Description

We need to ensure that rapids-logger doesn't change `BUILD_SHARED_LIBS` for other subsequently cloned dependencies when CPM clones the library.

## Checklist


robertmaynard (changes requested):
This should just go into rapids-logger.

Check if you are the root project:
* If root use `option(BUILD_SHARED_LIBS`
* If not root use `set(BUILD_SHARED_LIBS`

bdice (approved):
~Is there any change needed in rapids-logger, or just here?~

I see Rob's comment now.

vyasr:
Replaced by https://github.com/rapidsai/rapids-logger/pull/72.

PR #1004, opened 2026-04-14
Allow proprietary_binary-only packages without git_url/git_tag

### LINKED ISSUE #702: [FEA] Relax requirements for versions.json git_url / git_tag for proprietary binaries
cuDF has not supported nvcomp 2.2.0 for a long time, but rapids-cmake's `versions.json` still has an entry for nvcomp 2.2.0 and the outdated repo page (the nvcomp repo states, "This page will be retired soon" so the `git_url` will soon be irrelevant). However, we can't remove this outdated information because the `versions.json` parser expects `git_url` and `git_tag` to exist.

Update the parser so that proprietary packages can be used with only `proprietary_binary` information, without specifying `git_url` and `git_tag`.

One alternative would be to allow an empty string like `"git_url": ""` / `"git_tag": ""`, but that also fails with ``rapids_cmake can't parse 'nvcomp' json entry, it is missing a `git_url` entry.``

**Additional context**
Related to https://github.com/rapidsai/cudf/issues/16757.

bdice:
## Summary

- Allow `versions.json` entries with `proprietary_binary` to omit `git_url`/`git_tag`/`git_shallow`, since some packages are only consumed via proprietary binary
- Remove the outdated git fields from the nvcomp `versions.json` entry
- Warn when `rapids_cpm_nvcomp` is called without `USE_PROPRIETARY_BINARY ON`

Closes #702

## Details

The nvcomp entry in `versions.json` has long carried `git_url`/`git_tag` fields pointing to the old open-source nvcomp repository. This repo does not reflect the proprietary binary that RAPIDS actually consumes, but the fields were required by the `versions.json` parser. This PR removes that requirement for packages that have a `proprietary_binary` entry.

### Parser changes

`package_details.cmake`: The validation logic already supported two fetch modes (git and url). This adds a third: when a `proprietary_binary` JSON field exists, the parser allows omitting both `git_url`/`git_tag` and `url`/`url_hash`. The return path sets empty strings for the url and tag output variables in this case.

`package_info.cmake`: The CPM content builder now handles the case where both `_rapids_url` and `_rapids_tag` are empty by producing no fetch arguments. This works because `nvcomp.cmake` downloads the proprietary binary and sets `nvcomp_ROOT` before `rapids_cpm_find` is called, so `find_package` succeeds without CPM needing to fetch anything.

### nvcomp.cmake

Added a warning when `rapids_cpm_nvcomp` is called without `USE_PROPRIETARY_BINARY ON`, since there is no longer a source checkout fallback. A local/system install of nvcomp can still be found via `find_package` without the proprietary binary flag.

### Test changes

- `cpm_nvcomp-simple.cmake`: Rewritten to use `USE_PROPRIETARY_BINARY ON`
- `cpm_nvcomp-proprietary-off.cmake`: Removed (tested source checkout fallback, no longer applicable)
- `cpm_nvcomp-invalid-arch.cmake`: Removed (tested fallback to source on unsupported arch, no longer applicable)
- `cpm_nvcomp-proprietary-on.cmake`: Updated second call to pass `USE_PROPRIETARY_BINARY ON`
- `cpm_nvcomp-override-clears-proprietary_binary.cmake`: Added explicit `git_tag` to override JSON since it can no longer inherit from the default


KyleFromNVIDIA (approved):

testing/cpm/CMakeLists.txt
```diff
@@ -119,10 +119,8 @@ add_cmake_build_test(cpm_nvbench-conda.cmake SERIAL)
 add_cmake_config_test(cpm_nvbench-explicit-static.cmake SERIAL)
 
 add_cmake_config_test(cpm_nvcomp-export.cmake)
-add_cmake_config_test(cpm_nvcomp-proprietary-off.cmake)
```
KyleFromNVIDIA, line 122: Shouldn't we still have some sort of test for what happens if we have `USE_PROPRIETARY_BINARY OFF`?

KyleFromNVIDIA (changes requested):

testing/cpm/CMakeLists.txt
bdice: Good call. I’ll add that.

bdice: Fixed in 06eadb62b843c9dfb2a283c0c1708b6054afc5aa.

KyleFromNVIDIA (approved):

robertmaynard:
@bdice Please open a follow-up PR that updates the docs with these new rules for proprietary_binary, `git_url`, and `url`.

bdice:
Thanks @robertmaynard, sorry I forgot that item. I updated the docs in #1006.

PR #1007, opened 2026-04-15
export_cpm: Better <package>_DIR heuristic in the face of ambiguity

robertmaynard:
## Description

As https://github.com/rapidsai/rapids-cmake/pull/998 and the subsequent revert highlighted the current model for recording the `<package>_DIR` hint for CPM projects was insufficient.

The newly added tests highlight that the proper heursitic is to first see if the build directory has package files ( e.g. <package>-config.cmake ) before falling back to guessing that it could be the source directory.

## Checklist

### DIFF
--- a/rapids-cmake/export/cpm.cmake
+++ b/rapids-cmake/export/cpm.cmake
@@ -49,6 +49,7 @@ function(rapids_export_cpm type name export_set)
   list(APPEND CMAKE_MESSAGE_CONTEXT "rapids.export.cpm")
 
   string(TOLOWER ${type} type)
+  string(TOLOWER ${name} lowercase_name)
 
   set(options "")
   set(one_value EXPORT_SET)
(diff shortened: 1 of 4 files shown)


robertmaynard:
https://github.com/rapidsai/cuvs/pull/2010 shows that this PR doesn't break cuvs like https://github.com/rapidsai/rapids-cmake/pull/998 did.

rapids-cmake/export/cpm.cmake
```diff
@@ -49,19 +49,50 @@ function(rapids_export_cpm type name export_set)
...
   if(type STREQUAL build)
-    if(DEFINED ${name}_DIR AND ${name}_DIR)
-      # Export out where we found the existing local config module
-      set(possible_dir "${${name}_DIR}")
+
+    # Check if we have any of the support build directory config files first
+    FetchContent_GetProperties(${name} BINARY_DIR possible_build_dirs)
+    list(APPEND possible_build_dirs "${${name}_BINARY_DIR}")
+
+    set(possible_build_config_files "${lowercase_name}-config.cmake" "/${name}Config.cmake")
```
KyleFromNVIDIA: Nitpick (doesn't impact functionality, but is confusing for the reader):

```suggestion
    set(possible_build_config_files "${lowercase_name}-config.cmake" "${name}Config.cmake")
```

rapids-cmake/export/cpm.cmake
```diff
@@ -49,19 +49,50 @@ function(rapids_export_cpm type name export_set)
...
+      if(build_dir_config_found)
+        break()
+      endif()
+    endforeach()
+
+    if(NOT build_dir_config_found)
+      # Only when we `<package>_DIR` do we want to see if we can use the FetchContent info. This
+      # maintains compatibility with projects where we need to fall-back to the build directory
+      set(possible_src_dir "${${name}_DIR}")
+      if(DEFINED ${name}_DIR AND ${name}_DIR)
```
KyleFromNVIDIA: Checking `DEFINED` and truthiness of the same variable is weird. Just do this instead:


```suggestion
      if("${${name}_DIR}")
```

KyleFromNVIDIA (changes requested):

rapids-cmake/export/cpm.cmake
robertmaynard: :+1:

KyleFromNVIDIA (approved):

vyasr (changes requested):
Can we also see a test of cudf builds to verify that we don't break those? The last edition of this change caused issues in cudf as well as cuvs (see https://nvidia.slack.com/archives/C01PF84NXCL/p1775754621180679).

robertmaynard:
> Can we also see a test of cudf builds to verify that we don't break those? The last edition of this change caused issues in cudf as well as cuvs (see https://nvidia.slack.com/archives/C01PF84NXCL/p1775754621180679).

https://github.com/rapidsai/cudf/pull/22173 is now open to test against cuDF

vyasr (approved):
Approving now that the cudf test PR is passing.

PR #1005, opened 2026-04-14
Use GitHub tarball downloads for CPM dependencies

bdice:
## Description

Convert 7 CPM packages from `git clone` to GitHub tarball archive downloads (`url` + `url_hash`) in `versions.json`, reducing download sizes significantly.

Depends on #1011 (patch-application support for tarball-mode packages).

**Converted packages:** benchmark, bs_thread_pool, cuco, rapids_logger, GTest, nvbench, nvtx3

**Intentionally excluded:**
- `rmm` — uses dynamic version placeholders (`${rapids-cmake-version}`)
- `nvcomp` — has `proprietary_binary` entries requiring special handling

### Size Comparison

| Package | Tarball Download | Tarball Extracted | Git Clone (full) | Git .git Dir | Download Savings |
|---------|-----------------|-------------------|------------------|--------------|------------------|
| benchmark | 200.0 KB | 878.4 KB | 4.6 MB | 3.7 MB | 95% |
| bs_thread_pool | 80.7 KB | 348.1 KB | 951.3 KB | 603.1 KB | 87% |
| cuco | 1.0 MB | 3.9 MB | 12.3 MB | 8.4 MB | 88% |
| rapids_logger | 30.4 KB | 109.9 KB | 319.8 KB | 209.9 KB | 86% |
| GTest | 856.9 KB | 3.9 MB | 18.5 MB | 14.7 MB | 94% |
| nvbench | 513.3 KB | 2.4 MB | 4.6 MB | 2.3 MB | 78% |
| nvtx3 | 3.8 MB | 11.9 MB | 18.4 MB | 6.6 MB | 42% |

**Total tarball download: 6.5 MB**
**Total git download (approx .git dir): 36.4 MB**
**Total savings: 82%**

Notes:
- "Git .git Dir" is the approximate network transfer size (the `.git` directory after a full clone)
- All packages previously used `git_shallow: false` (full history clones), so the savings are real
- nvtx3 has the smallest relative savings because its repo contains large binary assets in the working tree
- CCCL (already converted in #968) and rmm/nvcomp (excluded) are not shown
- SHA256 hashes were computed from the GitHub archive tarballs and spot-checked for correctness

Closes #968

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/rapids-cmake/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.
- [x] The `cmake-format.json` is up to date with these changes.

PR #1011, opened 2026-04-27
Support tarball-mode packages in patch application

bdice:
## Description

Add a new patch application path that uses `patch -p1`, so CPM packages fetched as tarballs can have patches applied to them. Tarball-extracted directories have no `.git` repository and so cannot use `git apply` or `git am`. The existing git-mode path is retained unchanged for packages fetched as git clones. The fetch mode is detected from `versions.json` (`git_url`/`git_tag` vs `url`/`url_hash`) and the patch script dispatches accordingly.

Also includes a few adjacent fixes uncovered while adding the above:

- Fix `pinning_write_file.cmake` to respect mode switching when an override provides git fields for a package whose default is URL-mode. Previously the fallback to default JSON could incorrectly detect URL-mode even when the override explicitly switched to git-mode.
- Update `verify_generated_pins` to handle URL-mode packages in pin verification, checking `url`/`url_hash` fields instead of `git_tag`/`git_shallow`.
- Update `cpm_package_override-multiple` tests to provide `git_url` alongside `git_tag` in overrides so that partial git overrides remain valid regardless of whether the default entry is git- or url-mode.

### Test coverage

Splits the existing patch-command tests into explicit `-git` and `-tarball` variants for each scenario (regular, embedded, required, required-fails) and adds `-diff-git`/`-diff-tarball` pairs so both `.patch` and `.diff` files are exercised in both modes. Each test specifies its fetch mode explicitly in `override.json` (`git_url`/`git_tag` for `-git` tests, `url`/`url_hash` for `-tarball` tests) rather than relying on the default, so the tests remain correct regardless of how `versions.json` evolves. The override is applied before `rapids_cpm_package_details_internal` so the explicitly declared mode is used for the test's pre-populate step.

Coverage matrix:

|              | git mode | tarball mode |
|--------------|----------|--------------|
| .diff        | new      | new          |
| .patch       | existing | new          |
| inline patch | existing | new          |
| required     | existing | new          |
| required-fails | existing | new        |

This PR is a prerequisite for #1005 (converting `versions.json` entries to tarballs), since several of the packages being converted have patches applied via overrides in downstream RAPIDS projects.

## Checklist


bdice:
Testing jointly with #1005 downstream in cuDF: https://github.com/rapidsai/cudf/pull/22313

rapids-cmake/cpm/detail/generate_patch_command.cmake
```diff
@@ -59,11 +59,44 @@ function(rapids_cpm_generate_patch_command package_name version patch_command bu
...
+  endif()
+  if(override_has_url AND NOT override_has_git)
+    set(use_git_patch FALSE)
+  elseif(NOT override_has_git AND NOT override_has_url)
+    string(JSON _v ERROR_VARIABLE no_git_url_default GET "${full_default_json}" git_url)
+    string(JSON _v ERROR_VARIABLE no_url_default GET "${full_default_json}" url)
+    if(no_git_url_default AND NOT no_url_default)
+      set(use_git_patch FALSE)
+    endif()
+  endif()
```
robertmaynard, line 77: Should be able to replace this with:
```
  # Determine if the package uses git or URL (tarball) mode to select the patch tool. An override
  # that specifies git fields switches the package to git mode; an override that specifies url
  # fields switches it to url mode. Otherwise fall back to the default JSON.
  set(use_git_patch TRUE)
  if(full_override_json)
    string(JSON _v ERROR_VARIABLE no_url GET "${full_override_json}" url)
    string(JSON _v ERROR_VARIABLE no_git_url GET "${full_override_json}" git_url)
    if(no_url AND no_git_url)
      string(JSON _v ERROR_VARIABLE no_url GET "${full_default_json}" git_url)
      string(JSON _v ERROR_VARIABLE no_git_url GET "${full_default_json}" url)
    endif()

    if(NOT no_url)
      set(use_git_patch FALSE)
    endif()
  endif()
```

rapids-cmake/cpm/detail/pinning_write_file.cmake
```diff
@@ -272,14 +272,35 @@ function(rapids_cpm_pinning_add_json_entry package_name json_var)
...
+    if(override_has_url AND NOT override_has_git)
+      set(is_url_mode TRUE)
+    elseif(NOT override_has_url AND NOT override_has_git)
+      # Override doesn't specify mode, fall back to default
+      string(JSON url_value ERROR_VARIABLE no_url GET "${json_data}" url)
+      string(JSON url_hash_value ERROR_VARIABLE no_url_hash GET "${json_data}" url_hash)
+      if(NOT no_url AND NOT no_url_hash)
+        set(is_url_mode TRUE)
+      endif()
+    endif()
```
robertmaynard: You can simplify this like the above by assuming git and having a single state variable that represents git or url

robertmaynard (changes requested):

robertmaynard (approved):

PR #1014, opened 2026-04-29
Add Jetson Thor (sm_110) to architecture preset

Kaweees:
This pull request adds the Jetson Thor (`sm_110`) to the architecture preset in `rapids_cuda_set_architectures`. 

Before this change, projects consuming the previous preset emitted no SASS for Thor and failed at runtime on Jetson Thor with `cudaErrorInvalidDevice` on the first kernel launch.

## Checklist


robertmaynard:
RAPIDS doesn't support Thor so we won't be adding `sm_110` to the default set of architectures.
If your project needs to support Thor you will need to pass a custom set of CUDA architectures in via the supported methods ( `-DCMAKE_CUDA_ARCHITECTURES` or `CUDAARCHS` env var )

PR #1026, opened 2026-05-21
Add `rapids_json_compute_matrix_product()` function

KyleFromNVIDIA:
## Description
This function computes a matrix product based on a JSON document. Ported from
https://github.com/rapidsai/cuvs/blob/01aab633dbfdff641e4258ce432477b32a8acd55/cpp/cmake/modules/compute_matrix_product.py.

## Checklist


rapids-cmake/cmake/compute_json_matrix_product.cmake
```diff
@@ -0,0 +1,439 @@
...
+    _rapids_cmake_cjmp_leaf(result PATH "${_RAPIDS_PATH}" KEY "${_RAPIDS_KEY}" MATRIX
+                            "${_RAPIDS_MATRIX}" QUEUE "${_RAPIDS_QUEUE}" ENTRY "${_RAPIDS_ENTRY}"
+                            WARN_USED "${_RAPIDS_WARN_USED}" WARN_UNUSED "${_RAPIDS_WARN_UNUSED}")
+  endif()
+
+  set(${out_var} "${result}" PARENT_SCOPE)
+endfunction()
+
+#[=======================================================================[.rst:
+rapids_cmake_compute_json_matrix_product
```
KyleFromNVIDIA: I might add other JSON-related functionality in the future (this file alone has a few helper functions that we could split out into dedicated files), and I'm starting to wonder if we should create a `rapids_json` group for that purpose. Then we could call this `rapids_json_compute_matrix_product()`.

rapids-cmake/json/array_extend.cmake
```diff
@@ -0,0 +1,45 @@
...
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.
+# SPDX-License-Identifier: Apache-2.0
+# cmake-format: on
+# =============================================================================
+include_guard(GLOBAL)
+
+include("${rapids-cmake-dir}/json/array_append.cmake")
+
+# TODO: Remove this version gate once we require 4.3
+cmake_minimum_required(VERSION 4.3)
```
robertmaynard: This should be inside the function and setup to only trigger when somebody calls the function. We shouldn't error out on inclusion only

testing/json/array_append.cmake
```diff
@@ -0,0 +1,25 @@
...
+    string(REPLACE "\n" "\n  " formatted_expected_result "${expected_result}")
+    string(REPLACE "\n" "\n  " formatted_actual_result "${actual_result}")
+    message(SEND_ERROR "Expected result:\n  ${formatted_expected_result}\nActual result:\n  ${formatted_actual_result}"
+    )
+  endif()
+endfunction()
+
+test_array_append("[1]" "[]" "1")
+test_array_append("[0, 1]" "[0]" "1")
+test_array_append("[0, 1, 2]" "[0, 1]" "2")
```
robertmaynard, line 22: Please add a test where we are adding an array as an element `test_array_append( [0, [1,2]], [0], [1,2])`

testing/json/array_extend.cmake
```diff
@@ -0,0 +1,26 @@
...
+  string(JSON equal EQUAL "${expected_result}" "${actual_result}")
+  if(NOT equal)
+    string(REPLACE "\n" "\n  " formatted_expected_result "${expected_result}")
+    string(REPLACE "\n" "\n  " formatted_actual_result "${actual_result}")
+    message(SEND_ERROR "Expected result:\n  ${formatted_expected_result}\nActual result:\n  ${formatted_actual_result}"
+    )
+  endif()
+endfunction()
+
+test_array_extend("[1]" "[]" "[1]")
```
robertmaynard, line 20: Please add a test case where what we are extending by is not an array

rapids-cmake/json/compute_matrix_product.cmake
```diff
@@ -0,0 +1,422 @@
...
+# SPDX-License-Identifier: Apache-2.0
+# cmake-format: on
+# =============================================================================
+include_guard(GLOBAL)
+
+include("${rapids-cmake-dir}/json/array_append.cmake")
+include("${rapids-cmake-dir}/json/array_extend.cmake")
+
+# TODO: Remove this version gate once we require 4.3
+cmake_minimum_required(VERSION 4.3)
```
robertmaynard: This should be inside the function and setup to only trigger when somebody calls the function. We shouldn't error out on inclusion only

rapids-cmake/json/compute_matrix_product.cmake
```diff
@@ -0,0 +1,422 @@
+# =============================================================================
+# cmake-format: off
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.
+# SPDX-License-Identifier: Apache-2.0
+# cmake-format: on
+# =============================================================================
+include_guard(GLOBAL)
+
+include("${rapids-cmake-dir}/json/array_append.cmake")
```
robertmaynard: include should be in the function body

rapids-cmake/json/compute_matrix_product.cmake
```diff
@@ -0,0 +1,422 @@
...
+include("${rapids-cmake-dir}/json/array_extend.cmake")
+
+# TODO: Remove this version gate once we require 4.3
+cmake_minimum_required(VERSION 4.3)
+
+# rapids-lint: disable=C0111
+function(_rapids_json_cmp_queue_item out_var path key matrix)
```
robertmaynard: Let' move all the helper functions to a separate file under `detail`

robertmaynard (changes requested):

rapids-cmake/cmake/compute_json_matrix_product.cmake
vyasr: @KyleFromNVIDIA let's not overengineer what we put into rapids-cmake here. I think the long-term goal should be for the entire JIT compilation framework we're building to be centralized, and it's much more than just CMake. At that point I think we'll want this functionality to live there rather than in rapids-cmake.

KyleFromNVIDIA: This function isn't strictly for JIT+LTO though, it can be used in other contexts. See https://github.com/rapidsai/cuvs/blob/e7a6f597abe8a5d9bf819c51d0489d799048dbc2/cpp/cmake/modules/generate_inst_matrix.cmake#L31-L67 and https://github.com/rapidsai/cuvs/blob/e7a6f597abe8a5d9bf819c51d0489d799048dbc2/cpp/CMakeLists.txt#L228-L234 for an example - cuvs generates a number of nearly-identical files that instantiate templates with different template parameters, and this JSON matrix product algorithm is what enables that. This algorithm is generic enough that I think it makes sense to include it in rapids-cmake independently of JIT+LTO.

vyasr: OK that's fair, in that case I withdraw my objection. It looks like you've also already created the JSON namespace, which is fine with me.

rapids-cmake/json/detail/compute_matrix_product_impl.cmake
```diff
@@ -0,0 +1,252 @@
...
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.
+# SPDX-License-Identifier: Apache-2.0
+# cmake-format: on
+# =============================================================================
+include_guard(GLOBAL)
+
+include("${rapids-cmake-dir}/json/array_append.cmake")
+include("${rapids-cmake-dir}/json/array_extend.cmake")
+
+# rapids-lint: disable=C0111
```
vyasr: I assume this is the warning that says you need a docstring. You're disabling it a lot in this file. Maybe easier to put in actual one-line descriptions, even for the internal functions? It doesn't have to be complete enough as something that would go into published docs, just to help devs.

vyasr (approved):
The code looks good. I mostly skimmed the tests, they look good enough to me but I saw that Rob already looked at those once so I'll trust his follow-up review there.

rapids-cmake/json/detail/compute_matrix_product_impl.cmake
KyleFromNVIDIA: Added some simple description comments

robertmaynard (approved):

Issue #582, opened 2024-04-11
[FEA] Provide ability to setup default install name components per sub-project 


robertmaynard:
**Is your feature request related to a problem? Please describe.**
Most RAPIDS projects bring in thirdparty projects via `rapids_cpm` and either have them as part of the default `ALL` install set, or mark them `EXCLUDE_FROM_ALL` and therefore become uninstallable unless they have existing `install(COMPONENT` rules setup.

rapids-cmake can resolve these issues by marking each sub-project with default component names ( based on the project name ), and controls to overridable this default name when needed ( e.g. when we want all testing libraries to go into the `testing` component ).

**Describe the solution you'd like**

The solution is rather ingenious and [proposed by Craig Scott](https://gitlab.kitware.com/cmake/cmake/-/issues/25877#note_1507740).

What we have is a user function with the following call signature:
```
rapids_cmake_default_install_component(DEFAULT_USE_PROJECT_NAME)
rapids_cmake_default_install_component(PROJECT <name> INSTALL_COMPONENT_NAME <install_name>)
```
Which would have roughly the following implementation:
```cmake

function(rapids_cmake_default_install_component )
  set(options DEFAULT_USE_PROJECT_NAME)
  set(one_value PROJECT INSTALL_COMPONENT_NAME)
  set(multi_value )
  cmake_parse_arguments(_RAPIDS "${options}" "${one_value}" "${multi_value}" ${ARGN})

  # Need to make sure we only install our project hook once
  get_property(already_hooked GLOBAL PROPERTY rapids_cmake_default_install_component_hook)
  if(NOT already_hooked)
    list(APPEND CMAKE_PROJECT_INCLUDE "${CMAKE_CURRENT_LIST_DIR}/default_install_component_hook.cmake")
    set(CMAKE_PROJECT_INCLUDE "${CMAKE_PROJECT_INCLUDE}" PARENT_SCOPE)

    set_property(GLOBAL PROPERTY rapids_cmake_default_install_component_hook ON)
  endif()

  if(_RAPIDS_DEFAULT_USE_PROJECT_NAME)
    set_property(GLOBAL PROPERTY rapids_cmake_default_install_component_names ON)
  endif()
  if(_RAPIDS_PROJECT AND _RAPIDS_INSTALL_COMPONENT)
    # Should only be set if the property doesn't already exist
    string(TOLOWER ${_RAPIDS_PROJECT} project)
[...]

PR #1032, opened 2026-05-26
Support configuring default install components for subprojects

arhag23:
## Description
Implementation for a new `rapids_cmake_default_install_component` function that supports either enabling using the project name as the default install component name or a custom value for a specific project.

Addresses the feature request #582.

Disabled precommit check for hardcoded version, but tested and pass with the new precommit hook which widens the doc heuristics.

## Checklist


testing/cmake/default_install_component/default_dependency/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+# =============================================================================
+# cmake-format: off
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+# cmake-format: on
+# =============================================================================
+cmake_minimum_required(VERSION 3.30.4)
```
KyleFromNVIDIA: We now require 4.0:

```suggestion
cmake_minimum_required(VERSION 4.0)
```

testing/cmake/default_install_component/testing_dependency/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+# =============================================================================
+# cmake-format: off
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+# cmake-format: on
+# =============================================================================
+cmake_minimum_required(VERSION 3.30.4)
```
KyleFromNVIDIA:
```suggestion
cmake_minimum_required(VERSION 4.0)
```

testing/cmake/default_install_component/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
+# =============================================================================
+# cmake-format: off
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+# cmake-format: on
+# =============================================================================
+cmake_minimum_required(VERSION 3.30.4)
```
KyleFromNVIDIA:
```suggestion
cmake_minimum_required(VERSION 4.0)
```

rapids-cmake/cmake/default_install_component.cmake
```diff
@@ -0,0 +1,62 @@
...
+#]=======================================================================]
+function(rapids_cmake_default_install_component)
+  list(APPEND CMAKE_MESSAGE_CONTEXT "rapids.cmake.default_install_component")
+  set(options DEFAULT_USE_PROJECT_NAME)
+  set(one_value PROJECT INSTALL_COMPONENT_NAME)
+  set(multi_value)
+  cmake_parse_arguments(_RAPIDS "${options}" "${one_value}" "${multi_value}" ${ARGN})
+
+  # Need to make sure we only install our project hook once
+  get_property(already_hooked GLOBAL PROPERTY rapids_cmake_default_install_component_hook)
```
KyleFromNVIDIA: This property is `GLOBAL`, but `CMAKE_PROJECT_INCLUDE` gets set at the directory level. This means that if one subdirectory calls this function, subsequent calls in other subdirectories will have no effect, and not work correctly. Consider:

`CMakeLists.txt`:
```cmake
add_subdirectory(first)
add_subdirectory(second)
```

`first/CMakeLists.txt`:
```cmake
rapids_cmake_default_install_component(DEFAULT_USE_PROJECT_NAME)

# ...
```

`second/CMakeLists.txt`:
```cmake
rapids_cmake_default_install_component(DEFAULT_USE_PROJECT_NAME)

# ...
```

The hook will not execute in `second/CMakeLists.txt`, since the `GLOBAL`property was found to be `ON`, but `CMAKE_PROJECT_INCLUDE` was not set due to being confined to `first`'s directory scope.

KyleFromNVIDIA (changes requested):

rapids-cmake/cmake/detail/default_install_component_hook.cmake
```diff
@@ -0,0 +1,32 @@
...
+# rapids-pre-commit-hooks: disable[verify-hardcoded-version]
+#[=======================================================================[.rst:
+rapids_cmake_default_install_component_hook
+---------------------------------------
+
+.. versionadded:: v26.08.00
+
+Hook used by :cmake:command:`rapids_cmake_default_install_component` to set the default install component name for a project.
+
+#]=======================================================================]
```
KyleFromNVIDIA: There's no need for a docstring here, since this is an internal implementation detail and there's no function to document.


```suggestion
```

rapids-cmake/cmake/default_install_component.cmake
```diff
@@ -0,0 +1,62 @@
...
+``DEFAULT_USE_PROJECT_NAME``
+  Enables the use of project name as the default install component name.
+
+``PROJECT``
+  The project name to set a custom install component name for.
+
+``INSTALL_COMPONENT_NAME``
+  The install component name to set for a given project.
+
+#]=======================================================================]
```
KyleFromNVIDIA, line 35: Please re-enable the hook after the docstring:

```suggestion
#]=======================================================================]
# rapids-pre-commit-hooks: enable[verify-hardcoded-version]
```

KyleFromNVIDIA (changes requested):

rapids-cmake/cmake/default_install_component.cmake
KyleFromNVIDIA: I'm not sure what to do about this. Still thinking about it.

robertmaynard: I think we need todo is instead of using a global property hook, we check if ` "${CMAKE_CURRENT_FUNCTION_LIST_DIR}/detail/default_install_component_hook.cmake"` is in `CMAKE_PROJECT_INCLUDE` and only add it if not already there.

That will handle the sibling and nested use cases

testing/cmake/CMakeLists.txt
```diff
@@ -21,6 +21,8 @@ add_cmake_config_test(conda_env-invalid.cmake)
 add_cmake_config_test(conda_env-prefix.cmake)
 add_cmake_config_test(conda_env-prefix-envvar.cmake)
 
+add_cmake_build_test(default_install_component)
```
robertmaynard: We should add a test for the sibling example that @KyleFromNVIDIA brought up in review

robertmaynard (changes requested):

rapids-cmake/cmake/detail/default_install_component_hook.cmake
```diff
@@ -0,0 +1,22 @@
...
+
+get_property(_rapids_install_default_name GLOBAL
+             PROPERTY rapids_cmake_default_install_component_names)
+get_property(_rapids_install_project_name GLOBAL PROPERTY rapids_cmake_install_component_${project})
+
+if(_rapids_install_project_name)
+  set(CMAKE_INSTALL_DEFAULT_COMPONENT_NAME "${_rapids_install_project_name}")
+elseif(_rapids_install_default_name)
+  if(CMAKE_VERSION VERSION_GREATER_EQUAL 4.4)
+    set(CMAKE_INSTALL_DEFAULT_COMPONENT_NAME "<PROJECT_NAME>")
```
robertmaynard: I think we should always use the lower case name for this. Therefore we can drop the 4.4 branch and add it back once we have extended cmake with a `<PROJECT_LOWER_CASE_NAME>` placeholder

rapids-cmake/cmake/detail/default_install_component_hook.cmake
```diff
@@ -0,0 +1,22 @@
...
+  else()
+    set(CMAKE_INSTALL_DEFAULT_COMPONENT_NAME "${PROJECT_NAME}")
```
robertmaynard:
```suggestion
    set(CMAKE_INSTALL_DEFAULT_COMPONENT_NAME "${project}")
```

robertmaynard (changes requested):

rapids-cmake/cmake/default_install_component.cmake
arhag23: Is it then okay that calling `rapids_cmake_default_install_component(DEFAULT_USE_PROJECT_NAME)` anywhere will enable this behavior for sibling subprojects even if only a single project is given an explicit `INSTALL_COMPONENT_NAME`.

For the above example, if we make `second/CMakeLists.txt` have
```cmake
rapids_cmake_default_install_component(PROJECT dependency INSTALL_COMPONENT_NAME testing)

# ... (including subprojects other than dependency)
```
then those projects would also have the hook and since the global property for using the default project name is set in the sibling directory, any subprojects would have the `DEFAULT_USE_PROJECT_NAME` behavior. This may not be transparent to users.

robertmaynard: > Is it then okay that calling rapids_cmake_default_install_component(DEFAULT_USE_PROJECT_NAME) anywhere will enable this behavior for sibling subprojects 

Sibling suprojects shouldn't see any behavior change since the `CMAKE_PROJECT_INCLUDE` for them won't have the file, since that variable change is scoped to just the directory ( and subdirs ) that called `rapids_cmake_default_install_component`.

I recommend adding another sibling test that validates this statement

KyleFromNVIDIA (approved):

robertmaynard (approved):

PR #1038, opened 2026-06-03
Fixes to update version for `release/26.06`

jolorunyomi:
## Description
Fixes update versions script issue for `release/26.06`


samson-nvidia (approved):

robertmaynard (changes requested):
The whole sed should be removed as the entire example is gone and has been replaced by a folder of examples non of which use the same version value as rapids-cmake release

PR #1042, opened 2026-06-12
Add `rapids_cmake_check_conda_env()` function

arhag23:
## Description
When the conda environment is changed, CMake requires you to wipe the build directory and reconfigure. This may not be transparent to those unfamiliar with CMake, so this function provides a helper to error in such situations and show the command that solves this problem.

Implementation sourced from https://github.com/rapidsai/cuml/commit/62ef29e7ff1be122289cac90f03daa592cc2d749

## Checklist

### DIFF
--- a/rapids-cmake/cmake/check_conda_env.cmake
+++ b/rapids-cmake/cmake/check_conda_env.cmake
@@ -28,11 +28,6 @@ differs from the recorded value, configure terminates with a fatal error that in
   Instructions shown to the user on how to delete the build directory and reconfigure. If not provided, a
   default instruction ``rm -rf <build directory>`` is used.
 
-Result Variables
-^^^^^^^^^^^^^^^^
-  :cmake:variable:`rapids_conda_prefix` will be set to ``CONDA_PREFIX`` when
-  initially configured.
-
 #]=======================================================================]
 # rapids-pre-commit-hooks: enable[verify-hardcoded-version]
 function(rapids_cmake_check_conda_env)
--- a/rapids-cmake/cmake/check_conda_env.cmake
+++ b/rapids-cmake/cmake/check_conda_env.cmake
@@ -1,0 +1,53 @@
+# =============================================================================
+# cmake-format: off
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+# cmake-format: on
+# =============================================================================
+include_guard(GLOBAL)
+
+# rapids-pre-commit-hooks: disable[verify-hardcoded-version]
+#[=======================================================================[.rst:
+rapids_cmake_check_conda_env
+----------------------------
+
+.. versionadded:: v26.08.00
+
+Detect when the active Conda environment has changed since the build directory was initially configured.
+
+  .. code-block:: cmake
+
+    rapids_cmake_check_conda_env(rebuild_instruction)
+
+When called from within a Conda environment, records the current ``CONDA_PREFIX`` in the
+internal cache variable ``rapids_conda_prefix``. On subsequent configures, if ``CONDA_PREFIX``
+differs from the recorded value, configure terminates with a fatal error that includes
+``rebuild_instruction`` for the user.
+
+``rebuild_instruction``
+  Instructions shown to the user on how to delete the build directory and reconfigure.
+
+Result Variables
+^^^^^^^^^^^^^^^^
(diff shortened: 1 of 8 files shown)


rapids-cmake/cmake/check_conda_env.cmake
```diff
@@ -0,0 +1,53 @@
...
+rapids_cmake_check_conda_env
+----------------------------
+
+.. versionadded:: v26.08.00
+
+Detect when the active Conda environment has changed since the build directory was initially configured.
+
+  .. code-block:: cmake
+
+    rapids_cmake_check_conda_env(rebuild_instruction)
```
KyleFromNVIDIA: I think we should make `rebuild_instruction` optional and provide a default instruction message:

```suggestion
    rapids_cmake_check_conda_env([REBUILD_INSTRUCTION rebuild_instruction])
```

And the default rebuild instruction can be something like:

```
  rm -rf ${CMAKE_BINARY_DIR}
```

KyleFromNVIDIA (changes requested):
Requesting one small change in the API surface.

I also wonder if there's a way we can automatically call this as part of the rapids-cmake initialization process rather than requiring projects to opt into it. It seems like something that would be applicable across all of RAPIDS, not just cuml.

rapids-cmake/cmake/check_conda_env.cmake
KyleFromNVIDIA: ~~On second thought, I'm really questioning whether we even need this argument at all. What would be the use case for having it be anything other than the default I suggested above?~~ I just read the Slack thread you sent me and agree that this option should be kept. But we can still give it a sensible default.

arhag23:
> I also wonder if there's a way we can automatically call this as part of the rapids-cmake initialization process rather than requiring projects to opt into it. It seems like something that would be applicable across all of RAPIDS, not just cuml.

I think in the future this logic could be incorporated into `support_conda_env`.

testing/cmake/check_conda_env-mismatch.cmake
```diff
@@ -0,0 +1,13 @@
...
+# SPDX-License-Identifier: Apache-2.0
+# cmake-format: on
+# =============================================================================
+include(${rapids-cmake-dir}/cmake/check_conda_env.cmake)
+
+# Fails configure when CONDA_PREFIX changes
+set(ENV{CONDA_PREFIX} "/opt/conda/prefix-a")
+rapids_cmake_check_conda_env()
+set(ENV{CONDA_PREFIX} "/opt/conda/prefix-b")
+rapids_cmake_check_conda_env(REBUILD_INSTRUCTION foo)
```
KyleFromNVIDIA: Let's keep the default instruction in this one and add a `check_conda_env-mismatch-rebuild-instruction` test case

KyleFromNVIDIA (changes requested):

KyleFromNVIDIA (approved):

rapids-cmake/cmake/check_conda_env.cmake
```diff
@@ -0,0 +1,63 @@
...
+``REBUILD_INSTRUCTION`` for the user.
+
+``REBUILD_INSTRUCTION``
+  Instructions shown to the user on how to delete the build directory and reconfigure. If not provided, a
+  default instruction ``rm -rf <build directory>`` is used.
+
+Result Variables
+^^^^^^^^^^^^^^^^
+  :cmake:variable:`rapids_conda_prefix` will be set to ``CONDA_PREFIX`` when
+  initially configured.
```
KyleFromNVIDIA: This variable is an internal implementation detail, and I question the wisdom of documenting it. Any thoughts?

arhag23: Good point, I should have removed it. In a previous iteration, I had the user pass in the name of the flag they wanted to keep track of whether the configuration had occurred or not so the documentation was for that.

KyleFromNVIDIA (approved):

robertmaynard:
> Requesting one small change in the API surface.
> 
> I also wonder if there's a way we can automatically call this as part of the rapids-cmake initialization process rather than requiring projects to opt into it. It seems like something that would be applicable across all of RAPIDS, not just cuml.

rapids-cmake wants as much behavior to be opt-in instead of implicit

A longer term goal for rapids-cmake is to have something akin to epoch's which when enabled would enable a  is the ability to enable raft of features/behavior.

While not fully implicit it would reduce the amount of boiler-plate code

robertmaynard (approved):

PR #1050, opened 2026-07-07
Fix how tests with arguments are relocated in install tree

KyleFromNVIDIA:
## Description
Tests with arguments were not properly relocated due to the usage of the intermediate `run_gpu_test.cmake` script. Adjust the relocation logic to relocate substrings within an argument instead of the entire argument.

## Checklist


rapids-cmake/test/detail/generate_installed_CTestTestfile.cmake
```diff
@@ -34,23 +34,50 @@ in the `CTestTestfile` when we include it.
...
+    set(earliest_install_value)
+
+    foreach(file IN LISTS installed_files)
+      get_property(file_build_values GLOBAL PROPERTY "${file}_build")
+      get_property(file_install_value GLOBAL PROPERTY "${file}_install")
+
+      foreach(file_build_value IN LISTS file_build_values)
+        string(FIND "${build_value}" "${file_build_value}" index)
+        if(index GREATER_EQUAL 0 AND (NOT DEFINED earliest_index OR index LESS earliest_index))
+          message(STATUS "Found ${file_build_value} -> ${file_install_value}/${file}")
```
robertmaynard: Debug message

robertmaynard (approved):

testing/test/install_relocatable-installed-tests-run/main_args.cpp
bdice, line 1: Must this be a `.cu` file? Can it be `.cpp` and compile with host-only? Same for `callback.cu`.

KyleFromNVIDIA: The `verify_` target that was already here also used a `.cu` file, so I just followed the existing pattern.

bdice: We strive to use `.cu` only for files containing CUDA code. These seem like pure host code, so should be `.cpp`.

bdice (approved):

PR #1052, opened 2026-07-10
Fix errors introduced by CMake 4.4

KyleFromNVIDIA:
## Description
* CMake 4.4 changed `-Wno-dev` to `-Wno-author`. Update expected warning messages accordingly.
* Rather than `include(CTestTestfile.cmake)`, parse the output of `ctest --show-only=json-v1` to get the list of tests and relocate them.

## Checklist


robertmaynard (approved):

robertmaynard:
We can either admin merge this, or just not run these tests when CMake is >= 4.4

KyleFromNVIDIA:
I'm working on a fix for the remaining tests, I'll include it in this PR

robertmaynard (approved):
Approving.

One of the previous limitations of the test install logic was lack of support for multi-config generators. @KyleFromNVIDIA do you think by using the json test output we also fixed that issue?

KyleFromNVIDIA:
I think it should have worked before. We were setting `CTEST_CONFIGURATION_TYPE` to `${CMAKE_INSTALL_CONFIG_NAME}` before including `CTestTestfile.cmake`, so our `add_test()` mocks should have been called with different files depending on the configuration. If it didn't work, it was a bug in the implementation rather than a problem with the approach itself.

In any case, it should work now.

PR #1069, opened 2026-08-04
Update to rapids-logger 0.3

bdice:
Contributes to https://github.com/rapidsai/build-planning/issues/313.

Updates rapids-logger to version 0.3.

### DIFF
--- a/examples/pin-dependencies/pins.json
+++ b/examples/pin-dependencies/pins.json
@@ -18,9 +18,9 @@
     "rapids_logger": {
       "always_download": true,
       "git_shallow": false,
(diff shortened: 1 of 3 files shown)


robertmaynard (approved):
Can you make a issue that says we should update the GCC provided in our test images to be 13.3+

vyasr (approved):

bdice:
@robertmaynard We already provide GCC >=13.3 in all of our CI images. The test job for CUDA 12.2 explicitly installed GCC < 13 -- and this is the only job in RAPIDS that uses a GCC older than 13.3. We require GCC >=13.3 in cuDF and every other project, for `<format>` support, so this is safe to skip.

PR #1074, opened 2026-08-14
Add binutils as an explicit dependency

arhag23:
The write_language tests require binutils with `ar` available to work correctly. Newer versions of the `c-compiler` conda package no longer require `binutils` so `ar` is not available which causes these tests to fail.

## Checklist

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -35,11 +35,11 @@ dependencies:
           - ninja
       - output_types: conda
         packages:
-          - c-compiler
-          - cxx-compiler
-          - cuda-nvcc
-          - make
           - binutils
+          - c-compiler
+          - cuda-nvcc
+          - cxx-compiler
+          - make
     specific:
       - output_types: conda
         matrices:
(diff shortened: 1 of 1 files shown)


dependencies.yaml
```diff
@@ -39,6 +39,7 @@ dependencies:
           - cxx-compiler
           - cuda-nvcc
           - make
+          - binutils
```
jameslamb: Minor comment: could you move this up to keep this list alphabetized, please? Makes reading these a bit easier, especially as they get longer.

(I guess it already technically isn't because `cxx-compiler` is before `cuda-nvcc`, would support you swapping those too).

jameslamb (approved):
Just one extremely minor comment, could you please address it if you end up pushing another commit anyway?

otherwise makes sense, thanks for looking into this

bdice (approved):

bdice (approved):
RMM fix has been merged. I will rerun this and trigger a merge.

PR #1076, opened 2026-08-18
enforce 'rstcheck' checks

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/262

Proposes introducing `rstcheck` ([rstcheck/rstcheck](https://github.com/rstcheck/rstcheck)), a linter for ReStructuredText.

In my experience, it runs pretty quickly and catches things that `sphinx-build -W` does not, like:

* invalid syntax in `.. code-block::` blocks
* duplicated references (which can lead to unexpected behavior when sharing URLs with anchors)

## Checklist

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1,7 +1,7 @@
 API Reference
 #############
 
-.. _`common`:
+.. _`common-functions`:
 
 Common
 ******
@@ -23,7 +23,7 @@ require.
    /command/rapids_cmake_write_git_revision_file
    /command/rapids_cmake_write_version_file
 
-.. _`cpm`:
+.. _`cpm-functions`:
 
 CPM
 ***
@@ -53,7 +53,7 @@ package uses :ref:`can be found here <cpm_versions>`.
 
 .. include:: /packages/packages.rst
 
-.. _`cython`:
+.. _`cython-functions`:
 
 Cython
 ******
@@ -72,7 +72,7 @@ The ``rapids-cython-core`` module allows projects to easily build cython modules
    /command/rapids_cython_core_create_modules
    /command/rapids_cython_core_add_rpath_entries
 
-.. _`find`:
+.. _`find-functions`:
 
 Find
 ****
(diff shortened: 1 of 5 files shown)


docs/cpm.rst
```diff
@@ -71,16 +71,16 @@ used instead of ALL overrides specified in that project. This would allow you to
...
 
 To request the generation of a pinned package override file without having to modify
 the project use the :cmake:variable:`RAPIDS_CMAKE_CPM_PINNED_VERSIONS_FILE` variable:
 
-```
-cmake -DRAPIDS_CMAKE_CPM_PINNED_VERSIONS_FILE
-```
+.. code-block:: bash
+
+    cmake -DRAPIDS_CMAKE_CPM_PINNED_VERSIONS_FILE
```
jameslamb, line 83: `rstcheck` caught this. Using markdown-style triple backticks in an RST file doesn't created a code block.

At https://docs.rapids.ai/api/rapids-cmake/stable/cpm/#additional-cpm-command-line-controls, it looks like this right now:

Image

This change creates code blocks formatted the way you'd expect, similar to this from a few lines down.

Image

docs/api.rst
```diff
@@ -53,7 +53,7 @@ package uses :ref:`can be found here <cpm_versions>`.
 
 .. include:: /packages/packages.rst
 
-.. _`cython`:
+.. _`cython-functions`:
```
jameslamb, line 56: `rstcheck` complains because this target and the heading below both resolve as `#cython` in URLs, which is ambiguous.

Doesn't matter **that much** in practice (especially since they're right next to each other) but duplicate targets could matter in other cases (like where you accidentally forgot to rename a section much further down the page).

I think it's worth renaming these links to resolve this and keep that check running.

docs/cpm.rst
jameslamb: Yeah great, this is looking better in the PR preview: https://downloads.rapids.ai/ci/rapids-cmake/pull-request/1076/faf411f/docs/rapids-cmake/html/cpm/

vyasr (approved):
Changes seem to either be net neutral or minor improvements, so I'm fine adding this as long as the hook is cheap.

jameslamb:
Thank you. Yep the hook is pretty cheap to run in my experience. I'll merge this.

PR #1080, opened 2026-08-20
Correct CI failures building examples 

robertmaynard:
## Description
We are seeing failures in other PR's (https://github.com/rapidsai/rapids-cmake/pull/1079) due to CCCL deprecation warning becoming errors. 

This exposed the fact that our PIN example wasn't being invoked using the proper CMake preset

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -87,6 +87,7 @@ repos:
         exclude: |
           (?x)
           ^[.]github/workflows/|
+          ^examples/pin-dependencies/pins.json|
           (^|/)RAPIDS_BRANCH$|
           [.](md|rst)$
   - repo: https://github.com/shellcheck-py/shellcheck-py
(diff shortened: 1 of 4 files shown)


jameslamb (approved):

bdice (commented):
RMM has already fixed this upstream. Pulling a new RMM would also work here.

I am concerned that ignoring this warning is not really effective; updating RMM is probably less work. It also doesn't prevent future issues, because the ignore is so specific.

robertmaynard:
> RMM has already fixed this upstream. Pulling a new RMM would also work here.
> 
> I am concerned that ignoring this warning is not really effective; updating RMM is probably less work. It also doesn't prevent future issues, because the ignore is so specific.

Awesome that will allow me to drop the changes in examples/offline_build/use_offline_source_cache.sh. The PINS changes are all required as the testing wasn't actually testing the correct thing and was building using latest rmm/cccl instead of using the actual pin file.

bdice (approved):

PR #1089, opened 2026-09-15
Map download only to source subdir

### LINKED ISSUE #1088: [BUG] DOWNLOAD_ONLY seems to ignore overrides
## Context

* @jameslamb and I agreed to store internal URLs for rapids-cmake in branches on the internal fork, like https://gitlab-master.nvidia.com/RAPIDS/rapids-cmake/-/tree/release/26.10-internal?ref_type=heads
* In trying to build 26.10 standalone packages, I was hitting a lot of cases where those supposed overrides were not working, and LLM was falling back to trying to patch cudf/cuvs source. I put up https://github.com/NVIDIA/cudf/pull/24148 and https://github.com/NVIDIA/cuvs/pull/2601 as ways to make things more overridable.
* @robertmaynard advised me that those PRs should not be necessary. Instead, if we specified override behavior for rapids-cmake/CPM, it should be done at a low enough level that the PRs were superfluous.
* I closed those PRs and asked the LLM to move the content from https://gitlab-master.nvidia.com/RAPIDS/rapids-cmake/-/tree/release/26.10-internal?ref_type=heads into an override file. It didn't work.

## Current issues

I was confused on some standalone builds. One job out of 4 failed: https://prod.blsm.nvidia.com/sw-rapids-cuvs-static/blue/organizations/jenkins/standalone-builds%2Frelease-26.10-bundle/detail/release-26.10-bundle/26/pipeline/364, but the other other 3 builds for that project passed. I asked Codex, and here's what it came up with:

The failure was not lane-specific, and the remap failed in all four cuDF jobs.
The difference was GitHub retry luck:
- amd64 / CUDA 12.9: failed once, then cloned successfully.
- amd64 / CUDA 13.4: failed twice, then cloned successfully.
- arm64 / CUDA 12.9: cloned successfully immediately.
- arm64 / CUDA 13.4: failed all three attempts and aborted.
The underlying gap is DLPack’s DOWNLOAD_ONLY TRUE declaration. With CMake 4.2 and our vendored CPM 0.42.0, that path calls FetchContent_Populate(... ${ARGN}) with cuDF’s original GitHub arguments. It bypasses the source details pre-registered by the RAPIDS override. Other dependencies use the normal FetchContent path and are remapped correctly.
So Robert’s description is generally correct, but our earlier verification missed this specific CPM + DOWNLOAD_ONLY + CMake >= 3.30.3 path. The successful jobs merely concealed it through intermittent GitHub access.
We need to fix that deterministic gap, likely in the vendored CPM behavior or with an explicit DLPack CPM declaration. Simply rerunning could pass, but it would remain unreliable.

## followup

Me: Would a different cmake or CPM version make a difference?

### Codex answered

Yes, but only a CMake downgrade would change this behavior today.
- CMake 3.30.2 or older: CPM uses FetchContent_Populate(name), which consumes the previously registered override. This should make the DLPack remap work.
- CMake 3.30.3 through our current 4.2.3: CPM’s DOWNLOAD_ONLY path uses the full-arguments form, which ignores saved FetchContent details and therefore restores cuDF’s GitHub URL.
- Newer CPM: CPM 0.42.1–0.42.3 and the current 0.43.1/master retain the same relevant implementation, so upgrading CPM alone will not fix it. [CPM 0.43.1 implementation](https://github.com/cpm-cmake/CPM.cmake/blob/v0.43.1/cmake/CPM.cmake#L1195-L1230)
CMake documents that the full-arguments FetchContent_Populate() form ignores previously saved declaration details, which explains the behavior exactly. [CMake FetchContent documentation](https://cmake.org/cmake/help/v3.28/module/FetchContent.html#fetchcontent-populate)
I would not downgrade CMake for this. The better fix is narrowly addressing CPM/RAPIDS CMake’s DOWNLOAD_ONLY interaction so it propagates the override into CPM’s own declaration. Until that exists upstream, a targeted DLPack workaround is safer than changing the build’s CMake version.
ve

robertmaynard:
## Description
Fixes #1088 

In general we should not use `DOWNLOAD_ONLY` and instead use the pattern of `SOURCE_SUBDIR` pointing to a non-existent directory.

## Checklist


docs/conf.py
```diff
@@ -139,3 +139,7 @@
 numpydoc_class_members_toctree = False
 
 autoclass_content = "init"
+
+suppress_warnings = [
+    'intersphinx.missing',  # Suppress missing intersphinx inventory warnings
```
bdice: What caused this warning?

bdice (dismissed):

docs/conf.py
bdice: Ah, I see. https://github.com/rapidsai/rapids-cmake/pull/1089/commits/520e6e9ab6a9d25164e44e003956203d9a4e097e

robertmaynard: There was/is a DDOS on the CMake servers so any job fetching data from cmake.org ( we did via a sphinx plugin ) was failing due to 403's.

robertmaynard: We should be safe to remove the warning suppression as the plugin && docs are back online.

msarahan:
Should this be re-targeted to release/26.10? We will need it for the standalone NVAIE builds one way or another.

robertmaynard:
> Should this be re-targeted to release/26.10? We will need it for the standalone NVAIE builds one way or another.

Re-targetted to 26.10 to simplify our lives

msarahan (approved):

KyleFromNVIDIA (approved):

